PR #1, opened 2016-01-13
Initial script implementation.

doctoryes:
@macdiesel @feanil @clintonb @benpatterson 

Not the final form of these scripts - but this is the general direction. Use a fork of PyGithub to which I've added a new GitHub API call wrapper.

### DIFF
--- a/scripts/github/check_pr_against_branch.py
+++ b/scripts/github/check_pr_against_branch.py
@@ -1,5 +1,5 @@
 import click
-from github_utils import GitHubApiUtils, EDX_PLATFORM_REPO_ID
+from github_api_utils import GitHubApiUtils, EDX_PLATFORM_REPO_ID
 
 
 @click.command()
--- a/scripts/github/check_pr_against_branch.py
+++ b/scripts/github/check_pr_against_branch.py
@@ -1,0 +1,35 @@
+import click
+from github_utils import GitHubApiUtils, EDX_PLATFORM_REPO_ID
+
+
+@click.command()
+@click.option('--repo_id', '-r',
+              default=EDX_PLATFORM_REPO_ID,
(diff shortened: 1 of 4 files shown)


scripts/github/check_pr_against_branch.py
```diff
@@ -0,0 +1,35 @@
+import click
+from github_utils import GitHubApiUtils, EDX_PLATFORM_REPO_ID
```
feanil: Where is the `github_utils` file?

doctoryes: What? Me test the final code? :)
I changed the filename before committing, of course.

scripts/github/github_api_utils.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+import os
+import github
+
+
+# GitHub repo ID for https://github.com/edx/edx-platform
+EDX_PLATFORM_REPO_ID = 10391073
+
+
+class GitHubApiUtils(object):
```
macdiesel, line 12: nit: Since all actions on this class are tied to a particular repo this might be better named as GitHubRepoAPIUtils.

macdiesel:
We have the codes! :+1: :goat:

benpatterson:
Are tests coming?

doctoryes:
Any scripts that we use in the final pipeline should have tests - agreed. Since we're using these to perform discovery work in mocking up the pipeline in several different tools, I punted on tests for now. Thoughts?

benpatterson:
Ah gotcha. That makes sense. Let's just keep one another honest if this evolves. :)

scripts/github/github_api_utils.py
doctoryes, line 12: Probably so - but I expect this to be refactored a few times. :)

PR #2, opened 2016-01-19
GitHub script improvements

doctoryes:
Read parameters from environment variables.
Add some prints for logging.
Return the correct exit status code on success/failure.

I'm successfully calling these from a GoCD pipeline now - and getting the correct success/failure back.

@macdiesel @feanil @clintonb

### DIFF
--- a/scripts/github/check_pr_against_branch.py
+++ b/scripts/github/check_pr_against_branch.py
@@ -1,3 +1,5 @@
+import os
+import sys
 import click
 from github_api_utils import GitHubApiUtils, EDX_PLATFORM_REPO_ID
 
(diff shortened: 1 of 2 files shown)


scripts/github/check_pr_against_branch.py
```diff
@@ -6,29 +8,46 @@
...
     Check if the PR is against the specified branch,
     i.e. if the base of the PR is the specified branch.
     """
     # github.enable_console_debug_logging()
     gh = GitHubApiUtils(repo_id)
-    return gh.is_branch_base_of_pr(pr_number, branch_name)
+
+    if pr_number is None:
+        if pr_env_var:
+            pr_number = int(os.environ[pr_env_var])
```
feanil, line 37: This will throw an error if the environment var is not actually set.  Do we want to do anything in that case?

doctoryes, line 37: It's a good test of each pipeline tool's error capture! :)

I assume that the script will exit with a non-zero exit code on an un-caught exception?

feanil, line 37: Yes, it will.

PR #3, opened 2016-01-21
Add delay and HipChat posting scripts.

doctoryes:
Keeping on keeping on. Are folks finding these useful?

@macdiesel @feanil @clintonb @benpatterson

### DIFF
--- a/scripts/submit_hipchat_msg.py
+++ b/scripts/submit_hipchat_msg.py
@@ -1,0 +1,44 @@
+import os
+import sys
+import requests
+import click
+
+
+HIPCHAT_API_URL = "http://api.hipchat.com"
+NOTIFICATION_POST = "/v2/room/{}/notification"
+AUTH_HEADER = "Authorization: Bearer {}"
+
+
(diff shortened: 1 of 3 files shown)


scripts/submit_hipchat_msg.py
```diff
@@ -0,0 +1,44 @@
...
+import click
+
+
+HIPCHAT_API_URL = "http://api.hipchat.com"
+NOTIFICATION_POST = "/v2/room/{}/notification"
+AUTH_HEADER = "Authorization: Bearer {}"
+
+
+@click.command()
+@click.option('--auth_token_env_var', '-a',
```
feanil, line 13: Why not just pass in the token? or have a default env_var token name?

feanil:
One nit but generally :+1:

scripts/submit_hipchat_msg.py
doctoryes, line 13: As we discussed, feel free to change this or add a default.

PR #4, opened 2016-01-22
Make tubular an installable module.

doctoryes:
Also, add some GoCD-specific code, including a preliminary XML config file.
All this stuff is still just for the pipeline evaluations.

@feanil @macdiesel @benpatterson @clintonb


setup.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+setup(
+    name='tubular',
+    version='0.1',
+    description='Continuous Delivery scripts for pipeline evaluation',
+    packages=['tubular', 'tubular.gocd', 'tubular.scripts', 'tubular.scripts.github', 'tubular.scripts.hipchat'],
+    install_requires=[
+        'click',
+        'requests',
```
feanil: What about pygithub?  How do we ensure the correct fork gets installed?

setup.py
```diff
@@ -0,0 +1,13 @@
...
+    ],
+)
```
clintonb: Add a newline.

setup.py
clintonb: Requirements should be versioned. Ranges are acceptable.

setup.py
```diff
@@ -0,0 +1,13 @@
+from setuptools import setup
+
+
+setup(
+    name='tubular',
+    version='0.1',
```
clintonb: Semantic versioning: 0.0.1

clintonb:
Unless there are test-related requirements (e.g. `pylint`) I would remove `requirements.pip`. Too often I see requirements files become out-of-date with `setup.py`. This can be completely avoided by installing the package as editable: `pip install -e .`

When test-related requirements are needed, add them to a requirements file along with, `-e .`.

setup.py
doctoryes: @clintonb Done.

setup.py
doctoryes: Done.

setup.py
doctoryes: Done.

doctoryes:
Unfortunately, it seems that `setup.py` cannot properly install a GitHub fork. (@nedbat checked in the Python IRC channel - and I tried various things.) Do you guys know a way to do this? Hosting our own PyPI repo would work - but seems heavy-handed.

The scheme now will work with `pip install -r requirements.pip` - all requirements and tubular will install properly.

clintonb:
@doctoryes @nedbat see http://mike.zwobble.org/2013/05/adding-git-or-hg-or-svn-dependencies-in-setup-py/.

doctoryes:
Unfortunately, the method in the article above doesn't work - I tried it. First, `dependency_links` isn't supported unless an `pip` option is used - `--process-dependency-links`. Also, since PyGithub has an existing package in PyPI, the setup.py process looks at those PyPI versions instead of following the GitHub link provided.

clintonb:
What's special about this fork? What are the prospects of these changes being merged into the main repo?

doctoryes:
The fork just adds a check for combined status that's missing from the current module. I've submitted an upstream PR here:
https://github.com/PyGithub/PyGithub/pull/382

feanil:
We should make a seperate setup.py PR and PR for scripts you are adding. Also, it looks like you're using requirements.pip here and I merged in requirements.txt already.

doctoryes:
@feanil @macdiesel I've rebased and removed requirements.pip.

feanil:
:+1:

PR #5, opened 2016-02-08
Add an ec2 utility function for getting AMI Info.

feanil:
The edc_for_ami function will get the EDC tags of an AMI using boto.


feanil:
@macdiesel @doctoryes This bit is stand-alone and working so I thought I would get it reviewed independently and merge as a beginning for the code migration work.

tubular/ec2.py
```diff
@@ -0,0 +1,48 @@
...
+
+    try:
+        ami = ec2.get_all_images(ami_id)[0]
+    except EC2ResponseError as error:
+        raise ImageNotFoundException(error.message)
+
+    tags = ami.tags
+
+    # TODO How do we want to handle these tags not existing?
+    # raise an exception maybe? Right now this is not safe.
```
macdiesel: nit: It does raise an exception currently, should update Doc

tubular/ec2.py
```diff
@@ -0,0 +1,48 @@
...
+def edc_for_ami(ami_id):
+    """
+    Look up the EDC tags for an AMI.
+
+    Input: An AMI Id.
+    Output: An EDC Named Tuple.
+    Exceptions:
+        - ImageNotFoundException
+        - MissingTagException
+    """
```
macdiesel, line 25: Docstrings should match edx recommended style:
https://github.com/edx/edx-platform/wiki/Python-Guidelines (assuming this is still current)

tubular/utils.py
```diff
@@ -0,0 +1,7 @@
+from collections import namedtuple
+
+class TimeoutException(Exception):
```
macdiesel: What do you think about putting exceptions in their own Exceptions.py file? This seems a bit out of place.

macdiesel:
Done with my pass.

tubular/utils.py
feanil: Seems reasonable, would the ones currently in ec2.py also go there?

feanil:
@macdiesel comments addressed. I decided to move all the exceptions into the new exceptions module since I'm seeing a few libraries I'm using already using that pattern.

macdiesel:
:ship: :it: :+1:

PR #7, opened 2016-02-11
Feanil/asgard functions

feanil:
@macdiesel @doctoryes

### DIFF
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -1,0 +1,87 @@
+import os
+import logging
+import requests
+from requests.exceptions import ConnectionError
+from collections import Iterable
+
+ASGARD_API_ENDPOINT = os.environ.get("ASGARD_API_ENDPOINTS", "http://dummy.url:8091")
+ASGARD_API_TOKEN = os.environ.get("ASGARD_API_TOKEN")
+
+CLUSTER_LIST_URL= "{}/cluster/list.json".format(ASGARD_API_ENDPOINT)
+
+LOG = logging.getLogger(__name__)
+
+class BackendError(Exception):
+    pass
+
+
+class BackendConnectionError(BackendError):
+    pass
+
+
+class BackendDataError(BackendError):
+    pass
+
+
+def clusters_for_asgs(asgs):
+    """
+    An autoscaling group can belong to multiple clusters potentially.
+
+    This function finds all asgard clusters for a list of ASGs.
+    eg. get all clusters that have the 'edxapp' cluster tag..
+
+    Input:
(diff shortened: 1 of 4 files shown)


macdiesel:
@feanil there are lots of places in asgard.py where you are catching an exception and throwing a new one without including any information from the prior exception that happened. Since python exceptions are unchecked I would I would prefer if we could just use the exception that come out of the box with the requests library than trying to do some funky python 2.7 exception wrapping. 

This should ease troubleshooting of these scripts down the road. 

Also I was going through the docs for the request library and saw this: http://docs.python-requests.org/en/master/user/quickstart/#response-status-codes might simplify the code a little bit for response codes.

tubular/asgard.py
```diff
@@ -0,0 +1,76 @@
...
+    eg.
+    {
+        u'test-edx-edxapp': [
+            u'test-edx-edxapp-v007',
+        ],
+        u'test-edx-worker': [
+            u'test-edx-worker-v004',
+        ]
+    }
+
```
macdiesel, line 43: Documentation needs to have raises added for expected exceptions that are thrown.

tubular/asgard.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+
+    request = requests.Request('GET', CLUSTER_LIST_URL, params=ASGARD_API_TOKEN)
+    url = request.prepare().url
+    LOG.debug("Getting Cluster List from: {}".format(url))
+    try:
+        response = requests.get(CLUSTER_LIST_URL, params=ASGARD_API_TOKEN)
+    except ConnectionError as ce:
+        raise BackendConnectionError("Unable to connect to asgard: {}".format(url))
```
macdiesel, line 52: -- Exception wrapping

tubular/asgard.py
```diff
@@ -0,0 +1,76 @@
...
+
+    try:
+        cluster_json = response.json()
+    except ValueError as ve:
+        raise BackendDataError("Invalid Json returned for api request: {}".format(response.text))
```
macdiesel, line 57: -- Exception wrapping

tubular/tests/test_asgard.py
```diff
@@ -0,0 +1,124 @@
...
+bad_cluster_json2 = """
+[
+  {
+    "autoScalingGroups":
+    [
+      "loadtest-edx-edxapp-v058",
+      "loadtest-edx-edxapp-v059"
+    ]
+  }
+]"""
```
macdiesel, line 56: The formatting of the output from asgard is tested in asgard, I don't think we need to test it here as well.

macdiesel:
done with my first pass.

PR #9, opened 2016-02-23
Feanil/asgard test task completion

feanil:
@macdiesel @doctoryes Please take a look.  We will probably update these as we go if we find they don't behave how we would want but this is a good starting point for programmatically interacting with asgard.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
 # command to install dependencies
 install: "pip install -r requirements/test_requirements.txt"
 
-# command to run tests
-script: nosetests
+# command to run tests, the --process=-1 will run tests in
+# parallel with as many cores as are available.
+script: nosetests --processes=-1
 
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -6,8 +6,8 @@ import time
 import traceback
 from requests.exceptions import ConnectionError
 from collections import Iterable
-from .exception import *
-from .ec2 import *
+import exception
+import ec2
 
 
 ASGARD_API_ENDPOINT = os.environ.get("ASGARD_API_ENDPOINTS", "http://dummy.url:8091")
@@ -70,7 +70,7 @@ def clusters_for_asgs(asgs):
     for cluster in cluster_json:
         if "autoScalingGroups" not in cluster or "cluster" not in cluster:
             msg = "Expected 'cluster' and 'autoScalingGroups' keys in dict: {}".format(cluster)
-            raise BackendDataError(msg)
+            raise exception.BackendDataError(msg)
 
         for asg in cluster['autoScalingGroups']:
             LOG.debug("Membership: {} in {}: {}".format(asg, asgs, asg in asgs))
@@ -105,7 +105,7 @@ def asgs_for_cluster(cluster):
     except (KeyError,TypeError) as e:
         msg = "Expected a list of dicts with an 'autoScalingGroupName' attribute. " \
               "Got: {}".format(asgs)
-        raise BackendDataError(msg)
+        raise exception.BackendDataError(msg)
 
     return asg_names
 
@@ -135,7 +135,7 @@ def wait_for_task_completion(task_url, timeout):
             return response.json()
         time.sleep(1)
 
-    raise TimeoutException("Timedout while waiting for task {}".format(task_url))
+    raise exception.TimeoutException("Timedout while waiting for task {}".format(task_url))
 
 def new_asg(cluster, ami_id):
     """
@@ -149,7 +149,8 @@ def new_asg(cluster, ami_id):
         str: The name of the new ASG.
 
     Raises:
-        TimeoutException: When the task to bring up the new ASG fails.
+        TimeoutException: When the task to bring up the new ASG times out.
+        BackendError: When the task to bring up the new ASG fails.
     """
     payload = {
         "name": cluster,
@@ -163,7 +164,7 @@ def new_asg(cluster, ami_id):
     response = wait_for_task_completion(response.url, DEFAULT_WAIT_TIMEOUT)
     if response['status'] == 'failed':
         msg = "Failure during new ASG creation. Task Log: \n{}".format(response['log'])
-        raise BackendError(msg)
+        raise exception.BackendError(msg)
 
     # Potential Race condition if multiple people are making ASGs for the same cluster
     # Return the name of the newest asg
@@ -192,7 +193,7 @@ def enable_asg(asg):
     task_status = wait_for_task_completion(task_url, 301)
     if task_status['status'] == 'failed':
         msg = "Failure while enabling ASG. Task Log: \n{}".format(task_status['log'])
-        raise BackendError(msg)
+        raise exception.BackendError(msg)
 
 def disable_asg(asg):
     """
@@ -204,18 +205,18 @@ def disable_asg(asg):
     task_status = wait_for_task_completion(task_url, 300)
     if task_status['status'] == 'failed':
         msg = "Failure while disabling ASG. Task Log: \n{}".format(task_status['log'])
-        raise BackendError(msg)
+        raise exception.BackendError(msg)
 
 def deploy(ami_id):
     LOG.info( "Processing request to deploy {}.".format(ami_id))
 
     # Pull the EDC from the AMI ID
-    edc = edc_for_ami(ami_id)
+    edc = ec2.edc_for_ami(ami_id)
 
     LOG.info("Looking for which clusters to deploy to.")
 
     # These are all autoscaling groups that match the tags we care about.
-    asgs = asgs_for_edc(edc)
+    asgs = ec2.asgs_for_edc(edc)
 
     # All the ASGs except for the new one
     # we are about to make.
@@ -233,7 +234,7 @@ def deploy(ami_id):
             raise
 
     LOG.info("New ASGs: {}".format(new_asgs.values()))
-    wait_for_in_service(new_asgs.values(), 300)
+    ec2.wait_for_in_service(new_asgs.values(), 300)
     LOG.info("ASG instances are healthy. Enabling Traffic.")
 
     elbs_to_monitor = []
@@ -254,7 +255,7 @@ def deploy(ami_id):
 
     # Wait for all instances to be in service in all ELBs
     try:
-        wait_for_healthy_elbs(elbs_to_monitor, 600)
+        ec2.wait_for_healthy_elbs(elbs_to_monitor, 600)
     except:
         LOG.info(" Some instances are failing ELB health checks. "
               "Pulling out the new ASG.")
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -6,5 +6,5 @@ python:
 install: "pip install -r requirements/test_requirements.txt"
 
 # command to run tests
-script: nosetests
+script: nosetests --processes=-1
```
doctoryes: Nit: Maybe a comment here to tell what this option does? I had to look this up:
http://nose.readthedocs.org/en/latest/man.html#cmdoption--processes

tubular/asgard.py
```diff
@@ -1,14 +1,25 @@
+from datetime import datetime,timedelta
 import os
 import logging
 import requests
+import time
+import traceback
 from requests.exceptions import ConnectionError
 from collections import Iterable
 from .exception import *
+from .ec2 import *
```
doctoryes: Nit: My preference is for either:
-  `import ec2` to import everything and force all imported methods to be called like `ec2.foofunc()`
- `from ec2 import foofunc, barfunc` to explicitly import used funcs/defs

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+    Arguments:
+        cluster(str): The name of the asgard cluster.
+
+    Returns:
+        list: List of ASGs.
+    """
+
+    LOG.debug("URL: {}".format(CLUSTER_INFO_URL.format(cluster)))
+    url = CLUSTER_INFO_URL.format(cluster)
+    response = requests.get(url, params=ASGARD_API_TOKEN)
```
doctoryes: When no `timeout` parameter is specified to `requests.get`, the implication is that waiting an infinite amount of time is OK. I prefer to see explicit timeouts specified and failed calls after waiting a reasonable amount of time. Thoughts?

doctoryes: Applies to the other two `requests.get` calls here also....

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+    response = requests.post(NEW_ASG_URL, data=payload, params=ASGARD_API_TOKEN)
+    LOG.debug("Sent request to create new ASG in Cluster({}).".format(cluster))
+
+    #TODO: Make sure response is not an error.
+    response = wait_for_task_completion(response.url, DEFAULT_WAIT_TIMEOUT)
+    if response['status'] == 'failed':
+        msg = "Failure during new ASG creation. Task Log: \n{}".format(response['log'])
+        raise BackendError(msg)
+
+    # Potential Race condition if multiple people are making ASGs for the same cluster
```
doctoryes: I don't know the Asgard API at all - but is there a name we can give the new ASG so that we can look for that exact name below to remove this race condition?

tubular/asgard.py
doctoryes: And the `requests.post` calls....

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+        msg = "Failure while disabling ASG. Task Log: \n{}".format(task_status['log'])
+        raise BackendError(msg)
+
+def deploy(ami_id):
+    LOG.info( "Processing request to deploy {}.".format(ami_id))
+
+    # Pull the EDC from the AMI ID
+    edc = edc_for_ami(ami_id)
+
+    LOG.info("Looking for which clusters to deploy to.")
```
doctoryes: Is this log needed? If so, shouldn't it contain the edc info?

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+
+    new_asgs = {}
+    for cluster in existing_clusters.keys():
+        try:
+            new_asgs[cluster] = new_asg(cluster, ami_id)
+        except:
+            msg = "Failed to create new asg for {} but did make asgs for {}"
+            msg = msg.format(cluster, new_asgs.keys())
+            LOG.error(msg)
+            raise
```
doctoryes, line 250: Maybe `raise` with the constructed `msg`?

tubular/asgard.py
doctoryes: So - looks like there is a `name` parameter in the API call to create a new ASG. Can the returned JSON from `cluster/show` be searched for that specified name?

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+            elbs = response.json()['group']['loadBalancerNames']
+            elbs_to_monitor.extend(elbs)
+        except:
+            LOG.error(traceback.format_exc())
+            LOG.error("Something went wrong with {}, disabling traffic.".format(asg))
+            disable_asg(asg)
+            raise
+
+    LOG.info("All new ASGs are active.  The new instances "
+          "will be available when they pass the healthchecks.")
```
doctoryes, line 268: Include the names of the active ASGs in this log message?

tubular/ec2.py
```diff
@@ -4,9 +4,13 @@
 """
 import boto
 import logging
+import time
 from utils import EDC
-from exception import ImageNotFoundException, MissingTagException
+from exception import ImageNotFoundException
+from exception import MissingTagException
+from exception import TimeoutException
```
doctoryes: Here you can:

```
from exception import (
    ImageNotFoundException,
    MissingTagException,
    TimeoutException
)
```

doctoryes:
Whew - lots of test code here! I must admit that I didn't check for complete coverage - but the way it's done looks sound and understandable to me.

First pass complete.

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+    try:
+        asg_names = map(lambda x: x['autoScalingGroupName'], asgs)
+    except (KeyError,TypeError) as e:
+        msg = "Expected a list of dicts with an 'autoScalingGroupName' attribute. " \
+              "Got: {}".format(asgs)
+        raise BackendDataError(msg)
+
+    return asg_names
+
+def wait_for_task_completion(task_url, timeout):
```
macdiesel: you may get some additional mileage here if this function accepted a response object instead of just the URL. A request may need to have additional headers, query parameters, payloads or different http methods. Opening this up a bit more now may lead to an easier transition to spinnaker later.

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+        "name": cluster,
+        "imageId": ami_id,
+    }
+
+    response = requests.post(NEW_ASG_URL, data=payload, params=ASGARD_API_TOKEN)
+    LOG.debug("Sent request to create new ASG in Cluster({}).".format(cluster))
+
+    #TODO: Make sure response is not an error.
+    response = wait_for_task_completion(response.url, DEFAULT_WAIT_TIMEOUT)
+    if response['status'] == 'failed':
```
macdiesel: Can we use the status_code of the response or do we need to check the payload of the object?

tubular/asgard.py
macdiesel: Good idea @doctoryes. 

We may also want to think about postfixing some type of identifier to the ASG name as well that could identify perhaps the time it was created as well as the machine or user that it came from? Or it won't matter if we begin to specify the name?

tubular/asgard.py
macdiesel, line 250: Lets say we we have n = 5 clusters. ASGs could be made for n clusters then fail on the second iteration. At that point we have created 2 ASGs but the 3rd failed. We do not roll back the creation of the 2 that have failed and we do not proceed to attempt to create the ASGs for n = 4 and n = 5. Sound right? 

My concern is that this could leave the deployments in an inconsistent state that may need to be cleaned up manually later.

Is this the desired behavior? We may want to limit this functionality down to working on either a single ASG or attempting to create everything then reporting back on the errors.

tubular/asgard.py
```diff
@@ -72,3 +81,190 @@ def clusters_for_asgs(asgs):
...
+
+    LOG.info("New ASGs: {}".format(new_asgs.values()))
+    wait_for_in_service(new_asgs.values(), 300)
+    LOG.info("ASG instances are healthy. Enabling Traffic.")
+
+    elbs_to_monitor = []
+    for cluster, asg in new_asgs.iteritems():
+        try:
+            enable_asg(asg)
+            response = requests.get(ASG_INFO_URL.format(asg), params=ASGARD_API_TOKEN)
```
macdiesel: I would recommend pulling this out in to a function. It's the only time you mess with teh requests library in this function and if it's independent it will be easier to unit test this code path.

tubular/asgard.py
feanil: Unfortunately not.

tubular/asgard.py
feanil: The name field is the name of the cluster not the name of the autoscaling group.  We could add some smarts around this code because the ASGs have a numeric suffix that gets incremented but I didn't do that yet mostly because it involved parsing logic and I didn't think it was worth the time yet.  If it seems like it's worth it I'm happy to add that in now.

tubular/asgard.py
feanil: I feel like this whole function is specific to asgard and we would probably have different abstractions for spinnaker or other deploy tools in the future. I would think that when we go down that road we would build out a separate module for spinnaker that maps to it's abstractions.  I don't think that the functions here necessarily represent a good interface for a general deploy process necessarily.

tubular/asgard.py
feanil, line 250: I think we want the ability to work against multiple ASGs because we have this use case now, both the workers and the edxapp need the same AMI deployed to them, at about the same time.  But my thought was that if either of them failed, we wouldn't want to be too smart about doing as much as we can and instead failing as fast as possible.

feanil:
@doctoryes @macdiesel all comments addressed or replied to.

tubular/asgard.py
```diff
@@ -1,14 +1,26 @@
...
-from .exception import *
+import exception
+import ec2
+
 
 ASGARD_API_ENDPOINT = os.environ.get("ASGARD_API_ENDPOINTS", "http://dummy.url:8091")
-ASGARD_API_TOKEN = os.environ.get("ASGARD_API_TOKEN")
+ASGARD_API_TOKEN = {"asgardApiToken": os.environ.get("ASGARD_API_TOKEN", "dummy-token")}
+ASGARD_WAIT_TIMEOUT = int(os.environ.get("ASGARD_WAIT_TIMEOUT", 300))
+REQUESTS_TIMEOUT = os.environ.get("REQUESTS_TIMEOUT", 1)
```
doctoryes: Is one second the right default here? If we're encouraging users to set this, I'd make it more harsh (like 0.001). If we're wanting it to default to something reasonable, it should be ~10 seconds or more.

macdiesel: I fall into the make it something reasonable camp.

feanil:
@doctoryes @macdiesel last comment addressed and tested against the sandbox asgard.

PR #11, opened 2016-03-09
Adding executable python script to deploy AMIs

macdiesel:
This installs the asgard-deploy.py as an executable script in to the python environment.

@feanil @doctoryes PTAL.


feanil:
:+1:

PR #13, opened 2016-03-16
Don't pct encode the token.

feanil:
The tokens can have characters that requests doesn't percent encode correctly(the '%' character)

@edx/pipeline-team


macdiesel:
:+1:

PR #14, opened 2016-03-24
consolidating scripts, all scripts are now installed as executable when setup.py install is run

macdiesel:
@feanil @doctoryes PTAL


setup.py
```diff
@@ -11,5 +13,5 @@
         'click>=6.2',
         'requests>=2.9',
     ],
-    scripts=['scripts/asgard-deploy.py']
+    scripts=[os.path.join(dp, f) for dp, dn, filenames in os.walk(PATH) for f in filenames if os.path.splitext(f)[1] == '.py' and not f == '__init__.py']
```
feanil, line 16: Nit:  I'd rather see this as a readable multi line function instead of a weird double for-loop on-liner.

feanil:
:+1:

setup.py
doctoryes, line 16: Agreed - but not a nit. :smiley_cat:

doctoryes, line 16: Also, this code takes all Python files and puts them as executables, including `setup.py` and all test files. Seems like instead we want only files in `scripts` directories to be installed as executables.

macdiesel, line 16: I can split this out into a multiliner if that is what we want.

@doctoryes I relocated the asgard-deploy.py script to these directories so all the scripts would be grouped together. As far as I can tell this code is only pulling out the items under tubular/scripts, it was my understanding that we wanted all the scripts to be installed as executable.

macdiesel:
@feanil & @doctoryes do we even want to pull this forward now as we will be using this repository as a material inside of go-cd? We would probably be better off calling these scripts by path since they will be updated much more frequently. 

I vote for a close and perhaps a PR to remove the installing of the scripts from the setup.py entirely.

setup.py
doctoryes, line 16: Ah - you're correct. I missed that `PATH` detail.

doctoryes:
Good point. Better to have a standard way of calling these scripts than two different ways. Though - we'll have problems if we ever need to move the scripts around, as GoCD pipeline will point to paths. We should think about how to insulate ourselves from this to maintain backward compatibility.

feanil:
One way we could do it is to always call the installed paths and pip install the tubular repository as the first step of any scripts.  That way in the future if once the scripts have solidified a bit more, we can just pull out the pip install and all the paths will remain correct?  Not sure how much that helps because we would still need to go and remove all the pip install calls but at least then we can't use the scripts from the wrong location.

PR #17, opened 2016-04-14
Fixes for TE-1273 - AMIs now deploy correctly

macdiesel:
Also expanded testing for checking instance states inside of ELBs.

@edx/pipeline-team PTAL.

### DIFF
--- a/tubular/tests/test_ec2.py
+++ b/tubular/tests/test_ec2.py
@@ -1,17 +1,12 @@
 from collections import Iterable
 
-import boto
 import mock
 import unittest
-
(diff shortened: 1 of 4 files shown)


tubular/tests/test_ec2.py
```diff
@@ -4,14 +4,16 @@
 import mock
 import unittest
 
+from copy import copy
```
feanil: You don't need copy in this file anymore.

macdiesel: I will remove, are you opposed to removing the other unused imports in this file while I am at it?

feanil:
One nit but otherwise :+1:

PR #18, opened 2016-04-20
Feanil/cluster to play

feanil:
@macdiesel @doctoryes


doctoryes:
The rename looks correct to me. However, I don't know the context of why this is shifting from cluster to play. So - a provisional 👍

feanil:
@doctoryes the reason for the change is to match a concept change, cluster name is expected to uniquely refer to a single service per E-D.  Play is not.  It is okay for multiple clusters to use an AMI built from the same play.  For most services cluster and play have a 1-to-1 mapping.  In the case of the edx-platform, both the machines that run the lms/cms and the machines that run the celery workers use the same base image but just start different services.

PR #20, opened 2016-04-25
Use all command args.

doctoryes:
@macdiesel For you review. This script activates the virtualenv and calls a command with all included parameters.

 @feanil FYI.


macdiesel:
👍 🚢 🇮🇹

PR #21, opened 2016-04-28
Add top-level module path before tubular import.

doctoryes:
@macdiesel OK - this method works with no tubular install. The effects should only last for the life of the script. What do you think?

@feanil FYI.


macdiesel:
👍

PR #24, opened 2016-05-12
Ensure the new ASG(s) is not being deleted before deleting existing ASG(s)

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -9,7 +9,7 @@ from collections import Iterable
 import exception
 import ec2
 
-from exception import BackendError
+from exception import BackendError, ASGDoesNotExistException
 
 
 ASGARD_API_ENDPOINT = os.environ.get("ASGARD_API_ENDPOINTS", "http://dummy.url:8091/us-east-1")
@@ -198,7 +198,7 @@ def get_asg_info(asg):
     response = requests.get(url, params=ASGARD_API_TOKEN, timeout=REQUESTS_TIMEOUT)
 
     if response.status_code == 404:
-        raise BackendError('Autoscale group {} does not exist'.format(asg))
+        raise ASGDoesNotExistException('Autoscale group {} does not exist'.format(asg))
     elif response.status_code >= 500:
         raise BackendError('Asgard experienced an error: {}'.format(response.text))
     elif response.status_code != 200:
@@ -242,6 +242,7 @@ def is_asg_pending_delete(asg):
     else:
         return True
 
+
 def enable_asg(asg):
     """
     Enable an ASG in asgard.  This means it will have ELBs routing to it
@@ -264,6 +265,7 @@ def enable_asg(asg):
     if task_status['status'] == 'failed':
         msg = "Failure while enabling ASG. Task Log: \n{}".format(task_status['log'])
         raise exception.BackendError(msg)
+
 
 def disable_asg(asg):
     """
@@ -292,6 +294,7 @@ def disable_asg(asg):
         msg = "Failure while disabling ASG. Task Log: \n{}".format(task_status['log'])
         raise exception.BackendError(msg)
 
+
 def elbs_for_asg(asg):
     response = requests.get(ASG_INFO_URL.format(asg),
         params=ASGARD_API_TOKEN, timeout=REQUESTS_TIMEOUT)
@@ -318,6 +321,7 @@ def deploy(ami_id):
     Raises:
         TimeoutException: When the task to bring up the new instance times out.
         BackendError: When the task to bring up the new instance fails.
+        ASGDoesNotExistException: If the ASG being queried does not exist.
     """
     LOG.info( "Processing request to deploy {}.".format(ami_id))
 
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -178,17 +178,21 @@ def new_asg(cluster, ami_id):
 
     return new_asg
 
-def is_asg_pending_delete(asg):
-    """
-    Checks status of an ASG, specifically if it is pending deletion.
 
-    Argument:
-        asg(str): ASG whose status should be checked.
+def get_asg_info(asg):
+    """
+    Queries Asgard for the status info on an ASG
+
+    Arguments:
+        cluster(str): Name of the cluster.
 
     Returns:
-        True if the asg is in the "pending delete" status, else return False.
-    """
+        dict: a Dictionary with the information about an ASG.
 
+    Raises:
+        TimeoutException: when the request for an ASG times out.
+        BackendError: When a non 200 response code is returned from the Asgard API
+    """
     url = ASG_INFO_URL.format(asg)
     LOG.debug("URL: {}".format(url))
     response = requests.get(url, params=ASGARD_API_TOKEN, timeout=REQUESTS_TIMEOUT)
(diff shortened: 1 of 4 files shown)


tubular/ec2.py
```diff
@@ -172,7 +151,7 @@ def wait_for_healthy_elbs(elbs_to_monitor, timeout):
         TimeoutException: We we have run out of time.
     """
     boto_elb = boto.connect_elb()
-    elbs_left = list(elbs_to_monitor)
+    elbs_left = set(elbs_to_monitor)
```
macdiesel, line 153: When multiples of the same load balancer were in this list boto was not always returning the correct information about an ELB. Changing this to a set ensures we receive the correct information (at least during testing) about a load balancers.

tubular/tests/test_asgard.py
```diff
@@ -455,17 +481,6 @@ def test_disable_asg_pending_deletion(self, mock_connect_autoscale):
...
-        mock_inst0 = mock.Mock()
-        mock_inst0.lifecycle_state = "InService"
-        mock_inst1 = mock.Mock()
-        mock_inst1.lifecycle_state = "Terminating:Wait"
-        mock_group0 = mock.Mock()
-        mock_group0.instances = [ mock_inst0, mock_inst1 ]
-        mock_get_all_groups = mock.Mock()
-        mock_get_all_groups.get_all_groups.return_value = [ mock_group0, mock.Mock() ]
-        mock_connect_autoscale.return_value = mock_get_all_groups
-
```
doctoryes, line 468: Good riddance!

tubular/asgard.py
```diff
@@ -176,6 +178,36 @@ def new_asg(cluster, ami_id):
...
+        raise BackendError('Autoscale group {} does not exist'.format(asg))
+    elif response.status_code >= 500:
+        raise BackendError('Asgard experienced an error: {}'.format(response.text))
+    elif response.status_code != 200:
+        raise BackendError('Call to asgard failed with status code: {0}: {1}'
+                           .format(response.status_code, response.text))
+
+    LOG.debug("ASG info: {}".format(response.text))
+    asgs = response.json()
+    if asgs['group']['status'] is not None:
```
doctoryes: So a `status` of `None` exclusively means that a delete is pending?

tubular/asgard.py
```diff
@@ -176,6 +178,36 @@ def new_asg(cluster, ami_id):
...
+
+    Argument:
+        asg(str): ASG whose status should be checked.
+
+    Returns:
+        True if the asg is in the "pending delete" status, else return False.
+    """
+
+    LOG.debug("URL: {}".format(ASG_INFO_URL.format(asg)))
+    url = ASG_INFO_URL.format(asg)
```
doctoryes: Nit: Swap the order of these two lines to use the `url` variable in the log?

tubular/asgard.py
```diff
@@ -296,6 +339,11 @@ def deploy(ami_id):
 
     LOG.info("New instances have succeeded in passing the healthchecks. "
           "Disabling old ASGs.")
+    # ensure the new ASG is still healthy and not pending delete before disabling the old ASGs
+    for cluster, asg in new_asgs.iteritems():
+        if is_asg_pending_delete(asg):
+            raise BackendError("New Autoscale Group {} is pending delete, Did someone manually intervene? Aborting".format(asg))
```
doctoryes: Nit: Maybe the question should be removed in favor of adding `Aborting the disabling of old ASGs.`?

tubular/tests/test_asgard.py
```diff
@@ -133,6 +134,31 @@
...
+{{
+	"group": {{
+		"autoScalingGroupName": "{0}",
+		"loadBalancerNames": [
+			"app_elb"
+		],
+		"status": null
+	}}
+}}
+"""
```
doctoryes, line 160: Nit: Derive both these strings from a common string?

tubular/asgard.py
macdiesel: done

tubular/asgard.py
macdiesel: I made this section more clear. If the status is none, no action is taking place and the method returns True.

macdiesel: I made this section more clear. If the status is none, no action is taking place and the method returns False.

tubular/asgard.py
macdiesel: done

doctoryes:
👍

tubular/asgard.py
```diff
@@ -176,6 +178,36 @@ def new_asg(cluster, ami_id):
...
+    Returns:
+        True if the asg is in the "pending delete" status, else return False.
+    """
+
+    url = ASG_INFO_URL.format(asg)
+    LOG.debug("URL: {}".format(url))
+    response = requests.get(url, params=ASGARD_API_TOKEN, timeout=REQUESTS_TIMEOUT)
+
+    if response.status_code == 404:
+        raise BackendError('Autoscale group {} does not exist'.format(asg))
```
feanil: This could happen if a pending delete happened during the deploy, I don't think that's an error.

tubular/asgard.py
```diff
@@ -296,6 +339,11 @@ def deploy(ami_id):
 
     LOG.info("New instances have succeeded in passing the healthchecks. "
           "Disabling old ASGs.")
+    # ensure the new ASG is still healthy and not pending delete before disabling the old ASGs
+    for cluster, asg in new_asgs.iteritems():
+        if is_asg_pending_delete(asg):
```
feanil: The other use case which we should handle is the case of the new ASG being disabled manually as that could also cause downtime.  So we could check to see if the new ASGs are still enabled here or we could update the logic of the `disable_asg` function to do a check to see if it is disabling the last ASG in a cluster(I like this better).

feanil:
I think we should change how the case of a 404 is handled.  The scenario that represents in my mind is that a delete was initiated and finished before deploy of the new ASG finished.  I think this is a valid case.

tubular/asgard.py
macdiesel: @feanil what is your suggestion here? If an ASG doesn't exist, it doesn't exist. We could raise an exception of a different type in this situation, but ultimately the system ends up reacting the same way: Tubular uses this call to see if an ASG is pending delete, and if it is, it halts it's current operation. 

Lets think about it this way for a moment as well. Say the caller of this function passes a bad ASG name. The ASG doesn't exist and we return that this ASG is not pending delete. Their code then goes on from here to delete or disable other ASGs or do whatever it does, while in fact the ASG they had intended to query is actually in the process or a delete operation. The caller is essentially acting on false information.

macdiesel:
@edx/pipeline-team this has been updated and is ready for review again. I have added the call to see if an ASG is disabled. 

@feanil I have responded to your other inquiry, I believe that a 404 is a valid response and should raise an error. We can add a new Error class if you would like to clear up any ambiguity.

tubular/asgard.py
feanil: It seems reasonable to bubble up the decision making. +1 to the idea of a different type of exception.

macdiesel: I'll add it, but I'm interested to hear if/how you think the deploy code should change. 

If the new ASG is pending deletion or if it doesn't exist, we are still going to take the same action (do not proceed in disabling old ASGs and fail the deployment) correct?

feanil:
👍

PR #27, opened 2016-05-13
New ASGs are now checked to see if they are disabled or deleted before disabling existing ASGs

macdiesel:


macdiesel:
Tested this with @jibsheet earlier, merging.

PR #28, opened 2016-05-16
The ability to delete an ASG via asgard.

macdiesel:


tubular/asgard.py
```diff
@@ -294,6 +295,33 @@ def disable_asg(asg):
...
+    curl -d "name=helloworld-example-v004" http://asgardprod/us-east-1/cluster/delete
+
+    Arguments:
+        asg(str): The name of the asg to delete.
+
+    Returns:
+        None: When the asg has been deleted.
+
+    Raises:
+        TimeoutException: If the task to enable the ASG fails...
```
feanil: This should be `If the task to delete ASG fails...`

tubular/asgard.py
```diff
@@ -294,6 +295,33 @@ def disable_asg(asg):
...
+        BackendError: If asgard was unable to disable the ASG
```
feanil: `If asgard was unable to delete the ASG`

feanil:
Two documentation issues but otherwise 👍

macdiesel:
I should have tagged this as WIP as I wasn't ready for review just yet but thanks for the catch.

tubular/asgard.py
macdiesel: done

tubular/asgard.py
macdiesel: done

macdiesel:
@feanil @doctoryes PTAL. Added the ability to delete an ASG. When calling this method by default it will check to see if the ASG is still active. If this is the case an exception will be raised.

feanil:
The other case that we should handle is if they give us the name of an ASG that has already been deleted.

feanil:
👍  once we handle that case.

macdiesel:
@feanil This case is already handled. I have updated the docs to make this more clear.

scripts/delete-asg.py
```diff
@@ -0,0 +1,29 @@
...
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+@click.command()
+@click.option('--asg_name', envvar='ASG_NAME', help='the name of the Autoscale Group to delete', required=True)
+def delete_asg(asg_name):
+    try:
+        asgard.delete_asg(asg_name, True)
+    except Exception, e:
+        traceback.print_exc()
+        click.secho("Error Deploying AMI: {0}.\nMessage: {1}".format(asg_name, e.message), fg='red')
```
doctoryes, line 23: s/Deploying AMI/deleting ASG/

macdiesel, line 23: Bad copy/pasta. Will fix in the next PR.

PR #32, opened 2016-05-20
Added ability to set the sleep time while waiting for tasks to complete.

macdiesel:
@doctoryes @feanil This doesn't solve the issue error handling when we run into throttling, however it will reduce the number of calls being made against the AWS API. Not the solution but a step towards better AWS usage.

PTAL.


doctoryes:
👍

PR #35, opened 2016-05-25
Adding a retry decorator to retry calls against asgard. TE-1371

macdiesel:
@edx/pipeline-team PTAL!

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,4 +12,6 @@ Scripts for integrating pipelines with Asgard and ec2 to facilitate Continuous d
 | ASGARD_API_TOKEN     | dummy-token                     | String - The asgard token                                                                     |
 | ASGARD_WAIT_TIMEOUT  | 600                             | Integer - time in seconds to wait for an action such as instances healthy in a load balancer. |
 | REQUESTS_TIMEOUT     | 10                              | How long to wait for an http connection/response from Asgard.                                 |
-
+| RETRY_MAX_ATTEMPTS   | 5                               | Maximum number attempts to be made when asgard returns a 400 or 500 response.            |
+| RETRY_DELAY_SECONDS  | 5                               | How long in seconds to wait between retries to asgard                                         |
+| RETRY_MAX_TIME_SECONDS | None                          | How long in seconds to keep retrying asgard before giving up.                                 |
--- a/tubular/tests/test_asgard.py
+++ b/tubular/tests/test_asgard.py
@@ -1,3 +1,5 @@
+import sys
+
 import json
 import unittest
 import itertools
@@ -5,7 +7,7 @@ import itertools
 import boto
 import mock
 import httpretty
-from ..asgard import *
+from .. import asgard
 from ..exception import *
 
 from ddt import ddt, data, unpack
@@ -14,6 +16,12 @@ from moto.ec2.utils import random_ami_id
 from requests.exceptions import ConnectionError
 
 from .test_utils import create_asg_with_tags, create_elb
+
+# Disable the retry decorator and reload the asgard module. This will ensure that tests do not fail because of the retry
+# decorator recalling a method when using httpretty with side effect iterators
+sys._is_retry_enabled = False
+reload(asgard)
+
 
 sample_cluster_list = """
 [
@@ -267,34 +275,35 @@ sample_worker_asg_info = """
 }
 """
 
+
 @ddt
 class TestAsgard(unittest.TestCase):
     _multiprocess_can_split_ = True
 
     def test_bad_clusters_endpoint(self):
         relevant_asgs = []
-        self.assertRaises(ConnectionError, clusters_for_asgs, relevant_asgs)
+        self.assertRaises(ConnectionError, asgard.clusters_for_asgs, relevant_asgs)
 
     @httpretty.activate
     def test_clusters_for_asgs(self):
         httpretty.register_uri(
             httpretty.GET,
-            CLUSTER_LIST_URL,
+            asgard.CLUSTER_LIST_URL,
             body=sample_cluster_list,
             content_type="application/json")
 
         relevant_asgs = []
-        cluster_names = clusters_for_asgs(relevant_asgs)
+        cluster_names = asgard.clusters_for_asgs(relevant_asgs)
         self.assertEqual({}, cluster_names)
 
         relevant_asgs = ["loadtest-edx-edxapp-v058"]
         expected_clusters = { "loadtest-edx-edxapp" :
                 ["loadtest-edx-edxapp-v058", "loadtest-edx-edxapp-v059"]}
-        cluster_names = clusters_for_asgs(relevant_asgs)
+        cluster_names = asgard.clusters_for_asgs(relevant_asgs)
         self.assertEqual(expected_clusters, cluster_names)
 
         relevant_asgs = ["loadtest-edx-edxapp-v058", "loadtest-edx-worker-v034"]
-        cluster_names = clusters_for_asgs(relevant_asgs)
+        cluster_names = asgard.clusters_for_asgs(relevant_asgs)
         self.assertIn("loadtest-edx-edxapp", cluster_names)
         self.assertIn("loadtest-edx-worker", cluster_names)
         self.assertEqual(["loadtest-edx-worker-v034"], cluster_names['loadtest-edx-worker'])
@@ -308,17 +317,17 @@ class TestAsgard(unittest.TestCase):
         # The json is valid but not the structure we expected.
         httpretty.register_uri(
             httpretty.GET,
-            CLUSTER_LIST_URL,
+            asgard.CLUSTER_LIST_URL,
             body=response_json,
             content_type="application/json")
 
         relevant_asgs = []
-        self.assertRaises(BackendDataError, clusters_for_asgs, relevant_asgs)
+        self.assertRaises(BackendDataError, asgard.clusters_for_asgs, relevant_asgs)
 
     @httpretty.activate
     def test_asg_for_cluster(self):
         cluster = "prod-edx-edxapp"
-        url = CLUSTER_INFO_URL.format(cluster)
+        url = asgard.CLUSTER_INFO_URL.format(cluster)
         httpretty.register_uri(
             httpretty.GET,
             url,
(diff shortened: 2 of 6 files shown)


tubular/utils/retry.py
```diff
@@ -0,0 +1,145 @@
...
+MAX_ATTEMPTS = os.environ.get('RETRY_MAX_ATTEMPTS', 5)
+DELAY_SECONDS = os.environ.get('RETRY_DELAY_SECONDS', 5)
+MAX_TIME_SECONDS = os.environ.get('RETRY_MAX_TIME_SECONDS', None)
+
+LOG = logging.getLogger(__name__)
+
+
+def retry(attempts=MAX_ATTEMPTS, delay_seconds=DELAY_SECONDS, max_time_seconds=MAX_TIME_SECONDS):
+    """
+    Decorator wraps a function that will attempt to "retry" the function if an exception is raised during exeuction.
```
doctoryes: Nit: s/exeuction/execution.

tubular/utils/retry.py
```diff
@@ -0,0 +1,145 @@
...
+                                     When None the method will continue to be called until max_attempts is reached.
+        """
+        if max_attempts < 1:
+            raise RetryException(
+                "Must specify a max_attempts number greater than 1. Value: {0}".format(max_attempts))
+
+        if delay_seconds < 0:
+            raise RetryException(
+                "Must specify a delay_seconds number greater than 0. Value: {0}".format(delay_seconds))
+
```
doctoryes, line 70: Consider `max_attempts` = 3, `delay_seconds` = 5, and `max_time_seconds` = 4. In this case, no retries will be attempted. Do we want to warn the user in this case?

tubular/utils/retry.py
```diff
@@ -0,0 +1,145 @@
...
+        self.max_attempts = max_attempts
+        self.delay_seconds = delay_seconds
+
+    def max_attempts_reached(self):
+        """
+        Returns:
+            bool: True if the retry has reached it's maximum number of attempts
+                  False if the current attempt is less than the number of attempts
+        """
+        if self._current_attempt_number <= self.max_attempts:
```
doctoryes: Just `return self._current_attempt_number <= self.max_attempts`  ?

tubular/utils/retry.py
```diff
@@ -0,0 +1,145 @@
...
+        else:
+            return True
+
+    def max_time_reached(self):
+        """
+        Returns:
+            bool: True if the maximum runtime of the retry has been met or exceeded
+                  False if the max_time was not set or if the max time has not yet been reached
+        """
+        if self._max_datetime and self._max_datetime - datetime.utcnow() < timedelta(0, 0, 0):
```
doctoryes: Remove the `if` in favor of a single `return`?

tubular/utils/retry.py
```diff
@@ -0,0 +1,145 @@
...
+            return True
+        else:
+            return False
+
+    def get_delay_time(self):
+        """
+        Returns:
+            int: seconds to delay
+        """
+        return round(self.delay_seconds)
```
doctoryes: Maybe just `round` this in the `__init__`?

tubular/utils/retry.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self._current_attempt_number = 0
+        self._max_datetime = datetime.utcnow() + timedelta(0, max_time_seconds) if max_time_seconds else None
+        self.max_attempts = max_attempts
+        self.delay_seconds = delay_seconds
+
+    def max_attempts_reached(self):
+        """
+        Returns:
+            bool: True if the retry has reached it's maximum number of attempts
```
doctoryes: s/it's/its

tubular/utils/retry.py
```diff
@@ -0,0 +1,145 @@
...
+        If successful the result of the function will be returned.
+        If the call ultimately fails with an Exception that exception will be raised
+
+        Arguments:
+            fn (function): the function to execute
+            args(list<any>): Arguments to the wrapped function
+            kwargs(dict<str:any>): Keyword arguments to the wrapped function
+
+        """
+        while not self.max_attempts_reached() and not self.max_time_reached():
```
doctoryes: Maybe pull this logic into a local method?:

```
def done():
    return self.max_attempts_reached() or self.max_time_reached()

while not done():
    ...
```

tubular/tests/test_retry.py
```diff
@@ -0,0 +1,125 @@
+import sys
+import unittest
+import mock
+import datetime
+
+from ..utils import retry
+
+sys._is_retry_enabled = True
```
doctoryes: Why put this global var in `sys` versus in `utils/retry.py`? Seems like it belongs there.

tubular/tests/test_retry.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def test_max_attempts_reached(self):
+        manager = retry.LifecycleManager(1, 1, 1)
+        # first call current_attempt_number should be 0 and max_attempts_reached should return False
+        self.assertFalse(manager.max_attempts_reached())
+
+        manager._current_attempt_number += 1
+        self.assertFalse(manager.max_attempts_reached())
+
+        # incrementing again should put us past the max attempds threshold
```
doctoryes: `attempts`

tubular/tests/test_retry.py
```diff
@@ -0,0 +1,125 @@
...
+        manager._current_attempt_number += 1
+        self.assertTrue(manager.max_attempts_reached())
+
+    def test_max_time_reached_none(self):
+        """
+        When no max time is set, max_time_reached should always return False
+        """
+        manager = retry.LifecycleManager(1, 1, None)
+        self.assertFalse(manager.max_time_reached())
```
doctoryes, line 45: Maybe fake some time passage here and check `max_time_reached()` again?

tubular/tests/test_retry.py
```diff
@@ -0,0 +1,125 @@
...
+        class NewDateTime(datetime.datetime):
+            @classmethod
+            def utcnow(cls):
+                return curr_time
+
+        built_in_datetime = retry.datetime
+
+        # The instance of datetime becomes local to the module it's import in to. We must patch datetime using the
+        # module instance that is imported in to the ec2 module.
+        retry.datetime = NewDateTime
```
doctoryes, line 59: 🐒

README.md
```diff
@@ -12,4 +12,6 @@ Scripts for integrating pipelines with Asgard and ec2 to facilitate Continuous d
 | ASGARD_API_TOKEN     | dummy-token                     | String - The asgard token                                                                     |
 | ASGARD_WAIT_TIMEOUT  | 600                             | Integer - time in seconds to wait for an action such as instances healthy in a load balancer. |
 | REQUESTS_TIMEOUT     | 10                              | How long to wait for an http connection/response from Asgard.                                 |
-
+| RETRY_MAX_ATTEMPTS   | 5                               | Maximum number attempts to be made when asgard returns a 400 or 500 leve response.            |
```
doctoryes: `leve`?

doctoryes:
First pass complete.

tubular/tests/test_asgard.py
```diff
@@ -15,6 +17,11 @@
 
 from .test_utils import create_asg_with_tags, create_elb
 
+# Disable the retry decorator and reload the asgard module. This will ensure that tests do not fail because asgard
```
feanil: `because asgard`.....?

tubular/tests/test_retry.py
feanil: Why not an environment variable, we could have it disabled by default and only turn it on during the testing of the feature and in production?

tubular/utils/retry.py
macdiesel: done

tubular/tests/test_retry.py
macdiesel, line 45: The code in LifecycleManager.max_time_reached() always returns False if None is passed as the max_time_seconds in the dunder init. I don't think we need anything else here to validate this case. If the variable is set to None, it should return false regardless of the passing of time.

tubular/tests/test_retry.py
macdiesel, line 59: yeah, wish mock.patch would work for this stuff we didn't have to do it manually. =(

README.md
macdiesel: fat fingers... done

tubular/tests/test_retry.py
macdiesel: done

tubular/utils/retry.py
macdiesel: done(), now with more testing

tubular/utils/retry.py
macdiesel: done

tubular/utils/retry.py
macdiesel: done.

tubular/utils/retry.py
macdiesel: solid, done.

tubular/utils/retry.py
macdiesel: solid advice, done.

tubular/utils/retry.py
macdiesel, line 70: done I have added a log.warn statement

tubular/tests/test_asgard.py
macdiesel: done

macdiesel:
@doctoryes @feanil as to why use sys vs the retry module or an environment variable:
- Environment variable could be used, but it would need to be set before the retry module was loaded. Also since there is no common entry right now for launching tests there isn't a good place to put it. It's easy to specify the variable for travis but what about running the tests locally. I could add a test launching script but then there are docs to add and everyone need to actually read them before executing tests. It just didn't feel user friendly.

As to why in sys and not in the retry module. The variable needs to be set before the retry module is loaded. I needed a module that will not be destroyed during the lifecycle of the application and something that could be shared by all. sys seemed like a good place as any. 

Remember this is only used for testing. If you guys want a longer run down of what works and doesn't work when loading these modules to disable/enable a decorator it might be easier do a 15 minute huddle up and talk about it.

macdiesel:
@edx/pipeline-team Ready for another review.

macdiesel:
@doctoryes the sys code has been replaced with an environment variable. PTAL.

doctoryes:
👍

feanil:
👍

PR #38, opened 2016-05-27
Import cleanup and python 3 compatibility fixes

macdiesel:
Some import cleanup (absolute paths are now used) and some simple python 3 compatibility fixes.

@edx/pipeline-team PTAL.


macdiesel:
As a note, the reason the absolute imports work correctly now is that @doctoryes has added the root path to the python path: https://github.com/edx/tubular/blob/master/scripts/cleanup-asgs.py#L9

doctoryes:
@macdiesel I'm puzzled by the line of tests at the end that show up as skipped (`-`). These skipped tests did not appear on your PR - they're on master as well. Any ideas why?

But - 👍

PR #40, opened 2016-06-07
delete_asg expects a string not a boto AutoScale object.

feanil:
@edx/pipeline-team 

There a mismatch between the return type of the `get_asgs_pending_delete` and the input type of `asgard.delete_asg`.  The output is a list of `boto` objects but the input of the delete command is the string name of the ASG.


macdiesel:
👍

PR #48, opened 2016-06-29
Better handle when ASGs have been deleted.

feanil:
https://openedx.atlassian.net/browse/TE-1412
ASGs get deleted by janitor processes regularly and so we should generally
handle those cases better.  This change results in considering disabled ASGs
and deleted ASGs as equivalent, and if we try to disable an ASG that does
not exist it will not freak out.

@doctoryes


doctoryes:
👍

doctoryes:
https://openedx.atlassian.net/browse/TE-1466

PR #49, opened 2016-06-29
We don't care about deletion when trying to determine clusters.

feanil:
In the deploy function we get the list of ASGs so that we can deterimen
relevant clusters to deploy to.  We don't care if some of the existing
ASGs in that cluster are pending deletes.

https://openedx.atlassian.net/browse/TE-1466

### DIFF
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -389,7 +389,7 @@ def deploy(ami_id):
     edp = ec2.edp_for_ami(ami_id)
 
     # These are all autoscaling groups that match the tags we care about.
-    asgs = ec2.asgs_for_edp(edp)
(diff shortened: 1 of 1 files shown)


doctoryes:
👍  I agree this is necessary while the active ASGs have a pending deletion tag on them.

macdiesel:
Makes sense, have we tried deploying with it yet?

If we've run a test deploy and it works 👍

feanil:
Tested against loadtest-edx-progarms

https://gocd.tools.edx.org/go/pipelines/value_stream_map/feanil-test-deploy-edx-ami/13(Before new code)
https://gocd.tools.edx.org/go/pipelines/value_stream_map/feanil-test-deploy-edx-ami/14(After this code)

PR #51, opened 2016-07-25
Added in scripts to clear varnish cache and deploy code to stage or prod

Dillon-Dumesnil:
@edx/ecommerce FYI
https://openedx.atlassian.net/browse/ECOM-4919
https://openedx.atlassian.net/browse/ECOM-4920

### DIFF
--- a/scripts/drupal_backup_database.py
+++ b/scripts/drupal_backup_database.py
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+import click
+from tubular import drupal
+
+
+@click.command()
+@click.option("--env", help="The environment the database backup will take place in.", type=str, required=True)
+@click.option("--username", help="The Acquia username necessary to run the command.", type=str, required=True)
+@click.option("--password", help="The Acquia password necessary to run the command.", type=str, required=True)
+def backup_database(env, username, password):
+    """
+    Creates a backup of the database in the specified environment.
+
+    Args:
+        env (str): The environment the database backup will take place in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+    """
+    drupal.backup_database(env, username, password)
+
+if __name__ == "__main__":
+    backup_database()
--- a/scripts/drupal_clear_varnish.py
+++ b/scripts/drupal_clear_varnish.py
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+import click
+from tubular import drupal
+
+
+@click.command()
+@click.option("--env", help="The environment to clear varnish caches in.", type=str, required=True)
+@click.option("--username", help="The Acquia username necessary to run the command.", type=str, required=True)
+@click.option("--password", help="The Acquia password necessary to run the command.", type=str, required=True)
+def clear_varnish_cache(env, username, password):
+    """
+    Clears the Varnish cache from a Drupal domain.
+
+    Args:
+        env (str): The environment to clear varnish caches in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+    """
+    drupal.clear_varnish_cache(env, username, password)
+
+if __name__ == "__main__":
+    clear_varnish_cache()
--- a/tubular/drupal.py
+++ b/tubular/drupal.py
@@ -1,0 +1,164 @@
+import requests
+import logging
+from requests.auth import HTTPBasicAuth
+from tubular.utils.retry import retry
+from tubular.exception import BackendError
+
+
+ACQUIA_ENDPOINT = "https://cloudapi.acquia.com/v1"
+REALM = "prod"
+SITE = "edx"
+DATABASE = "edx"
+CLEAR_CACHE_URL = "{root}/sites/{realm}:{site}/envs/{{env}}/domains/{{domain}}/cache.json".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE
+)
+DEPLOY_URL = "{root}/sites/{realm}:{site}/envs/{{env}}/code-deploy.json?path=tags%2F{{tag}}".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE
+)
+BACKUP_DATABASE_URL = "{root}/sites/{realm}:{site}/envs/{{env}}/dbs/{database}/backups.json".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, database=DATABASE
+)
+CHECK_TASKS_URL = "{root}/sites/{realm}:{site}/tasks/{{id}}.json".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE
+)
+# Maps environments to domains.
+VALID_ENVIRONMENTS = {
+    "test": "edxstg.prod.acquia-sites.com",
+    "prod": "edx.prod.acquia-sites.com",
+}
+logger = logging.getLogger(__name__)
+
+
+def get_api_client(username, password):
+    """
+    Creates an API Client and authenticates the client.
+
+    Args:
+        username (str): The username used to authenticate the client
+        password (str): The password used to authenticate the client
+
+    Returns:
+        The authenticated API Client
+    """
+    api_client = requests.Session()
+    api_client.auth = HTTPBasicAuth(username, password)
+    return api_client
+
+
+def parse_response(response, error_message):
+    """
+    Parses the response.
+
+    Args:
+        response (requests.Response):
+        error_message (str):
+
+    Returns:
+        The JSON representation of the response if no errors.
+
+    Raises:
+        BackendError: Raised if the response's status code is not 200.
+    """
+    if response.status_code != 200:
+        msg = "{specific}\nStatus Code: {status}\nBody: {body}".format(specific=error_message,
+                                                                       status=response.status_code, body=response.text)
+        logger.error(msg)
+        raise BackendError(msg)
+    return response.json()
+
+
+@retry()
+def clear_varnish_cache(env, username, password):
+    """
+    Clears the Varnish cache from a Drupal domain.
+
+    Args:
+        env (str): The environment to clear varnish caches in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+
+    Returns:
+        True if the Varnish cache is successfully cleared.
+
+    Raises:
+        KeyError: Raised if env value is invalid.
+    """
+    api_client = get_api_client(username, password)
+    domain = VALID_ENVIRONMENTS[env]
+    response = api_client.delete(CLEAR_CACHE_URL.format(env=env, domain=domain))
+    response_json = parse_response(response, "Failed to clear cache.")
+    return check_state(response_json["id"], username, password)
+
+
+@retry()
+def deploy(env, username, password, tag):
+    """
+    Deploys a given tag to the specified environment.
+
+    Args:
+        env (str): The environment to deploy code in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+        tag (str): The tag to deploy to the specified environment.
+
+    Returns:
+        True if the code is successfully deployed.
+
+    Raises:
+        KeyError: Raised if env value is invalid.
+    """
+    VALID_ENVIRONMENTS[env]
+    api_client = get_api_client(username, password)
+    response = api_client.post(DEPLOY_URL.format(env=env, tag=tag))
+    response_json = parse_response(response, "Failed to deploy code.")
+    return check_state(response_json["id"], username, password)
+
+
+@retry()
+def backup_database(env, username, password):
+    """
+    Creates a backup of the database in the specified environment.
+
+    Args:
+        env (str): The environment the database backup will take place in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+
+    Returns:
+        True if a database backup is successfully created.
+
+    Raises:
+        KeyError: Raised if env value is invalid.
+    """
+    VALID_ENVIRONMENTS[env]
+    api_client = get_api_client(username, password)
+    response = api_client.post(BACKUP_DATABASE_URL.format(env=env))
+    response_json = parse_response(response, "Failed to backup database.")
+    return check_state(response_json["id"], username, password)
+
+
+@retry(attempts=10, delay_seconds=10, max_time_seconds=100)
+def check_state(id, username, password):
+    """
+    Checks the state of the response to verify it is "done"
+
+    Args:
+        id (int): The task id to check the state of.
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+
+    Returns:
+        True if state of the response is "done"
+
+    Raises:
+        BackendError: Raised so the method will retry since immediately after receiving the
+            response, the state will still be "waiting". Can"t rely on parse_response since
+            the response should return a 200, just not the state wanted.
+    """
+    api_client = get_api_client(username, password)
+    response = api_client.get(CHECK_TASKS_URL.format(id=id))
+    response_json = parse_response(response, "Failed to check state of response.")
+    if response_json["state"] == "done":
+        return True
+    raise BackendError("Check state failed. The state of the response was {state}, not done as expected.".format(
+        state=response_json["state"]))
--- a/tubular/tests/test_drupal.py
+++ b/tubular/tests/test_drupal.py
@@ -1,0 +1,282 @@
+import os
+import httpretty
+import unittest
+import tubular.drupal as drupal
+from tubular.exception import BackendError
+
+os.environ["TUBULAR_RETRY_ENABLED"] = "false"
+reload(drupal)
+
+clear_cache_response_waiting = """
+{
+"id":"1",
+"queue":"purge-domain",
+"state":"waiting",
+"created":"1469543243",
+"started":null,
+"percentage":null,
+"completed":null,
+"sender":"test@edx.org",
+"recipient":null,
+"result":null,
+"hidden":"0",
+"cookie":null,
+"received":null,
+"tags":[]
+}
+"""
+
+clear_cache_response_done = """
+{
+"id":"1",
+"queue":"purge-domain",
+"state":"done",
+"description":"Clear web cache",
+"created":"1469481715",
+"started":"1469481715",
+"completed":"1469481716",
+"sender":"test@edx.org",
+"result":"",
+"cookie":null,
+"logs":"[21:21:55] [21:21:55] Started\\n[21:21:56] [2016-07-25 21:21:56] Cleared domain cache\\n"
+}
+"""
+
+deploy_response_waiting = """
+{
+"id":"2",
+"queue":"code-push",
+"state":"waiting",
+"description":"Deploy code to extra",
+"created":"1469542531",
+"started":null,
+"percentage":null,
+"completed":null,
+"sender":"test@edx.org",
+"recipient":null,
+"result":null,
+"hidden":"0",
+"cookie":null,
+"received":null,
+"tags":[]
+}
+"""
+
+deploy_response_done = """
+{
+"id":"2",
+"queue":"code-push",
+"state":"done",
+"description":"Deploy code to extra",
+"created":"1469542531",
+"started":"1469542531",
+"completed":"1469542534",
+"sender":"test@edx.org",
+"result":"",
+"cookie":null,
+"logs":"[14:15:31] [14:15:31] Started\\n[14:15:34] [2016-07-26 14:15:32] updating web_servers[staging-8744].\\nUpdating to deploy tag\\nDeploying tag on edx\\n[2016-07-26 14:15:34] Starting hook: post-code-deploy\\n[2016-07-26 14:15:34] Finished hook: post-code-deploy\\n"
+}
+"""
+
+backup_database_response_waiting = """
+{
+"id":"3",
+"queue":"create-db-backup-ondemand",
+"state":"waiting",
+"description":"Backup database in extra.",
+"created":"1469721806",
+"started":null,
+"percentage":null,
+"completed":null,
+"sender":"test@edx.org",
+"recipient":null,
+"result":null,
+"hidden":"0",
+"cookie":null,
+"received":null,
+"tags":[]
+}
+"""
+
+backup_database_response_started = """
+{
+"id":"3",
+"queue":"create-db-backup-ondemand",
+"state":"started",
+"description":"Backup database in extra.",
+"created":"1469721806",
+"started":"1469721807",
+"completed":null,
+"sender":"test@edx.org",
+"result":null,
+"cookie":null,
+"logs":"[16:03:27] [16:03:27] Started\\n"
+}
+"""
+
+backup_database_response_done = """
+{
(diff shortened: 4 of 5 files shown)


scripts/drupal_clear_varnish.py
```diff
@@ -0,0 +1,38 @@
...
+        env: The environment to clear varnish caches in (e.g. test or prod)
+        username: The Acquia username necessary to run the command.
+        password: The Acquia password necessary to run the command.
+    """
+    api_client = requests.Session()
+    # Need to authenticate the user so that Acquia will complete the request.
+    api_client.auth = HTTPBasicAuth(username, password)
+    # Domain to clear varnish on is different for stage (test) and prod
+    if env == 'test':
+        api_client.delete('{root}/sites/{realm}:{site}/envs/{env}/domains/{domain}/cache.json'.format(
```
clintonb: Don't call this code twice. You only need an if-statement to define the domain variable.

scripts/drupal_clear_varnish.py
```diff
@@ -0,0 +1,38 @@
...
+            root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, env=env, domain='edxstg.prod.acquia-sites.com'
+        ))
+    elif env == 'prod':
+        api_client.delete('{root}/sites/{realm}:{site}/envs/{env}/domains/{domain}/cache.json'.format(
+            root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, env=env, domain='edx.prod.acquia-sites.com'
+        ))
+
+if __name__ == "__main__":
+    clear_varnish_cache()
```
clintonb: You need to pass arguments to the function.

scripts/drupal_clear_varnish.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Clears the Varnish cache from a Drupal domain.
+
+    Args:
+        env: The environment to clear varnish caches in (e.g. test or prod)
+        username: The Acquia username necessary to run the command.
+        password: The Acquia password necessary to run the command.
+    """
+    api_client = requests.Session()
+    # Need to authenticate the user so that Acquia will complete the request.
```
clintonb: Remove this and the similar comment below. No need to add comments for obvious actions.

clintonb:
👍 (once comments addressed)

scripts/drupal_clear_varnish.py
Dillon-Dumesnil: Based on the other scripts living in the repository, it is okay to leave the function call under main with no arguments and they are filled in when the script is called.

tubular/drupal.py
```diff
@@ -0,0 +1,45 @@
...
+        env (str): The environment to deploy code in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+        tag (str): The tag to deploy to the specified environment.
+    """
+    api_client = requests.Session()
+    api_client.auth = HTTPBasicAuth(username, password)
+    api_client.post('{root}/sites/{realm}:{site}/envs/{env}/code-deploy.json?path=tags%2F{tag}'.format(
+        root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, env=env, tag=tag
+    ))
```
doctoryes: The `retry` decorator works with catching exceptions - but shouldn't these `api_client` calls also check the HTTP return code? Are there failures that won't raise an exception?

tubular/drupal.py
```diff
@@ -0,0 +1,45 @@
...
+    Args:
+        env (str): The environment to clear varnish caches in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+    """
+    api_client = requests.Session()
+    api_client.auth = HTTPBasicAuth(username, password)
+    if env == 'test':
+        domain = 'edxstg.prod.acquia-sites.com'
+    elif env == 'prod':
```
clintonb: What if env is not one of these two values? Raise an appropriate exception so that there are no surprises.

tubular/drupal.py
Dillon-Dumesnil: Will be adding in checking return codes now.

tubular/drupal.py
Dillon-Dumesnil: Will do.

Dillon-Dumesnil:
@edx/pipeline-team Please review changes.
@clintonb Please review changes.

tubular/drupal.py
clintonb: This needs to be addressed.

Dillon-Dumesnil: Done. Added in assert statement in beginning of the method

tubular/drupal.py
```diff
@@ -0,0 +1,137 @@
...
+
+    api_client = requests.Session()
+    api_client.auth = HTTPBasicAuth(username, password)
+    if env == 'test':
+        domain = 'edxstg.prod.acquia-sites.com'
+    elif env == 'prod':
+        domain = 'edx.prod.acquia-sites.com'
+    response = api_client.delete(CLEAR_CACHE_URL.format(env=env, domain=domain))
+
+    if response.status_code == 401:
```
clintonb: You don't seem to actually be doing anything with the raised errors. If this is true, all you really care about is if the status is 200. If not, it doesn't really matter how the call failed (from an execution standpoint). In the failure case, log the response status code and body before raising an error. The error message can simply be "Failed to clear cache."

tubular/drupal.py
```diff
@@ -0,0 +1,137 @@
...
+            Explicitly looks for any status codes >= 500.
+    """
+    assert (env == 'test' or env == 'prod'), \
+        'Site environment ({env}) is not supported. Only test and prod'.format(env=env)
+
+    api_client = requests.Session()
+    api_client.auth = HTTPBasicAuth(username, password)
+    response = api_client.post(DEPLOY_URL.format(env=env, tag=tag))
+
+    if response.status_code == 401:
```
clintonb: Same comment as above regarding error handling.

tubular/drupal.py
```diff
@@ -0,0 +1,137 @@
...
+        id (int): The task id to check the state of.
+
+    Raises:
+        DrupalError: DrupalError is raised to capture 401 and 404 status codes.
+            Each is accompanied with a messaging specifying 401 or 404.
+            Also raised if the check on the response's state fails.
+        BackendError: Raised to capture any status codes that are not 401, 404, or 200.
+            Explicitly looks for any status codes >= 500.
+    """
+    api_client = requests.Session()
```
clintonb: These two lines are repeated multiple times. Move them to a helper function (e.g. `get_api_client`).

tubular/drupal.py
```diff
@@ -0,0 +1,137 @@
...
+    api_client.auth = HTTPBasicAuth(username, password)
+    response = api_client.get(CHECK_TASKS_URL.format(id=id))
+
+    if response.status_code == 401:
```
clintonb: Same as above regarding error handling.

clintonb: In fact, this might be another place for a helper function (e.g. `parse_response`) that takes the response, raises an error if necessary, and returns the JSON in the success case.

tubular/exception.py
```diff
@@ -26,3 +26,6 @@ class ASGDoesNotExistException(Exception):
 
 class CannotDeleteActiveASG(Exception):
     pass
+
+class DrupalError(Exception):
```
clintonb: Since you aren't actually doing anything with the raised exceptions, there is not need for a custom exception class.

tubular/tests/test_drupal.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @httpretty.activate
+    def test_clear_caches_401(self):
+        """
+        Tests clear_varnish_cache raises DrupalError when status code is 401.
+        """
+        httpretty.register_uri(
+            httpretty.DELETE,
+            drupal.CLEAR_CACHE_URL.format(env=ACQUIA_ENV, domain=ACQUIA_DOMAIN),
+            body=status_401_response,
```
clintonb: Since you're not actually using the body, you can simply pass `'{}'`.

clintonb: The same applies to pretty much all of these.

tubular/tests/test_drupal.py
```diff
@@ -0,0 +1,200 @@
...
+            content_type="application/json",
+            status=401
+        )
+        with self.assertRaises(DrupalError):
+            drupal.clear_varnish_cache(env="test", username="foo", password="bar")
```
clintonb: `env=ACQUIA_ENV` here and everywhere else.

clintonb: The username and password should be constants here and everywhere else in this file.

tubular/drupal.py
```diff
@@ -0,0 +1,137 @@
...
+    Returns:
+        True if the Varnish cache is successfully cleared. Raises DrupalError (from check_status) if failed.
+
+    Raises:
+        DrupalError: DrupalError is raised to capture 401 and 404 status codes.
+            Each is accompanied with a messaging specifying 401 or 404.
+        BackendError: Raised to capture any status codes that are not 401, 404, or 200.
+            Explicitly looks for any status codes >= 500.
+    """
+    assert (env == 'test' or env == 'prod'), \
```
clintonb: Add a test.

clintonb: This logic can be simplified: `(env in ('test', 'prod',))`

tubular/tests/test_drupal.py
Dillon-Dumesnil: So I actually need the state and id fields of the body for any tests not dealing with errors. Is it preferable to feed in the body how I am or to switch it passing in "{'id': '1', 'state': 'waiting'}" or "{'id': '1', 'state': 'done'}" based on what I need? The rest of the body is unused so that's really the only field that matters.

clintonb: If you need a real body for successful cases, use one. For these other cases, use an empty body. This provides a hint to future developers that the body is irrelevant.

Dillon-Dumesnil:
@clintonb Changes have been addressed. Please see updated code

tubular/drupal.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Raises:
+        StandardError: Since no action will be taken if there is an error, StandardError is raised
+            just to let the user know there was an error. Also activates the retry in the calling method.
+    """
+    if response.status_code != 200:
+        print "Response failed to succeed with status code: {status} and body: {body}".format(
+            status=response.status_code, body=response.text
+        )
+        raise StandardError(error_message)
```
macdiesel: If we must used a built in exception, could we just use Exception please? I would prefer BackEndError.

tubular/tests/test_drupal.py
```diff
@@ -0,0 +1,188 @@
...
+            drupal.CLEAR_CACHE_URL.format(env=ACQUIA_ENV, domain=ACQUIA_DOMAIN),
+            body="{}",
+            content_type="application/json",
+            status=401
+        )
+        with self.assertRaises(StandardError):
+            drupal.clear_varnish_cache(env=ACQUIA_ENV, username=TEST_USERNAME, password=TEST_PASSWORD)
+
+    @httpretty.activate
+    def test_clear_caches_success(self):
```
macdiesel: def test_clear_varnish_cache_success

tubular/tests/test_drupal.py
```diff
@@ -0,0 +1,188 @@
...
+        httpretty.register_uri(
+            httpretty.GET,
+            drupal.CHECK_TASKS_URL.format(id="1"),
+            body=valid_clear_cache_response_done,
+            content_type="application/json"
+        )
+        self.assertTrue(drupal.check_state(id="1", username=TEST_USERNAME, password=TEST_PASSWORD))
+
+    @httpretty.activate
+    def test_clear_caches_failure(self):
```
macdiesel: def test_clear_varnish_cache_failure

tubular/tests/test_drupal.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+        httpretty.register_uri(
+            httpretty.GET,
+            drupal.CHECK_TASKS_URL.format(id="2"),
+            body=valid_deploy_response_done,
+            content_type="application/json"
+        )
+        self.assertTrue(drupal.deploy(env=ACQUIA_ENV, username=TEST_USERNAME, password=TEST_PASSWORD, tag=TEST_TAG))
+
+    def test_invalid_environment(self):
```
macdiesel: def test_clear_varnish_cache_invalid_environment

tubular/drupal.py
```diff
@@ -0,0 +1,125 @@
...
+    Args:
+        env (str): The environment to deploy code in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+        tag (str): The tag to deploy to the specified environment.
+
+    Returns:
+        True if the code is successfully deployed.
+    """
+    assert (env in ('test', 'prod',)), 'Site environment ({env}) is not supported. Only test and prod'.format(env=env)
```
macdiesel: This is done a few times, pull it out in to a function, write a test.

Alternatively you could do something like this at the top of the file:

```
environments = 
{
    'test': 'edxstg.prod.acquia-sites.com',
    'prod': 'edx.prod.acquia-sites.com'
}
```

Then when you need to get an environment:

```
environments[env]
```

will raise a key error if the environment is not defined.

though the message on this is probably a bit more user friendly.

macdiesel:
@Dillon-Dumesnil pass complete.

tubular/drupal.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Returns:
+        The JSON representation of the response if no errors.
+
+    Raises:
+        StandardError: Since no action will be taken if there is an error, StandardError is raised
+            just to let the user know there was an error. Also activates the retry in the calling method.
+    """
+    if response.status_code != 200:
+        print "Response failed to succeed with status code: {status} and body: {body}".format(
```
clintonb: Use `logger.error`. See https://docs.python.org/2/library/logging.html.

`print` is bad in environments where there is no console.

clintonb: Requested failed with status code...

clintonb: You should incorporate the error message...

`msg = '{specific}\nStatus Code: {status}\nBody: {body}'.format(specific=error_message, status=response.status_code, body=response.text)`

tubular/drupal.py
```diff
@@ -0,0 +1,125 @@
...
+        StandardError: Raised so the method will retry since immediately after receiving the
+            response, the state will still be "waiting". Can't rely on parse_response since
+            the response should return a 200, just not the state wanted.
+    """
+    api_client = get_api_client(username, password)
+    response = api_client.get(CHECK_TASKS_URL.format(id=id))
+    response_json = parse_response(response, "Failed to check state of response.")
+    if response_json['state'] == 'done':
+        return True
+    raise StandardError
```
clintonb: You should always include a message when you raise an exception. This makes it easier to debug later without necessarily having to dig into the code.

tubular/tests/test_drupal.py
```diff
@@ -0,0 +1,188 @@
...
+"completed":"1469542534",
+"sender":"test@edx.org",
+"result":"",
+"cookie":null,
+"logs":"[14:15:31] [14:15:31] Started\\n[14:15:34] [2016-07-26 14:15:32] updating web_servers[staging-8744].\\nUpdating to deploy tag\\nDeploying tag on edx\\n[2016-07-26 14:15:34] Starting hook: post-code-deploy\\n[2016-07-26 14:15:34] Finished hook: post-code-deploy\\n"
+}
+"""
+
+ACQUIA_ENV = "test"
+ACQUIA_DOMAIN = 'edxstg.prod.acquia-sites.com'
```
clintonb: Single quotes or double quotes. Pick one and use it throughout your code.

Dillon-Dumesnil:
@clintonb @macdiesel Changes have been integrated. Let me know

tubular/drupal.py
```diff
@@ -0,0 +1,133 @@
...
+    Args:
+        env (str): The environment to deploy code in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+        tag (str): The tag to deploy to the specified environment.
+
+    Returns:
+        True if the code is successfully deployed.
+    """
+    # assert (env in ("test", "prod",)), "Site environment ({env}) is not supported. Only test and prod".format(env=env)
```
macdiesel: You should remove this if you are not going to use it.

tubular/drupal.py
```diff
@@ -0,0 +1,133 @@
...
+CLEAR_CACHE_URL = "{root}/sites/{realm}:{site}/envs/{env}/domains/{domain}/cache.json".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, env="{env}", domain="{domain}"
+)
+DEPLOY_URL = "{root}/sites/{realm}:{site}/envs/{env}/code-deploy.json?path=tags%2F{tag}".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, env="{env}", tag="{tag}"
+)
+CHECK_TASKS_URL = "{root}/sites/{realm}:{site}/tasks/{id}.json".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, id="{id}"
+)
+# Maps environments to domains. Could also include databases by turning values into arrays.
```
clintonb: `Could also include databases by turning values into arrays.`

If you aren't going to do it, ditch the comment. It will only confuse others.

tubular/drupal.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    # assert (env in ("test", "prod",)), "Site environment ({env}) is not supported. Only test and prod".format(env=env)
+    VALID_ENVIRONMENTS[env]
+    api_client = get_api_client(username, password)
+    response = api_client.post(DEPLOY_URL.format(env=env, tag=tag))
+    response_json = parse_response(response, "Failed to deploy code.")
+    return check_state(response_json["id"], username, password)
+
+
+@retry(attempts=10, delay_seconds=5, max_time_seconds=50)
```
macdiesel: @Dillon-Dumesnil @clintonb or anyone else who has deployed this app before, any idea how long it typically takes for the site to come up as done after a deployment has been initiated?

clintonb: It usually takes less than 30 seconds.

tubular/drupal.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Raises:
+        BackendError: Since no action will be taken if there is an error, BackendError is raised
+            just to let the user know there was an error. Also activates the retry in the calling method.
+    """
+    if response.status_code != 200:
+        msg = "{specific}\nStatus Code: {status}\nBody: {body}".format(specific=error_message,
+                                                                       status=response.status_code, body=response.text)
+        logger.error(msg)
+        raise BackendError(error_message)
```
clintonb: Use the `msg` value.

tubular/drupal.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Args:
+        response (requests.Response):
+        error_message (str):
+
+    Returns:
+        The JSON representation of the response if no errors.
+
+    Raises:
+        BackendError: Since no action will be taken if there is an error, BackendError is raised
```
clintonb: `BackendError: Raised if the response's status code is not 200.`

This one function doesn't care about retry logic, so there shouldn't be a mention of it in its docstring.

tubular/drupal.py
```diff
@@ -0,0 +1,133 @@
...
+    Args:
+        env (str): The environment to clear varnish caches in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+
+    Returns:
+        True if the Varnish cache is successfully cleared.
+    """
+    api_client = get_api_client(username, password)
+    domain = VALID_ENVIRONMENTS[env]
```
clintonb: Document the fact that this will raise a `KeyError` if the `env` value is invalid. Do the same for the call in the `deploy` function, too.

clintonb:
👍 (once comments addressed)

macdiesel:
👍 🎈

tubular/drupal.py
Dillon-Dumesnil: Based on what I saw with the curl command tests I was running, it would go from waiting to done in roughly 3-4 seconds which is why I have it set at 5 seconds, but also allowing for 10 retries in the craziest of cases. Can always be easily changed if desired.

Dillon-Dumesnil:
Version 2.0. All comments have been addressed. @clintonb @macdiesel

clintonb:
Remember to squash your commits before merging. I'm not seeing a need for this to be three separate commits.

Dillon-Dumesnil:
@clintonb @macdiesel If you could go over this one last time. Added in backing up database script. Tests have been added. Should just take a minute. Most of the code is very similar to the other scripts.

tubular/drupal.py
```diff
@@ -0,0 +1,164 @@
...
+from tubular.exception import BackendError
+
+
+ACQUIA_ENDPOINT = "https://cloudapi.acquia.com/v1"
+REALM = "prod"
+SITE = "edx"
+DATABASE = "edx"
+CLEAR_CACHE_URL = "{root}/sites/{realm}:{site}/envs/{env}/domains/{domain}/cache.json".format(
+    root=ACQUIA_ENDPOINT, realm=REALM, site=SITE, env="{env}", domain="{domain}"
+)
```
macdiesel: @Dillon-Dumesnil for these strings instead of passing back in env and domain that you don't want to have interpolated here, just escape them instead of passing them back in:

```
"{root}/sites/{realm}:{site}/envs/{{env}}/domains/{{domain}}/cache.json".format(
     root=ACQUIA_ENDPOINT, realm=REALM, site=SITE
 )
```

Dillon-Dumesnil: Didn't know you could do that. Will make that change.

scripts/drupal_backup_database.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Args:
+        env (str): The environment the database backup will take place in (e.g. test or prod)
+        username (str): The Acquia username necessary to run the command.
+        password (str): The Acquia password necessary to run the command.
+    """
+    drupal.backup_database(env, username, password)
+
+if __name__ == "__main__":
+    backup_database()
```
clintonb, line 22: Add a newline.

Dillon-Dumesnil, line 22: Done

Dillon-Dumesnil:
Comments addressed

tubular/drupal.py
```diff
@@ -0,0 +1,164 @@
...
+        KeyError: Raised if env value is invalid.
+    """
+    VALID_ENVIRONMENTS[env]
+    api_client = get_api_client(username, password)
+    response = api_client.post(BACKUP_DATABASE_URL.format(env=env))
+    response_json = parse_response(response, "Failed to backup database.")
+    return check_state(response_json["id"], username, password)
+
+
+@retry(attempts=10, delay_seconds=10, max_time_seconds=100)
```
macdiesel, line 140: @Dillon-Dumesnil I see you adjusted this up, I assume this is due to the length of the database backup?

Dillon-Dumesnil, line 140: @macdiesel I actually adjusted it from Clinton's comment about the deploy taking him ~30 seconds. I realized I was running my manual verification on the environment extra which is potentially a lot lighter of a load than stage or prod. I think 100 seconds should be enough to capture any process, but can always change if the pipeline is timing out.

macdiesel:
👍 🚢 🇮🇹

PR #58, opened 2016-08-09
Changed clear varnish cache to clear all domains in the environments

Dillon-Dumesnil:
@edx/pipeline-team @clintonb Please review

### DIFF
--- a/tubular/drupal.py
+++ b/tubular/drupal.py
@@ -26,8 +26,21 @@ CHECK_TASKS_URL = "{root}/sites/{realm}:{site}/tasks/{{id}}.json".format(
 )
 # Maps environments to domains.
 VALID_ENVIRONMENTS = {
-    "test": "edxstg.prod.acquia-sites.com",
-    "prod": "edx.prod.acquia-sites.com",
+    "test": [
+        "edxstg.prod.acquia-sites.com",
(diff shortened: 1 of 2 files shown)


clintonb:
👍

tubular/drupal.py
```diff
@@ -100,24 +113,26 @@ def fetch_deployed_tag(env, username, password, path_name):
...
     Raises:
         KeyError: Raised if env value is invalid.
     """
     api_client = get_api_client(username, password)
-    domain = VALID_ENVIRONMENTS[env]
-    response = api_client.delete(CLEAR_CACHE_URL.format(env=env, domain=domain))
-    response_json = parse_response(response, "Failed to clear cache.")
-    return check_state(response_json["id"], username, password)
+    domains = VALID_ENVIRONMENTS[env]
+    for domain in domains:
```
macdiesel: Just so I am clear, if one domain fails, the remaining domains should not be cleared? Would it be better here to continue and try to clear as many domains as possible?

Perhaps attempt them all and still raise an exception listing those that failed? 

My concern here is that you could leave these caches in a partially cleared state.

Dillon-Dumesnil:
@macdiesel Updated to continue through all domains despite an error and to raise an error at the end.

macdiesel:
👍

PR #60, opened 2016-09-12
Do not raise exception when trying to tag an ASG that does not exist

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/tubular/ec2.py
+++ b/tubular/ec2.py
@@ -157,15 +157,13 @@ def tag_asg_for_deletion(asg_name, seconds_until_delete_delta=1800):
 
     Returns:
         None
-
-    Raises:
-        ASGDoesNotExistException: if the Autoscale group does not exist
     """
     tag = create_tag_for_asg_deletion(asg_name, seconds_until_delete_delta)
     autoscale = boto.connect_autoscale()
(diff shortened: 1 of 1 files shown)


tubular/ec2.py
```diff
@@ -157,15 +157,13 @@ def tag_asg_for_deletion(asg_name, seconds_until_delete_delta=1800):
...
-
-    Raises:
-        ASGDoesNotExistException: if the Autoscale group does not exist
     """
     tag = create_tag_for_asg_deletion(asg_name, seconds_until_delete_delta)
     autoscale = boto.connect_autoscale()
-    if len(autoscale.get_all_groups([asg_name])) < 1:
-        raise ASGDoesNotExistException("Could not apply tags to Autoscale group: {0} does not exist.".format(asg_name))
-    autoscale.create_or_update_tags([tag])
+    if len(autoscale.get_all_groups([asg_name])) > 0:
```
doctoryes: Is this condition correct? Shouldn't it be `< 1` for the ASG to no longer exist?

doctoryes: And, if so, why did tests pass?

doctoryes:
👍

PR #61, opened 2016-09-12
Improved error messaging when an AMI is not found during EDC lookup

macdiesel:
@edx/pipeline-team PTAL.


tubular/ec2.py
```diff
@@ -40,6 +40,8 @@ def edp_for_ami(ami_id):
 
     try:
         ami = ec2.get_all_images(ami_id)[0]
+    except IndexError as error:
```
doctoryes: Nit: Since you're not using `error`:
`except IndexError:` ?

doctoryes:
👍

PR #67, opened 2016-09-21
Add rollback functionality to asgard module.

doctoryes:
Add click script to rollback to ASGs. Add tests for rollback and failed rollback.

### DIFF
--- a/scripts/rollback_asg.py
+++ b/scripts/rollback_asg.py
@@ -16,9 +16,10 @@ logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
 
 @click.command()
-@click.option('--config-file', envvar='CONFIG_FILE', help='The config file to to get the ami_id from.')
-@click.option('--dry-run', envvar='DRY_RUN', help='Don\'t actually deploy.', is_flag=True, default=False)
-def rollback(config_file, dry_run):
+@click.option('--config_file', envvar='CONFIG_FILE', help='The config file from which to get the previous deploy information.')
+@click.option('--dry_run', envvar='DRY_RUN', help='Don\'t actually rollback.', is_flag=True, default=False)
+@click.option('--out_file', envvar='OUT_FILE', help='Output file for the YAML rollback information.', default=None)
+def rollback(config_file, dry_run, out_file):
     """
     Roll back to an existing ASG. If the desired ASG(s) are not available to roll back (have since been deleted)
     and an AMI_ID is specified a new ASG using that AMI ID will be used.
@@ -39,10 +40,6 @@ def rollback(config_file, dry_run):
     ami_id: ami-a1b1c1d0
 
     The disabled_asgs will be enabled and the current_asgs will be disabled.
-
-    Args:
-        config_file:
-        dry_run:
     """
     config = yaml.safe_load(open(config_file, 'r'))
     current_asgs = config['current_asgs']
@@ -51,16 +48,16 @@ def rollback(config_file, dry_run):
 
     try:
         if not dry_run:
-            deploy_info = asgard.rollback(current_asgs, disabled_asgs, ami_id)
+            rollback_info = asgard.rollback(current_asgs, disabled_asgs, ami_id)
         else:
             click.echo('Would have triggered a rollback of {}'.format(ami_id))
-            deploy_info = {}
+            rollback_info = {}
 
         if out_file:
             with open(out_file, 'w') as stream:
-                yaml.safe_dump(deploy_info, stream, default_flow_style=False, explicit_start=True)
+                yaml.safe_dump(rollback_info, stream, default_flow_style=False, explicit_start=True)
         else:
-            print yaml.dump(deploy_info, default_flow_style=False, explicit_start=True)
+            print yaml.safe_dump(rollback_info, default_flow_style=False, explicit_start=True)
 
     except Exception as e:
         traceback.print_exc()
--- a/scripts/rollback_asg.py
+++ b/scripts/rollback_asg.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python
+import sys
+import logging
+import traceback
+import click
+import yaml
+from os import path
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append( path.dirname( path.dirname( path.abspath(__file__) ) ) )
+
+from tubular import asgard
+
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+
+@click.command()
+@click.option('--config-file', envvar='CONFIG_FILE', help='The config file to to get the ami_id from.')
+@click.option('--dry-run', envvar='DRY_RUN', help='Don\'t actually deploy.', is_flag=True, default=False)
+def rollback(config_file, dry_run):
+    """
+    Roll back to an existing ASG. If the desired ASG(s) are not available to roll back (have since been deleted)
+    and an AMI_ID is specified a new ASG using that AMI ID will be used.
+
+    Configuration file input:
+    ---
+    current_asgs:
+      loadtest-edx-edxapp:
+        - loadtest-edx-edxapp-v099
+      loadtest-edx-worker:
+        - loadtest-edx-worker-v099
+    disabled_asgs:
+      loadtest-edx-edxapp:
+        - loadtest-edx-edxapp-v058
+        - loadtest-edx-edxapp-v059
+      loadtest-edx-worker:
+        - loadtest-edx-worker-v034
+    ami_id: ami-a1b1c1d0
+
+    The disabled_asgs will be enabled and the current_asgs will be disabled.
+
+    Args:
+        config_file:
+        dry_run:
+    """
+    config = yaml.safe_load(open(config_file, 'r'))
+    current_asgs = config['current_asgs']
+    disabled_asgs = config['disabled_asgs']
+    ami_id = config['ami_id']
+
+    try:
+        if not dry_run:
+            deploy_info = asgard.rollback(current_asgs, disabled_asgs, ami_id)
+        else:
+            click.echo('Would have triggered a rollback of {}'.format(ami_id))
+            deploy_info = {}
+
+        if out_file:
+            with open(out_file, 'w') as stream:
+                yaml.safe_dump(deploy_info, stream, default_flow_style=False, explicit_start=True)
+        else:
+            print yaml.dump(deploy_info, default_flow_style=False, explicit_start=True)
+
+    except Exception as e:
+        traceback.print_exc()
+        click.secho('Error rolling back AMI: {0}.\nMessage: {1}'.format(ami_id, e.message), fg='red')
+        sys.exit(1)
+
+    sys.exit(0)
+
+if __name__ == "__main__":
+    rollback()
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -20,7 +20,7 @@ from tubular.exception import (
         TimeoutException,
         ClusterDoesNotExistException,
 )
-from utils import WAIT_SLEEP_TIME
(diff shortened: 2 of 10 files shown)


tubular/asgard.py
```diff
@@ -468,6 +466,32 @@ def elbs_for_asg(asg):
...
+    Returns:
+        dict(str, dict): Returns a dictionary with the keys: 'current_asgs' and 'disabled_asgs'
+
+    Raises:
+        TimeoutException: When the task to bring up the new instance times out.
+        BackendError: When the task to bring up the new instance fails.
+        ASGDoesNotExistException: If the ASG being queried does not exist.
+    """
+    try:
+        _red_black_deploy(rollback_to_clustered_asgs, current_clustered_asgs)
```
macdiesel: Before you attempt to rollback you should attempt to remove the delete_on_ts tag from the rollback_to_clustered_asgs. If the janitor runs while the process is in flight it may lead to some strange results. 

I would suggest removing the tag, checking to see if deletion is in progress then proceeding with the _red_black_deploy call.

tubular/asgard.py
```diff
@@ -468,6 +466,32 @@ def elbs_for_asg(asg):
...
+        return {'current_asgs': rollback_to_clustered_asgs, 'disabled_asgs': current_clustered_asgs }
```
macdiesel: Question, prior to this change the current_asgs and disabled_asgs were generated by the code as the deployments actually happened. With this change we will now assume that this is actually the case. Do we want to make that assumption? 

It might be better to let _red_black_deploy actually tell us what it did.

tubular/asgard.py
```diff
@@ -477,7 +501,7 @@ def deploy(ami_id):
         ami_id(str): AWS AMI ID
 
     Returns:
-        dict(str, [str]): Returns a dictionary with the keys: 'current_asgs' and 'disabled_asgs'
+        dict(str, dict): Returns a dictionary with the keys: 'current_asgs' and 'disabled_asgs'
```
macdiesel: Could we add the ami_id to the return dict structure please? This Information is necessary for the rollback call.

tubular/asgard.py
doctoryes: Will do.

tubular/asgard.py
doctoryes: This code was written with the assumption that `_red_black_deploy()` will either succeed completely _or_ will raise an exception. But, after reviewing the code, I see that's not necessarily the case. That means I'll need to move from an exception-raising pattern to a return of (success, asg_clusters_state) in order to return the state properly before raising an error in the caller.

tl;dr - will do.

tubular/asgard.py
doctoryes: Will do.

doctoryes:
@macdiesel Comments are addressed - ready for another review.

scripts/rollback_asg.py
```diff
@@ -0,0 +1,73 @@
...
+            deploy_info = asgard.rollback(current_asgs, disabled_asgs, ami_id)
+        else:
+            click.echo('Would have triggered a rollback of {}'.format(ami_id))
+            deploy_info = {}
+
+        if out_file:
+            with open(out_file, 'w') as stream:
+                yaml.safe_dump(deploy_info, stream, default_flow_style=False, explicit_start=True)
+        else:
+            print yaml.dump(deploy_info, default_flow_style=False, explicit_start=True)
```
macdiesel: safe_dump?

tubular/asgard.py
```diff
@@ -490,78 +545,150 @@ def deploy(ami_id):
...
-    disabled_asg = defaultdict(list)
-    for cluster,asgs in existing_clusters.iteritems():
+    except Exception as wait_fail:
+        LOG.info("Some ASGs are failing ELB health checks. Disabling traffic to all new ASGs.")
+        for cluster, asgs in newly_enabled_asgs.iteritems():
+            for asg in asgs:
+                try:
+                    disable_cluster_asg(cluster, asg)
+                except:
+                    LOG.warning("Unable to disable ASG '{}' after waiting for healthy ELBs.".format(asg))
```
macdiesel: Should this continue here?

tubular/asgard.py
```diff
@@ -490,78 +545,150 @@ def deploy(ami_id):
...
-    for cluster, asg in new_asgs.iteritems():
-        if is_asg_pending_delete(asg) or not is_asg_enabled(asg):
-            raise BackendError("New Autoscale Group {} is pending delete, Aborting the disabling of old ASGs.".format(asg))
-
-    disabled_asg = defaultdict(list)
-    for cluster,asgs in existing_clusters.iteritems():
+    except Exception as wait_fail:
+        LOG.info("Some ASGs are failing ELB health checks. Disabling traffic to all new ASGs.")
+        for cluster, asgs in newly_enabled_asgs.iteritems():
+            for asg in asgs:
```
macdiesel: NIT: We essentially run this logic twice for failures. It might be good to pull it out in to a local method like you did with enable_cluster_asg and disable_cluster_asg.

tubular/asgard.py
```diff
@@ -490,78 +545,150 @@ def deploy(ami_id):
...
+                elbs_to_monitor.extend(elbs_for_asg(asg))
+                newly_enabled_asgs[cluster].append(asg)
+            except Exception:
+                LOG.error("Error enabling ASG '{}'. Disabling traffic to all new ASGs.".format(asg))
+                LOG.error(traceback.format_exc())
+                # Disable the ASG which failed first.
+                disable_cluster_asg(cluster, asg)
+                # Then disable any new other ASGs that have been newly enabled.
+                for cluster, asgs_to_disable in newly_enabled_asgs.iteritems():
+                    for asg in asgs_to_disable:
```
macdiesel: The other place we rollback the ASGs see nit about moving this to method.

tubular/asgard.py
```diff
@@ -490,78 +545,150 @@ def deploy(ami_id):
...
+        LOG.info("Some ASGs are failing ELB health checks. Disabling traffic to all new ASGs.")
+        for cluster, asgs in newly_enabled_asgs.iteritems():
+            for asg in asgs:
+                try:
+                    disable_cluster_asg(cluster, asg)
+                except:
+                    LOG.warning("Unable to disable ASG '{}' after waiting for healthy ELBs.".format(asg))
+        return (False, asgs_enabled, asgs_disabled)
+
+    # Ensure the new ASGs are still healthy and not pending delete before disabling the old ASGs.
```
macdiesel, line 682: Something to consider in the future, perhaps we should have a timer here where we sleep for a (few) minute(s) before disabling the old ASGs? This way if something happens just after they get load that is causing them to fall out of the load balancer the deploy will fail.

@edx/devops any thoughts?

tubular/ec2.py
```diff
@@ -168,6 +168,24 @@ def tag_asg_for_deletion(asg_name, seconds_until_delete_delta=1800):
...
+
+    Returns:
+        None
+    """
+    tag_to_delete = create_tag_for_asg_deletion(asg_name, 0)
+    autoscale = boto.connect_autoscale()
+    if len(autoscale.get_all_groups([asg_name])) < 1:
+        LOG.info("ASG {} no longer exists, will not remove deletion tag.".format(asg_name))
+    else:
+        autoscale.delete_tags([tag_to_delete])
```
macdiesel, line 190: Have you had a chance to try this against the ASG? While looking through the boto 2 docs it seems is the value of the tags don't match they will not be removed, though the documentation for the autoscale class isn't nearly as clear as the other docs.

Autoscale:
http://boto.cloudhackers.com/en/latest/ref/autoscale.html#boto.ec2.autoscale.AutoScaleConnection.delete_tags

Other delete tags:
http://boto.cloudhackers.com/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.delete_tags

```
tags (dict or list) – Either a dictionary containing name/value pairs or a list containing just tag names. If you pass in a dictionary, the values must match the actual tag values or the tag will not be deleted. If you pass in a value of None for the tag value, all tags with that name will be deleted.
```

TLDR: We should probably set the tag value to None.

macdiesel (commented):
@doctoryes Pass complete.

scripts/rollback_asg.py
doctoryes: Will do. This script has other problems which I'll fix as well.

tubular/asgard.py
doctoryes: Will do.

tubular/asgard.py
doctoryes: Will do.

tubular/asgard.py
doctoryes: I decided yes. The code is trying to disable as many ASGs as possible in the list - it shouldn't stop attempting to disable other ASGs if one can't be disabled.

tubular/asgard.py
```diff
@@ -490,78 +545,150 @@ def deploy(ami_id):
...
+                # Disable the ASG which failed first.
+                disable_cluster_asg(cluster, asg)
+                # Then disable any new other ASGs that have been newly enabled.
+                for cluster, asgs_to_disable in newly_enabled_asgs.iteritems():
+                    for asg in asgs_to_disable:
+                        try:
+                            disable_cluster_asg(cluster, asg)
+                        except:
+                            LOG.warning("Unable to disable ASG '{}' after failure.".format(asg))
+                            continue
```
doctoryes: This `continue` does nothing - I will remove it.

tubular/asgard.py
doctoryes: My previous comment was wrong. A `continue` here would do nothing, just as the `continue` above does. So I won't add one here, since it won't change the program flow.

macdiesel: As long as it's consistent and moved in to a function it really doesn't matter. Let me ask though, do you think having the continue there makes the code more readable/understandable?

tubular/asgard.py
maxrothman: I would prefer to decouple this step entirely and make it possible to have cleanup be in a separate pipeline stage to support use cases such as waiting for manual input before switching over, "soaking" new instances with no requests, etc.

macdiesel: Max, I don't think that manual input is something that would provide a lot of value here and goes against the grain of automate everything. 

That said, parsing your response it seems that perhaps soaking the instances may be a good idea?

maxrothman: Not necessarily? I'm having trouble articulating a good usecase for separating these two stages, but I can definitely imagine someone wanting to put additional pipeline stages in between them.

doctoryes: @maxrothman For this iteration of deploy and rollback, the goal is to automate our current release process - what DevOps currently does. While I could see useful reasons for keeping both ASGs active, such as directing a fractional amount of traffic to a new release indefinitely, it's my understanding that we're not currently doing that sort of thing. Nothing prevents us from implementing this in the future. But we'd like to just implement the same process that DevOps uses for now. How long does DevOps wait in the red/red state before disabling the old release's ASGs?

tubular/asgard.py
doctoryes: With the code broken into a separate method, it's pretty clear without a `continue`.

tubular/ec2.py
doctoryes: I have not tried tag removal from an actual ASG yet. But I will set the tag value to `None` here.

tubular/asgard.py
doctoryes: I added a env-var-controlled delay between starting the new ASGs _and_ performing health checks on them.

doctoryes:
@macdiesel Ready for another review.

tubular/asgard.py
maxrothman: 👍

macdiesel:
🚢 🇮🇹 👍

PR #70, opened 2016-10-05
Fix script error

doctoryes:
This PR is a fix for this error:
https://gocd.tools.edx.org/go/tab/build/detail/deploy-edx-ami/500/deploy_ami/1/deploy_ami_job

@edx/pipeline-team PTAL

### DIFF
--- a/scripts/asgard-deploy.py
+++ b/scripts/asgard-deploy.py
@@ -5,6 +5,7 @@ Command-line script used to deploy an AMI.
 from __future__ import unicode_literals
 
 from os import path
+import os
 import sys
 import logging
 import traceback
@@ -29,6 +30,14 @@ def deploy(ami_id, out_file, config_file, dry_run):
     """
     Method which deploys an AMI.
     """
+    env_ami_id = os.environ.get('AMI_ID', None)
+    if env_ami_id and ami_id != env_ami_id:
+        click.secho(
+            'Error: Command-line and env var AMI_ID do not match. ({} != {})'.format(ami_id, env_ami_id),
+            fg='red'
+        )
+        sys.exit(1)
+
     if config_file:
         config = yaml.safe_load(open(config_file, 'r'))
         if config:
--- a/scripts/asgard-deploy.py
+++ b/scripts/asgard-deploy.py
@@ -31,12 +31,12 @@ def deploy(ami_id, out_file, config_file, dry_run):
     """
     if config_file:
         config = yaml.safe_load(open(config_file, 'r'))
+        if config:
+            if not ami_id and 'ami_id' in config:
+                ami_id = config['ami_id']
     if not ami_id:
-        if 'ami_id' in config:
-            ami_id = config['ami_id']
-        else:
-            click.secho('AMI ID not specified in environment, on cli or in config file.', fg='red')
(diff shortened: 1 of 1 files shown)


doctoryes:
Note: I removed the commit related to the drupal test...

jzoldak:
Why don't we have tests for these scripts? Too heavyweight? 
Would they catch issues like this? I'm not sure what the impact in terms of wasted time/resources was for this bug (or ones like it). If it's high we should add tests. 

Also, I just realizing that there is no coverage calculated/reported for this repo so we're flying blind on tests. Could you please put in a JIRA ticket for that?

doctoryes:
@jzoldak The idea was that the `scripts` would contain simple scripts that marshalled parameters into the code that did the real work - and _that_ code would be tested, while the scripts didn't need testing. That's still mostly true, though we should keep an eye on it. This particular bug wasn't high-impact and was simple to fix.

I created the coverage ticket here:
https://openedx.atlassian.net/browse/TE-1744

macdiesel:
It's possible with the current implementation to pass in a config file with one AMI ID and a command line option AMI ID that do not match. If this happens we fail and return a non zero exit code.

macdiesel (approved):

jzoldak:
@doctoryes ok I get it. Thx for entering the coverage ticket.

doctoryes:
@macdiesel I added a check for the non-matching env var and command-line params for AMI id. I wish `click` just handled this automagically, but alas. I used this code to test things:

```
import os
import click

@click.command()
@click.option('--username', envvar='USERNAME')
def greet(username):
    env_username = os.environ.get('USERNAME', None)
    if env_username and username != env_username:
        print "Error: {} != {} !!!!".format(username, env_username)
    else:
        click.echo('Hello %s!' % username)

if __name__ == '__main__':
    greet()
```

If this change looks good to you, merge away.

scripts/asgard-deploy.py
```diff
@@ -29,14 +30,22 @@ def deploy(ami_id, out_file, config_file, dry_run):
     """
     Method which deploys an AMI.
     """
+    env_ami_id = os.environ.get('AMI_ID', None)
+    if env_ami_id and ami_id != env_ami_id:
```
macdiesel, line 34: TBH I hadn't thought of this case, but it could happen. The use case I was thinking of was an AMI id passed via an env variable or command line and a different AMI_ID coming in via the config file.

doctoryes: Then - perhaps the best way to handle the AMI id is to establish and implement a clear priority order, such as:
The ami_id used will be taken using the order of precedence:
- '--ami_id' command line option
- AMI_ID environment variable
- ami_id key/value from config file

The other option is to error out when any of the set values above don't match. I'm starting to agree with @jzoldak now - perhaps these scripts _do_ need tests.

macdiesel: I'm of the mind on this right now that we should move forward on this as it stands to deal with the error case and backlog a ticket to implement something better than what we have now.

doctoryes: Story created:
https://openedx.atlassian.net/browse/TE-1765
Merging now...

PR #71, opened 2016-10-12
When fetching ASGs for an EDP, continue fetching if there are > 50 items

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/tubular/ec2.py
+++ b/tubular/ec2.py
@@ -123,25 +123,33 @@ def asgs_for_edp(edp, filter_asgs_pending_delete=True):
     """
     autoscale = boto.connect_autoscale()
     all_groups = autoscale.get_all_groups()
-    LOG.debug("All groups: {}".format(all_groups))
-    for group in all_groups:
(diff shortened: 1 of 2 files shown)


tubular/ec2.py
```diff
@@ -123,25 +123,33 @@ def asgs_for_edp(edp, filter_asgs_pending_delete=True):
...
 
-            group_edp = EDP(group_env, group_deployment, group_play)
+                group_edp = EDP(group_env, group_deployment, group_play)
 
-            if group_edp == edp:
-                yield group.name
+                if group_edp == edp:
+                    yield group.name
+        if all_groups.next_token:
+            all_groups = autoscale.get_all_groups(next_token=all_groups.next_token)
```
doctoryes, line 149: Any other places where we need to check these `next_token`s?

macdiesel: I did an audit of the code last night. This get_all_groups is used in a few places. I'm going to pull it out in to a method. 

There is one other call that paginates "get_all_load_balancers" I will also move that out in to a method.

PR #72, opened 2016-10-12
asgs_for_edp is no longer a generator, simply returns a list

macdiesel:
@doctoryes the generator in question.


tubular/asgard.py
```diff
@@ -98,7 +98,7 @@ def clusters_for_asgs(asgs):
     cluster_json = _parse_json(url, response)
 
     # need this to be a list so that we can test membership.
-    asgs = list(asgs)
+    asgs = asgs
```
doctoryes: This statement is now a no-op - remove?

doctoryes (commented):
After the one removal, 👍

PR #75, opened 2016-11-02
Rollback now respects 404 from asgard when an ASG does not exist

macdiesel:
@edx/pipeline-team @maxrothman PTAL. I would like to have this finished in time for the release next week.

@doctoryes you may want to take a good pass on this. Lets consider it a WIP.

### DIFF
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -406,6 +406,8 @@ def disable_asg(asg):
     Raises:
         TimeoutException: If the task to enable the ASG fails..
         BackendError: If asgard was unable to disable the ASG
+        ASGDoesNotExistException: If the ASG does not exist
+        CannotDisableActiveASG: if the current ASG is active
     """
     try:
         if is_asg_pending_delete(asg):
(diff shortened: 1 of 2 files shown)


tubular/asgard.py
```diff
@@ -657,6 +664,12 @@ def _disable_cluster_asg(cluster, asg):
         Shifts ASG from enabled to disabled.
         """
         disable_asg(asg)
+        _move_asg_from_enabled_to_disaled(cluster, asg)
+
+    def _move_asg_from_enabled_to_disaled(cluster, asg):
```
doctoryes: disaled => disabled

doctoryes: Nit: Perhaps a general purpose method for moving an ASG between lists for both enabled=>disabled and disabled=>enabled should be created?

tubular/asgard.py
```diff
@@ -727,11 +740,23 @@ def _disable_clustered_asgs(clustered_asgs, failure_msg):
...
+                elif asg in asgs_enabled[cluster]:
+                    # If the asg is not enabled, but we have it in the enabled list remove it. This may occur by
+                    # pulling from 2 different sources of truth at different intervals. The asg could have been disabled
+                    # in the intervening time.
+                    _move_asg_from_enabled_to_disaled(cluster, asg)
+            except ASGDoesNotExistException:
+                # This operation should not fail if one of the baseline ASGs was removed during the deployment process
+                LOG.info("ASG {asg} in cluster {cluster} no longer exists, removing it from the enabled cluster list"
+                         .format(asg=asg, cluster=cluster))
+                _move_asg_from_enabled_to_disaled(cluster, asg)
```
doctoryes: If an ASG no longer exists, should it be in the list of disabled ASGs? A philosophical question....

macdiesel: Probably not, though I really don't want to add additional k/v pair to the dict to cover this use case and we've already programmed defensively against this case. Also the output is potentially confusing to the user if the input asked for this ASG to be disabled but then we don't report back at all on it's state.

A backlog ticket to make it more clear?

doctoryes:
👍

PR #77, opened 2016-11-16
Change default timeout for ASG creation.

doctoryes:
https://openedx.atlassian.net/browse/TE-1805

This PR changes the name of ASG creation timeout var from `ASGARD_WAIT_TIMEOUT` to `ASGARD_NEW_ASG_CREATION_TIMEOUT`, which is more specific. I checked the edx-gomatic repo and no pipelines currently specify this environment variable - they all use the default timeout.

It appears that Asgard itself times out at 25 minutes. This PR changes the tubular timeout default to 20 minutes.


macdiesel:
@doctoryes I was thinking that we should set the timeout to whatever the asgard timeout is now or lower the asgard timeout and also make this match.  @feanil thoughts?

feanil:
@macdiesel agreed.  Let's set it to match Asgard for now.

doctoryes:
@edx/pipeline-team I bumped up the timeout. Ready for another review.

macdiesel (approved):

PR #79, opened 2016-11-22
move code to cut branches/PRs from the testeng-ci repo

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/requirements/test_requirements.txt
+++ b/requirements/test_requirements.txt
@@ -6,5 +6,6 @@ pep8==1.7.0
 pylint
 caniusepython3
 edx_lint
+responses==0.5.0
 
 -r ../requirements.txt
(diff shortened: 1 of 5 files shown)


doctoryes:
@macdiesel and I discussed this PR - it copies code from the testeng-ci repo. That's fine with me - but IMO this code should move away from using straight-up requests and instead use the `github3` Python module, which is our internally agreed-upon standard:
https://github.com/sigmavirus24/github3.py
The tech debt incurred should be accounted for in a created JIRA ticket.

macdiesel:
@doctoryes the follow on ticket:
https://openedx.atlassian.net/browse/TE-1860

doctoryes:
👍 👍

PR #80, opened 2016-11-23
Implement polling for build results after triggering a jenkins job

jzoldak:
@edx/platform-team Please review
@cpennington FYI

### DIFF
--- a/scripts/jenkins_trigger_build.py
+++ b/scripts/jenkins_trigger_build.py
@@ -25,35 +25,37 @@ from tubular import jenkins  # pylint: disable=wrong-import-position
     "--user_name",
     help="The Jenkins username for triggering the job.",
     type=str,
-    required=True
+    required=True,
 )
 @click.option(
     "--user_token",
     help="API token for the user. Available at {url}/user/{user_name)/configure",
     type=str,
-    required=True
+    required=True,
+    envvar='JENKINS_USER_TOKEN'
 )
 @click.option(
     "--job",
     help="The name of the jenkins job. E.g. test-project",
     type=str,
-    required=True
+    required=True,
 )
 @click.option(
     "--token",
     help="The authorization token for the job. Must match that configured in the job definition.",
     type=str,
-    required=True
+    required=True,
+    envvar='JENKINS_JOB_TOKEN'
 )
 @click.option(
     "--cause",
     help="Text that will be included in the recorded build cause.",
     type=str,
-    required=False
+    required=False,
 )
 @click.option(
     '--param',
-    help='Key/value pairs to pass to the job as parameters. E.g. --param FOO bar --param BAZ biz',
+    help=u'Key/value pairs to pass to the job as parameters. E.g. --param FOO bar --param BAZ biz',
     multiple=True,
     required=False,
     type=(str, str)
--- a/tubular/jenkins.py
+++ b/tubular/jenkins.py
@@ -1,22 +1,22 @@
 """
 Methods to interact with the Jenkins API to perform various tasks.
 """
-from __future__ import unicode_literals
-
 import logging
-import os
-import requests
+
+import backoff
+from jenkinsapi.jenkins import Jenkins
+from jenkinsapi.custom_exceptions import JenkinsAPIException
+from requests.exceptions import HTTPError
 
 from tubular.exception import BackendError
 
-
-REQUESTS_TIMEOUT = float(os.environ.get("REQUESTS_TIMEOUT", 10))
-
 LOG = logging.getLogger(__name__)
+
+MAX_TRIES = 12
 
 
 def trigger_build(base_url, user_name, user_token, job_name, job_token, job_cause=None, job_params=None):
-    """
+    u"""
     Trigger a jenkins job/project (note that jenkins uses these terms interchangeably)
 
     Args:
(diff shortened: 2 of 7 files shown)


scripts/jenkins_trigger_build.py
```diff
@@ -19,43 +19,50 @@
...
-    required=False
+    required=False,
+    envvar='JENKINS_CAUSE'
 )
 @click.option(
     '--param',
     help='Key/value pairs to pass to the job as parameters. E.g. --param FOO bar --param BAZ biz',
     multiple=True,
     required=False,
+    envvar='JENKINS_PARAM',
```
jzoldak: @edx/platform-team Does anybody have a suggestion for how to handle multiple values?

jmbowman: In the environment variable?  Looks like you can just [space separate them](http://click.pocoo.org/5/options/#multiple-values-from-environment-values).

macdiesel: Just a personal preference but I find it easier to troubleshoot the jobs if you pass these items in via the command line rather than environment variables. I find it more descriptive of what's being passed in in used. 

On the flip side when you use environment variables there is a lot of scrolling around and log interpretation necessary to figure what exact variables the program was triggered with.

my 2 cents is we shouldn't put this envvar stuff in.

doctoryes: @macdiesel and I have discussed this before - there's also the issue of when the envar is set *and* the cmd-line param is passed-in. It's a confusing situation that might cause unexpected behavior, where you might think that the value is X but the value getting used is actually Y.

scripts/jenkins_trigger_build.py
```diff
@@ -25,35 +25,37 @@
...
-    required=True
+    required=True,
 )
 @click.option(
     "--user_token",
     help="API token for the user. Available at {url}/user/{user_name)/configure",
     type=str,
-    required=True
+    required=True,
+    envvar='JENKINS_USER_TOKEN'
```
doctoryes, line 35: We've been moving away from using `envvar` in click options - due to the possible confusion of when both the cmd-line arg and the env var is specified.

scripts/jenkins_trigger_build.py
```diff
@@ -25,35 +25,37 @@
...
-    required=True
+    required=True,
 )
 @click.option(
     "--token",
     help="The authorization token for the job. Must match that configured in the job definition.",
     type=str,
-    required=True
+    required=True,
+    envvar='JENKINS_JOB_TOKEN'
```
doctoryes, line 48: Ditto.

tubular/jenkins.py
```diff
@@ -1,22 +1,24 @@
...
+from requests.exceptions import HTTPError
 
-REQUESTS_TIMEOUT = float(os.environ.get("REQUESTS_TIMEOUT", 10))
+from tubular.exception import BackendError
 
 LOG = logging.getLogger(__name__)
+# exponential backoff with a base of 2 and factor of 1 (defaults)
+# at 9 times will be 12.8 minutes (768 sec) and 10 times will be 25.6 minutes (1536 sec)
+# Warning: setting MAX_TRIES to 0 will loop infinitely.
+MAX_TRIES = 9
```
doctoryes: So the maximum time it'll ever wait on a Jenkins job completion is 12.8 minutes? Or is that just the final 9th try?

scripts/jenkins_trigger_build.py
jzoldak: Hm, I want to make sure that these two values stay private (in console logs, jenkins build parameter displays, etc.)
What's the alternative pattern?

jzoldak (commented):
I'll put this chart right in the code as a comment to explain and help people decide on the best value:

|wait x sec|make attempt #|total time (sec)|total time (min)|
|----:|---:|----:|-----:|
|0    |1   |0    | 0    |
|1    |2   |1    |0.02  |
|2    |3   |3    |0.05  |
|4    |4   |7    |0.12  |
|8    |5   |15   |0.25  |
|16   |6   |31   |0.52  |
|32   |7   |63   |1.05  |
|64   |8   |127  |2.12  |
|128  |9   |255  |4.25  |
|256  |10  |511  |8.52  |
|512  |11  |1023 |17.05 |
|1024 |12  |2047 |34.12 |
|2048 |13  |4095 |68.25 |

scripts/jenkins_trigger_build.py
doctoryes: We've relied on the envvar expansion in the bash command to pass the cmd-line param directly into Ansible/Python scripts. So you'd remove the `envvar` here and then specify `--user_token $JENKINS_USER_TOKEN --job_token $JENKINS_JOB_TOKEN` in your Gomatic.

doctoryes: Here's some similar gomatic:
https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/tasks.py#L298

jzoldak: I'm inclined to get this merged in as-is for now. It will need to be refactored later anyhow to provide for the ability to define jenkins usernames, associated user tokens, and job tokens on a per-job basis.

doctoryes: Sounds fine - 👍

doctoryes (approved):

PR #81, opened 2016-11-30
Add method to extract last base AMI ID used from a particular EDP.

doctoryes:
https://openedx.atlassian.net/browse/TE-1814

Phase 1 of 2: @macdiesel could I get a preliminary review here? It's not finished. I'll make a couple of other notes.

### DIFF
--- a/scripts/retrieve_base_ami.py
+++ b/scripts/retrieve_base_ami.py
@@ -1,0 +1,68 @@
+"""
+Command-line script used to retrieve the last base AMI ID used for an environment/deployment/play.
+"""
+# pylint: disable=invalid-name,open-builtin
+from __future__ import unicode_literals
+
+from os import path
+import sys
+import logging
+import traceback
+import click
+import yaml
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
+
+from tubular import ec2  # pylint: disable=wrong-import-position
+
+logging.basicConfig(level=logging.INFO)
+
+
+@click.command()
+@click.option(
+    '--environment', '-e',
+    help='Environment for AMI, e.g. prod, stage',
+    required=True
+)
+@click.option(
+    '--deployment', '-d',
+    help='Deployment for AMI e.g. edx, edge',
+    required=True
+)
+@click.option(
+    '--play', '-p',
+    help='Play for AMI, e.g. edxapp, insights, discovery',
+    required=True
+)
+@click.option(
+    '--out_file',
+    help='Output file for the AMI information yaml.',
+    default=None
+)
+def retrieve_base_ami(environment, deployment, play, out_file):
+    """
+    Method used to retrieve the last base AMI ID used for an environment/deployment/play.
+    """
+    try:
+        ami_id = ec2.active_ami_for_edp(environment, deployment, play)
+        ami_info = {'base_ami_id': ami_id}
+        logging.info("Found active AMI ID for {env}-{dep}-{play}: {ami_id}".format(
+            env=environment, dep=deployment, play=play, ami_id=ami_id
+        ))
+
+        if out_file:
+            with open(out_file, 'w') as stream:
+                yaml.safe_dump(ami_info, stream, default_flow_style=False, explicit_start=True)
+        else:
+            print yaml.safe_dump(ami_info, default_flow_style=False, explicit_start=True)
+
+    except Exception as err:  # pylint: disable=broad-except
+        traceback.print_exc()
+        click.secho('Error finding base AMI ID.\nMessage: {}'.format(err.message), fg='red')
+        sys.exit(1)
+
+    sys.exit(0)
+
+if __name__ == "__main__":
+    retrieve_base_ami()  # pylint: disable=no-value-for-parameter
--- a/tubular/ec2.py
+++ b/tubular/ec2.py
@@ -13,6 +13,7 @@ from boto.ec2.autoscale.tag import Tag
 from tubular.utils import EDP, WAIT_SLEEP_TIME
 from tubular.exception import (
     ImageNotFoundException,
+    MultipleImagesFoundException,
     MissingTagException,
     TimeoutException,
 )
@@ -49,9 +50,10 @@ def get_all_load_balancers(names=None):
     Get all the ELBs
 
     Arguments:
-        names (list) - A list of ELB names as strings
+        names (list): A list of ELB names as strings
+
     Returns:
-        a list of  of :class:`boto.ec2.elb.loadbalancer.LoadBalancer`
+        a list of :class:`boto.ec2.elb.loadbalancer.LoadBalancer`
     """
     elb_conn = boto.connect_elb()
     fetched_elbs = elb_conn.get_all_load_balancers(names)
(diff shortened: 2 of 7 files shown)


tubular/tests/test_ec2.py
```diff
@@ -73,6 +78,65 @@ def test_restrict_ami_to_stage(self):
...
+        }
+        fake_asg = create_asg_with_tags(
+            fake_asg_name,
+            fake_asg_tags,
+            ami_id=fake_ami_id,
+            elbs=[fake_elb]
+        )
+        self.assertEqual(ec2.ami_for_edp('foo', 'bar', 'baz'), fake_ami_id)
+
+    @unittest.skip("Test always fails due to not successfuly creating two different AMI IDs in single ELB.")
```
doctoryes: I couldn't get this test to work. It was unclear whether it was moto or boto that thwarted me. Is it important enough to test this edge case?

tubular/tests/test_utils.py
```diff
@@ -55,6 +62,17 @@ def create_asg_with_tags(asg_name, tags, ami_id="ami-abcd1234", elbs=None):
...
+
+    # Each ASG tag that has 'propagate_at_launch' set to True is *supposed* to be set on the instances.
+    # However, it seems that boto (or moto) does not properly set the tags on the instances created by the ASG.
+    # So set the tags on the ASG instances manually instead.
+    ec2_conn = boto.connect_ec2()
+    for asg in conn.get_all_groups():
+        if asg.name == asg_name:
+            asg_instance_ids = [instance.instance_id for instance in asg.instances]
+            for instance_id in asg_instance_ids:
+                ec2_conn.create_tags(instance_id, tags)
```
doctoryes, line 74: Despite adding `propagate_at_launch_time` to the tags above, I found this extra step to be necessary to get the tags on the ASG instances. Again, it was unclear if it was moto causing this issue. Thoughts?

doctoryes:
The part that isn't done is getting the rest of the code to use this new method for base AMI selection, BTW.

tubular/ec2.py
```diff
@@ -65,6 +67,62 @@ def get_all_load_balancers(names=None):
...
+    """
+    instance_elbs = []
+    for elb in elbs:
+        elb_instance_ids = [inst.id for inst in elb.instances]
+        if instance_id in elb_instance_ids:
+            instance_elbs.append(elb)
+    return instance_elbs
+
+
+def ami_for_edp(env, dep, play):
```
macdiesel: Could we name this active_ami_for_edp ?

tubular/tests/test_ec2.py
macdiesel: I believe your issue may be related to this: https://github.com/spulec/moto/issues/718

Since the code here: https://github.com/edx/tubular/pull/81/files#diff-62bdf1ee91eb21bc467805ff1a25b581R107 
is using multiple filters.

macdiesel: Please add a note, leave the test in and skip it for now.

tubular/tests/test_utils.py
macdiesel: Tagging on ASGs in MOTO is new. Are you sure your dev environment has the latest version of moto installed? 

if so I would add a note about why this is here and leave it in.

tubular/ec2.py
```diff
@@ -65,6 +67,62 @@ def get_all_load_balancers(names=None):
...
 
+def _instance_elbs(instance_id, elbs):
+    """
+    Given an EC2 instance and ELBs, return the ELB(s) in which it is active.
+
+    Arguments:
+        instance_id(boto.ec2.instance.Reservation): Instance used to find out which ELB it is active in.
+        elbs(list(boto.ec2.elb.loadbalancer.LoadBalancer)): List of ELBs to us in checking.
+    Returns:
+        boto.ec2.elb.loadbalancer.LoadBalancer: One or more ELBs used by the passed-in instance -or- None.
```
macdiesel: fwiw I think we may have been doing these google doc strings a big wrong and the returns type should have backticks:
```
Args:
            param1 (str): Description of `param1`.
            param2 (:obj:`int`, optional): Description of `param2`. Multiple
                lines are supported.
            param3 (:obj:`list` of :obj:`str`): Description of `param3`.
```

macdiesel (commented):
I think the testing difficulties you are having are due to moto.

Also a few suggestions.

tubular/ec2.py
doctoryes: Will do.

tubular/ec2.py
doctoryes: Will do.

tubular/tests/test_utils.py
doctoryes: I updated it to have the most recent moto version - 0.4.30. Then I removed this section and tried tests. The test failed without this section. I'll leave it in with a note.

tubular/tests/test_ec2.py
doctoryes: I tried the newest version of moto (0.4.30) - no dice. Keeping the skip.

tubular/ec2.py
macdiesel: I was incorrect about this, please ignore me.

feanil (approved):

doctoryes:
@edx/pipeline-team I've added some commits since your initial reviews. Please review again.

scripts/retrieve_base_ami.py
```diff
@@ -0,0 +1,77 @@
...
+
+from tubular import ec2  # pylint: disable=wrong-import-position
+
+logging.basicConfig(level=logging.INFO)
+
+
+@click.command()
+@click.option(
+    '--environment', '-e',
+    envvar='AMI_ENVIRONMENT',
```
macdiesel: I thought we were going to stop using environment variables unless they contained private data?

scripts/retrieve_base_ami.py
```diff
@@ -0,0 +1,77 @@
...
+    '--out_file',
+    help='Output file for the AMI information yaml.',
+    default=None
+)
+def retrieve_base_ami(environment, deployment, play, out_file):
+    """
+    Method used to retrieve the last base AMI ID used for an environment/deployment/play.
+    """
+    kwargs = {
+        'aws_access_key_id': os.environ.get('AWS_ACCESS_KEY_ID'),
```
macdiesel: Can we move these items in to click and pass them in via method signature?

tubular/ec2.py
```diff
@@ -44,16 +45,20 @@ def get_all_autoscale_groups(names=None):
...
+        names (list): A list of ELB names as strings
+
     Returns:
-        a list of  of :class:`boto.ec2.elb.loadbalancer.LoadBalancer`
+        a list of :class:`boto.ec2.elb.loadbalancer.LoadBalancer`
     """
-    elb_conn = boto.connect_elb()
+    elb_conn = boto.connect_elb(
+        aws_access_key_id=kwargs.get('aws_access_key_id'),
+        aws_secret_access_key=kwargs.get('aws_secret_access_key')
```
macdiesel: I think boto will put these out of the environment directly:
http://boto.cloudhackers.com/en/latest/boto_config_tut.html#introduction

>  Some options, such as credentials, can also be read from environment variables (e.g. AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECURITY_TOKEN and AWS_PROFILE)

scripts/retrieve_base_ami.py
macdiesel: I take this back, see my other comment about boto and reading these variables directly out of the environment.

>  Some options, such as credentials, can also be read from environment variables (e.g. AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECURITY_TOKEN and AWS_PROFILE)

tubular/ec2.py
```diff
@@ -65,6 +70,75 @@ def get_all_load_balancers(names=None):
...
+    Returns:
+        str: Base AMI id of current active deployment for the EDP.
+    Raises:
+        MultipleImagesFoundException: If multiple AMI IDs are found within the EDP's ELB.
+        ImageNotFoundException: If no AMI IDs are found for the EDP.
+    """
+    LOG.info("Looking up AMI for {}-{}-{}...".format(env, dep, play))
+    ec2_conn = boto.connect_ec2(
+        aws_access_key_id=kwargs.get('aws_access_key_id'),
+        aws_secret_access_key=kwargs.get('aws_secret_access_key')
```
macdiesel: Again these should come out of the environment.

macdiesel (commented):
a few comments.

scripts/retrieve_base_ami.py
doctoryes: This task has dragged on for so long that we probably made the decision after I wrote this code. :) Will do.

scripts/retrieve_base_ami.py
doctoryes: In my test pipeline system, the boto code was *not* using the envvars - so I added them explicitly. I'll do some more digging to figure out why.

doctoryes: I had conflated two separate issues here - all is well! I've removed the explicit reference to the AWS creds.

doctoryes:
@macdiesel All issues addressed - ready for another round.

macdiesel (approved):

PR #83, opened 2016-12-21
script that will create a branch on a github repository

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/scripts/cut_branch.py
+++ b/scripts/cut_branch.py
@@ -1,0 +1,129 @@
+"""
+Command-line script to create a release candidate for an application
+"""
+from __future__ import unicode_literals
+
+from os import path
+import sys
+import logging
+import click
+import yaml
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
+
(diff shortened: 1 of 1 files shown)


scripts/cut_branch.py
```diff
@@ -0,0 +1,129 @@
...
+            rc=target_branch,
+            sha=commit_hash,
+            msg=commit_message
+        )
+    )
+    try:
+        github_api.delete_branch(target_branch)
+    except GithubException:
+        LOG.error("Unable to delete branch {branch_name}. " +
+                  "Will attempt to recreate"
```
feanil, line 104: This message is a bit misleading. You mean it's going to try creating the new branch right?  Is it possible to check for branch existence before creating it?

feanil (commented):
One question, otherwise seems good.

scripts/cut_branch.py
macdiesel: We can, but again could we defer this work to a subsequent PR.

feanil: I think if you take out the 'will attempt to recreate' part I'm good with the rest of this.

feanil (approved):

PR #84, opened 2016-12-21
Do not check the CI of a commit when creating a branch

macdiesel:
@edx/pipeline-team PTAL. This removes the checking of CI when cutting a branch.


doctoryes (approved):

benpatterson:
Like... This lets us build on a hash with failed tests?

macdiesel:
The idea here is that we would pull from master to create the RC while the tests run, then check the CI status at a later point in the pipeline.

benpatterson:
Gotcha

PR #91, opened 2017-01-05
Add method for finding PRs given a commit range

rlucioni:
Given a start SHA and an end SHA, the method returns a list of PRs between the two, excluding the start SHA and including the end SHA.

TE-1899

@edx/pipeline-team please review.

### DIFF
--- a/tubular/release.py
+++ b/tubular/release.py
@@ -3,6 +3,7 @@
 from __future__ import print_function, unicode_literals
 
 import logging
+import os
 import string
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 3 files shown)


tubular/release.py
```diff
@@ -205,6 +205,71 @@ def most_recent_good_commit(self, branch):
...
+        comparison = self.github_repo.compare(start_sha, end_sha)
+        shas = [commit.sha for commit in comparison.commits]
+
+        # The Search API limits search queries to 256 characters. SHA1s are 40
+        # characters long. To help stay under the rate limit while staying within
+        # the max search query size, we can batch up to 5 SHA1s at a time. For more
+        # about searching issues, see https://help.github.com/articles/searching-issues.
+        issues = []
+        for sha_batch in batch(shas, 5):
+            issues += self.github_connection.search_issues(
```
cpennington: Is it worth catching rate-limit errors and then retrying after a backoff?

feanil (approved):
Looks good to me.

cpennington (approved):
I like how you have this. Another possible way of working around the rate-limiting would be to have a set of commits that you haven't found PRs for yet, and then to look at the list of commits in the PRs that you find after a search, and remove those from your pending list. That would mean that you'd never search for a commit that you've already identified the PR for.

tubular/release.py
rlucioni: @cpennington I thought about that, but decided against it for now. Hitting the rate limit would require >150 commits in the comparison. That didn't seem likely enough to make adding and testing retry worth it. If you feel differently, I'm open to adding it.

cpennington: Well, or running the comparison multiple times within the same minute from different jobs.

cpennington: I'm leaning towards addressing it, because if it fails, it'll fail in the middle of a release (which would be pretty inconvenient).

tubular/release.py
```diff
@@ -205,6 +205,71 @@ def most_recent_good_commit(self, branch):
...
+
+            Yields:
+                list
+            """
+            length = len(batchable)
+            for index in range(0, length, batch_size):
+                yield batchable[index:index + batch_size]
+
+        comparison = self.github_repo.compare(start_sha, end_sha)
+        shas = [commit.sha for commit in comparison.commits]
```
doctoryes: From the docs, it appears you can use only 7+ characters from the SHA. Perhaps `commit.sha[:10]` or the like here?

Good stuff BTW...

rlucioni: @doctoryes can you link me to that section in the docs?

doctoryes: Along with an increase in the number batched as well, of course...

doctoryes: https://help.github.com/articles/searching-issues/#search-by-the-commit-shas-within-a-pull-request

tubular/release.py
rlucioni: @cpennington fair, I hadn't considered that. I'll see what I can do.

tubular/release.py
rlucioni: @doctoryes interesting. For the record, the 40 character SHAs are supported just fine, but trimming them down to 10 character SHAs would allow us to increase the batch size from 5 to 20. That means we'd need to search for >600 commits in a minute to trigger the rate limiting.

@cpennington this seems more elegant than complicating things with retries. Thoughts?

cpennington: Yeah, makes sense to me.

tubular/release.py
```diff
@@ -228,30 +228,38 @@ def get_pr_range(self, start_sha, end_sha):
...
+        # The Search API limits search queries to 256 characters. Untrimmed SHA1s
+        # are 40 characters long. To avoid exceeding the rate and search query size
+        # limits, we can batch SHAs in our searches. Reserving 56 characters for
+        # qualifiers (i.e., type, base, user, repo) leaves us with 200 characters.
+        # As with all other terms in the query, the batched SHAs need to be separated
+        # by a character of whitespace. A batch size of 18 10-character SHAs requires
+        # 17 characters of whitespace, for a total of 18*10 + 17 = 197 characters.
+        # We'd need to search for >540 commits in a minute to exceed the rate limit.
+        SHA_LENGTH = 10
+        BATCH_SIZE = 18
```
macdiesel: Could you set these like this so we can override via the environment? https://github.com/edx/tubular/blob/master/tubular/asgard.py#L35-L36

rlucioni: @macdiesel sure, you mean check for environment variables and use a default otherwise?

macdiesel: 👍

rlucioni: @macdiesel making this change is no big deal, but I think it's dangerous. These values are tuned to the GitHub Search API. They shouldn't need to change across environments.

rlucioni: If we make it possible for these to be overridden, we give others an easy way to shoot themselves in the foot if they don't understand why these are set as they are.

macdiesel: I agree it could be an easy way to hurt yourself. On the flip side if we have to make a change in the code to change these parameters the entire pipeline needs to be restarted (possibly several stages/pipelines up stream) that could really slow down the release process should we run in to throttling issues.

The reason for this is that tubular is locked in as a material at the start of the very first pipeline. If we make these variables settable via the environment we can modify the pipeline itself mid stream to adjust for throttling that we may encounter.

rlucioni: @macdiesel thanks for the context, that makes sense. I'll make the change.

rlucioni:
@edx/pipeline-team all feedback addressed. I don't have write permissions to this repo. Unless there's any more feedback for me to address, could someone merge it for me?

doctoryes (approved):

macdiesel (approved):
I will merge after CI passes. Thanks @rlucioni

PR #92, opened 2017-01-05
New version of pygithub is raising a different Exception class when a deleted branch does not exist

macdiesel:
@edx/pipeline-team @cpennington PTAL.

### DIFF
--- a/scripts/cut_branch.py
+++ b/scripts/cut_branch.py
@@ -13,7 +13,6 @@ import yaml
 sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
 
 from tubular.release import GitRelease, NoValidCommitsError  # pylint: disable=wrong-import-position
-from github.GithubException import GithubException  # pylint: disable=wrong-import-position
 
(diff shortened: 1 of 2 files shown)


doctoryes (commented):
Same create/delete_branch calls exist in create_release_candidate.py also. Perhaps change the exception catching there as well?

cpennington (approved):
It seems like there might be something subtler going on here (or some other way we could catch it before that particular call), but for now, let's get the release unblocked.

scripts/cut_branch.py
```diff
@@ -1,6 +1,7 @@
 """
 Command-line script to create a release candidate for an application
 """
+# pylint: disable=broad-except
```
cpennington: It might be better to put this on a per-line basis, so that other spots don't slip through.

cpennington (approved):

macdiesel:
@doctoryes comments addressed.

scripts/cut_branch.py
macdiesel: done.

PR #93, opened 2017-01-06
increase the number of attempts when checking the state of drupal calls

macdiesel:
@mikedikan Looking over the docs on the retry decorator we should have also bumped the number of attempts as well. Even though the max time was increased to 300, attempts were only set to 10 with a 10 second delay. This would still cause the function to try 10 times, with a 10 second delay adding up to the retry being exhausted at around 100 seconds. 

Perhaps as an improvement to the retry decorator we could have a way to have it attempt until success within a time limit. But for now this should get your deploy going.

### DIFF
--- a/tubular/drupal.py
+++ b/tubular/drupal.py
@@ -196,7 +196,7 @@ def backup_database(env, username, password):
     return check_state(response_json["id"], username, password)
 
 
-@retry(attempts=10, delay_seconds=10, max_time_seconds=300)
+@retry(attempts=30, delay_seconds=10, max_time_seconds=300)
 def check_state(task_id, username, password):
     """
     Checks the state of the response to verify it is "done"


mikedikan (approved):

mikedikan:
Ah, ok, thanks for the explanation

cpennington:
Does the retry decorator need to be passed all three arguments? It seems like any 2 would be sufficient.

macdiesel:
I think it would be a good improvement for the decorator.

macdiesel:
Lets also take this scenario under consideration. What about a task/function that may take a variable amount of time. Say a server or DNS is not in service. Your typical 10ms task now takes much longer, maybe 150ms, but due to other SLAs you are required to return a response within 200ms. The logic behind this was so you could cap the total amount of time a retry could take.

Also max_time_seconds is an optional argument, perhaps the better solution in this case is to remove the cap from this particular method and just let it retry until it reaches the retry limit.

PR #94, opened 2017-01-09
Add code to merge branches at end of release process.

doctoryes:
https://openedx.atlassian.net/browse/TE-1886

Work includes the above ticket plus other items. A summary:
- Revamped README file plus LICENSE addition.
- Got rid of obsolete tubular/scripts directory.
  - Moved those scripts out to the scripts directory.
- Added git_repo.py file to add general-purpose git CLI commands.
- Split release.py into release.py/github_api.py files.
  - Split up tests into separate files as well.
- Added create_pr.py script to create new PRs via GitHub API.
- Added merge_pr.py script to merge PRs via GitHub API.
- Added merge_branch.py script to merge branches via git CLI.

The new functionality is missing tests - I acknowledge this and I'll add them in a subsequent ticket/PR. Since I've re-organized code here, I need to get these changes in on the soon side.


tubular/git_repo.py
```diff
@@ -0,0 +1,170 @@
...
+
+        Arguments:
+            branch_name (str): Branch name to use in cloning history.
+
+        Raises:
+            subprocess.CalledProcessError: if the cmd fails
+        """
+        cmd_args = ['git', 'clone']
+        if branch_name:
+            cmd_args.extend(['--branch', branch_name])
```
doctoryes, line 69: So - shallow cloning. This code needs enough history to be able to merge two arbitrary branches - in our case, release/release-candidate/master. However, `--depth=1`  didn't clone enough history to enable merging. This `--branch` option only shallows-out the cloned repository a small amount. There's a  `--shallow-since` option in git now - unfortunately, GitHub doesn't support it yet.

I'm open to suggestions here.

scripts/check_pr_against_branch.py
```diff
@@ -0,0 +1,62 @@
...
+    type=int,
+    required=True
+)
+@click.option(
+    u'--branch_name', u'-b',
+    help=u'Branch to check if the base of the PR.',
+    required=True
+)
+def cli(org, repo, token, pr_number, branch_name):
+    u"""
```
macdiesel, line 43: Does the comment need to be a unicode string?

scripts/check_pr_against_branch.py
```diff
@@ -0,0 +1,62 @@
...
+    help=u'Repo name from the GitHub repository URL of https://github.com/<org>/<repo>',
+    required=True
+)
+@click.option(
+    u'--token',
+    envvar=u'GIT_TOKEN',
+    help=u'The github access token, see https://help.github.com/articles/creating-an-access-token-for-command-line-use/'
+)
+@click.option(
+    u'--pr_number', u'-p',
```
macdiesel: we have been opting to not specify any short command line options (harder to troubleshoot down the road). I would prefer we don't use these.

scripts/check_pr_tests_status.py
```diff
@@ -0,0 +1,64 @@
...
+    default=None,
+    help=u'Pull request number to check.',
+    type=int,
+)
+@click.option(
+    u'--commit_hash', u'-c',
+    help=u'Commit hash to check.',
+)
+def cli(org, repo, token, pr_number, commit_hash):
+    u"""
```
macdiesel, line 41: Is the comment required to be unicode?

scripts/check_pr_tests_status.py
```diff
@@ -0,0 +1,64 @@
...
+    Check the combined status of a GitHub PR/commit in a repo.
+    """
+    # github.enable_console_debug_logging()
+    gh_utils = GitHubAPI(org, repo, token)
+
+    status_success = False
+    if pr_number:
+        status_success = gh_utils.check_pull_request_test_status(pr_number)
+        print u"{}: Combined status of PR #{} is {}.".format(
+            sys.argv[0], pr_number, u"success" if status_success else u"failed"
```
macdiesel, line 58: Are we lose some resolution here? What about the pending state? Wouldn't we want to display that to the user?

scripts/check_pr_tests_status.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+    elif commit_hash:
+        status_success = gh_utils.is_commit_successful(commit_hash)
+        print u"{}: Combined status of commit hash {} is {}.".format(
+            sys.argv[0], commit_hash, u"success" if status_success else u"failed"
+        )
+
+    # An exit code of 0 means success and non-zero means failure.
+    sys.exit(not status_success)
```
macdiesel, line 67: Should probably also exit with non zero if both PR number and a SHA are specified.

tubular/git_repo.py
```diff
@@ -0,0 +1,170 @@
...
+
+        Raises:
+            subprocess.CalledProcessError: if the cmd fails
+        """
+        cmd_args = ['git', 'checkout', '-t', 'origin/{}'.format(branch_name)]
+        self._exec_cmd(cmd_args)
+
+    def merge(self,
+              source_branch,
+              ff_only):
```
macdiesel, line 101: should we set this to True by default?

tubular/git_repo.py
```diff
@@ -0,0 +1,170 @@
...
+    try:
+        yield
+    finally:
+        os.chdir(initial_directory)
+
+
+def merge_branch(repo_url,
+                 source_branch,
+                 target_branch,
+                 ff_only):
```
macdiesel: Perhaps here is the better place to set ff_only to True by default.

tubular/git_repo.py
```diff
@@ -0,0 +1,170 @@
...
+    repo.clone(source_branch)
+    try:
+        with _change_dir(repo):
+            repo.track_branch(target_branch)
+            repo.merge(source_branch, ff_only)
+            repo.push(target_branch)
+    except subprocess.CalledProcessError as exc:
+        raise GitMergeFailed(repr(exc.cmd))
+    finally:
+        repo.cleanup()
```
macdiesel, line 179: I like how this api ended semantically. :1st_place_medal:

scripts/merge_pr.py
```diff
@@ -0,0 +1,80 @@
...
+    u'--repo',
+    help=u'Repo name from the GitHub repository URL of https://github.com/<org>/<repo>'
+)
+@click.option(
+    u'--token',
+    envvar=u'GIT_TOKEN',
+    help=u'The github access token, see https://help.github.com/articles/creating-an-access-token-for-command-line-use/'
+)
+@click.option(
+    u'--pr_number', u'-p',
```
macdiesel: from above, I would rather opt out of the short options.

tubular/gocd/gocd_api_utils.py
```diff
@@ -5,7 +5,7 @@
```
macdiesel, line 5: Since you are moving things around, this stuff should go in a module under the tubular directory called gocd.py. Though it does not necessary need to be part of this PR.

tubular/release.py
```diff
@@ -1,284 +1,16 @@
...
-        for issue in issues:
-            # Merge commits link back to the same PR as the actual commits merged
-            # by that PR. We want to avoid listing the PR twice in this situation,
-            # and also when a PR includes more than one commit.
-            if not pulls.get(issue.number):
-                pulls[issue.number] = self.github_repo.get_pull(issue.number)
-
-        return list(pulls.values())
-
     # Day of week constant
```
macdiesel: I was thinking we should just carry these over with the github API class and just rename/remove the module. These methods are not really useful on their own and this module doesn't really help us "release" the software at all without the github api stuff anyway.

scripts/check_pr_against_branch.py
doctoryes: Will remove.

scripts/check_pr_against_branch.py
doctoryes: It's weird - but click can't easily handle `import unicode_literals` and throws a warning. Details are here:
http://click.pocoo.org/5/python3/#unicode-literals
But when/if you remove that import, then all the strings in the script need `u''`. Pick your poison - runtime warnings or unicode markers.

scripts/check_pr_tests_status.py
doctoryes: Ditto.

scripts/check_pr_tests_status.py
doctoryes: This is a script that I created several months ago and hasn't been used anywhere yet. Quite frankly, it's not very useful in its current form. I thought about removing it - but I left it as a skeleton for a more useful script. I'll clearly specify in the docstring about what success/failure means - but I'm inclined to leave its behavior the same until it's actually used.

scripts/check_pr_tests_status.py
doctoryes: Will do.

tubular/git_repo.py
doctoryes: Will do.

tubular/git_repo.py
doctoryes: Will do below.

tubular/gocd/gocd_api_utils.py
doctoryes: Agreed on both points.

tubular/release.py
doctoryes: These methods don't call the GitHub API - but I can see that they're git-related. How about I move the class over to `github_api.py` but leave it seperate?

scripts/merge_pr.py
doctoryes: Will do.

scripts/check_pr_tests_status.py
doctoryes: Done.

scripts/check_pr_against_branch.py
macdiesel: more of a nit, whatever works.

tubular/release.py
macdiesel: since they don't need any context they are just helper methods, do you want to just put them on the module and take them out of the class?

doctoryes: Will do. And I'll merge the tests back together somewhat.

tubular/git_repo.py
cpennington: Maybe do an expanding window? Clone at `--depth=10` to start (or something), and then double the window each time until you get sufficient depth?

scripts/check_pr_against_branch.py
cpennington: We should just convert the whole repo to python 3.

tubular/git_repo.py
doctoryes: The tricky part there will be differentiating a clone depth error from a "can't fast-forward-merge" or other error. Not impossible - it'll just likely involve the parsing of git output.

How about I leave this as is for now - and log a ticket to come back and optimize it? I'd definitely like to optimize this cloning later - but it'll unblock the release master automation work for now.

scripts/check_pr_against_branch.py
doctoryes: Agreed - we should also make a ticket for this.

tubular/git_repo.py
cpennington: Yeah, you might need another command that is specifically just about figuring out if you have enough depth. Maybe `git merge-base`?

doctoryes:
@cpennington @edx/pipeline-team Looking for more comments and/or thumbs on this one.

macdiesel (approved):
I'm good with what's here with the addition of the followup tickets.

tubular/git_repo.py
```diff
@@ -0,0 +1,179 @@
...
+        if not match:
+            raise InvalidGitRepoURL()
+        self.name = match.group('name')
+
+    def _exec_cmd(self, cmd_args):
+        """
+        Utility method for executing a command.
+        """
+        LOGGER.info("Executing: {}".format(subprocess.list2cmdline(cmd_args)))
+        return subprocess.check_call(cmd_args, stderr=subprocess.STDOUT)
```
cpennington, line 55: Is it worth using something like http://gitpython.readthedocs.io/en/stable/ to wrap git, instead of using our own calls to subprocess directly?

cpennington: Doing so might make it easier to work with (say) a bare repo, so that we don't have the working directory.

cpennington (approved):
I had one more question, but am ok with the code as is.

tubular/git_repo.py
doctoryes: The advantage of the way it's currently done is avoiding another dependency. But if it adds enough value, I'm OK with using it.

doctoryes: I'll mention this technique in the follow-on ticket.

doctoryes:
https://openedx.atlassian.net/browse/TE-1928
https://openedx.atlassian.net/browse/TE-1929
https://openedx.atlassian.net/browse/TE-1930

were created as follow-on work (minimal cloning and Python3 conversion).

I will now merge.

PR #99, opened 2017-01-12
Add tagging script to be called by GoCD.

doctoryes:
Part of: https://openedx.atlassian.net/browse/TE-1739

Also:
* Add an output artifact containing the commit SHA of the merged branch.
* Fix a bug when a GitHub user with a non-public GitHub email address tags a SHA.

### DIFF
--- a/scripts/create_tag.py
+++ b/scripts/create_tag.py
@@ -1,0 +1,123 @@
+"""
+Command-line script to create a tag for a particular SHA.
+"""
+from __future__ import unicode_literals
+
+from os import path
+import sys
+import logging
+import datetime
+import yaml
+from pytz import timezone
+import click
+
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
+
+from tubular.github_api import GitHubAPI  # pylint: disable=wrong-import-position
+from github.GithubException import GithubException  # pylint: disable=wrong-import-position
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+LOG = logging.getLogger(__name__)
+
+
+@click.command()
+@click.option(
+    '--org',
+    help='Org from the GitHub repository URL of https://github.com/<org>/<repo>',
+    default='edx'
+)
+@click.option(
+    '--repo',
+    help='Repo name from the GitHub repository URL of https://github.com/<org>/<repo>'
+)
+@click.option(
+    '--token',
+    envvar='GIT_TOKEN',
+    help='The github access token, see https://help.github.com/articles/creating-an-access-token-for-command-line-use/'
+)
+@click.option(
+    '--commit_sha',
+    help='GitHub repository commit SHA to tag'
+)
+@click.option(
+    '--input_file',
+    help='File from which to read the SHA from YAML with key "sha".'
+)
+@click.option(
+    '--branch_name',
+    help='Branch name whose HEAD commit SHA will be tagged.',
+)
+@click.option(
+    '--tag_name',
+    help='Name of the tag to use.',
+)
+@click.option(
+    '--tag_message',
+    help='Message to use when tagging.',
+)
+def create_tag(org,
+               repo,
+               token,
+               commit_sha,
+               input_file,
+               branch_name,
+               tag_name,
+               tag_message):
+    """
+    Creates a tag at a specified commit SHA with a tag name/message.
+    The commit SHA is passed in using *one* of these ways:
+        - input_file: input YAML file containing a 'sha' key
+        - commit_sha: explicitly passed-in commit SHA
+        - branch_name: HEAD sha obtained from this branch name
+    """
+    github_api = GitHubAPI(org, repo, token)
+
+    # Check for one and only one of the mutually-exclusive params.
+    if (commit_sha, input_file, branch_name).count(True) != 1:
+        LOG.error(
+            "Exactly one of commit_sha ({commit_sha}), input_file ({input_file}),"
+            " and branch_name ({branch_name}) should be specified.".format(
+                commit_sha=commit_sha,
+                input_file=input_file,
+                branch_name=branch_name
+            )
+        )
+
+    if input_file:
+        input_vars = yaml.safe_load(open(input_file, 'r'))  # pylint: disable=open-builtin
+        commit_sha = input_vars['sha']
+    elif branch_name:
+        commit_sha = github_api.get_head_commit_from_branch_name(branch_name)
+
+    # If no tag name was given, generate one using the date/time.
+    time_now = datetime.datetime.now(timezone('US/Eastern'))
+    if not tag_name:
+        tag_name = 'release-{}'.format(
+            time_now.strftime("%Y-%m-%d-%H.%M")
+        )
+
+    # If no tag message was given, generate one using the date/time.
+    if not tag_message:
+        tag_message = 'Release for {}'.format(
+            time_now.strftime("%b %d, %Y %H:%M EST")
+        )
+
+    LOG.info(
+        "Tagging commit sha {sha} as tag '{tag}' with message '{message}'".format(
+            sha=commit_sha,
+            tag=tag_name,
+            message=tag_message
+        )
+    )
+
+    try:
+        github_api.create_tag(commit_sha, tag_name, tag_message)
+    except GithubException:
+        LOG.error("Unable to create tag. Aborting")
+        raise
+
+
+if __name__ == "__main__":
+    create_tag()  # pylint: disable=no-value-for-parameter
--- a/tubular/git_repo.py
+++ b/tubular/git_repo.py
@@ -35,11 +35,7 @@ class GitRepo(object):
     def __init__(self, clone_url):
         # Verify and save the clone URL.
         parsed = urlparse(clone_url)
-        self.clone_url = '{scheme}://{netloc}{path}'.format(
-            scheme=parsed.scheme,
-            netloc=parsed.netloc,
-            path=parsed.path
-        )
+        self.clone_url = parsed.geturl()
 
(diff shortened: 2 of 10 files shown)


scripts/check_pr_tests_status.py
```diff
@@ -1,71 +1,97 @@
...
+            sys.argv[0], pr_number, "success" if status_success else "failed"
+        ))
     elif commit_hash:
         status_success = gh_utils.is_commit_successful(commit_hash)
-        print u"{}: Combined status of commit hash {} is {}.".format(
-            sys.argv[0], commit_hash, u"success" if status_success else u"failed"
-        )
+        LOG.info("{}: Combined status of commit hash {} is {}.".format(
+            sys.argv[0], commit_hash, "success" if status_success else "failed"
+        ))
```
macdiesel: nit, you could probably make this a bit more dry and eliminate some of the repetitive logging.

scripts/create_tag.py
```diff
@@ -0,0 +1,106 @@
...
+    help='The github access token, see https://help.github.com/articles/creating-an-access-token-for-command-line-use/'
+)
+@click.option(
+    '--commit_sha',
+    help='GitHub repository commit SHA to tag'
+)
+@click.option(
+    '--input_file',
+    help='File from which to read the SHA from YAML with key "sha".',
+    default='target/merge_branch_sha.yml'
```
macdiesel: seems strange to provide the default input file in tubular as much of this wiring is done outside this repository.

scripts/create_tag.py
```diff
@@ -0,0 +1,106 @@
...
+               tag_message):
+    """
+    Creates a tag at a specified commit SHA with a tag name/message.
+    First, uses the commit SHA specified via the "commit_sha" parameter.
+    Else, uses the "sha" value from the specified input YAML file.
+    """
+    github_api = GitHubAPI(org, repo, token)
+
+    if not commit_sha:
+        input_vars = yaml.safe_load(open(input_file, 'r'))  # pylint: disable=open-builtin
```
macdiesel: is it worth putting the reading of the file in a try catch block to provide better error messaging?

scripts/create_tag.py
```diff
@@ -0,0 +1,106 @@
...
+    # If no tag name was given, generate one using the date/time.
+    time_now = datetime.datetime.now(timezone('US/Eastern'))
+    if not tag_name:
+        tag_name = 'release-{}'.format(
+            time_now.strftime("%Y-%m-%d-%H.%M")
+        )
+
+    # If no tag message was given, generate one using the date/time.
+    if not tag_message:
+        tag_message = 'Release for {}'.format(
```
macdiesel, line 103: Should date tagging come from a method in here? https://github.com/edx/tubular/blob/91ba52b864a36e110a244031cfc9b157b23bee82/tubular/github_api.py#L55-L55

tubular/git_repo.py
```diff
@@ -35,8 +35,9 @@ class GitRepo(object):
     def __init__(self, clone_url):
         # Verify and save the clone URL.
         parsed = urlparse(clone_url)
-        self.clone_url = '{scheme}://{netloc}{path}'.format(
+        self.clone_url = '{scheme}{divider}{netloc}{path}'.format(
```
macdiesel: 👍

tubular/git_repo.py
```diff
@@ -173,7 +196,9 @@ def merge_branch(repo_url,
             repo.track_branch(target_branch)
             repo.merge(source_branch, ff_only)
             repo.push(target_branch)
+            merge_sha = repo.get_head_sha(target_branch)
     except subprocess.CalledProcessError as exc:
         raise GitMergeFailed(repr(exc.cmd))
     finally:
         repo.cleanup()
+    return merge_sha
```
macdiesel, line 202: Needs returns statement added to docs.

macdiesel (commented):
First pass complete.

tubular/git_repo.py
macdiesel: Should geturl() yield the same results?

scripts/check_pr_tests_status.py
doctoryes: Will do.

scripts/create_tag.py
doctoryes: Agreed - I'll remove it.

scripts/create_tag.py
doctoryes: If I wanted to avoid a stack trace altogether, yes. But a stack trace will clearly tell whether the file couldn't be found/opened as opposed to the YAML couldn't be parsed. Since there's no recovery, I'm inclined to leave it like this?

tubular/git_repo.py
doctoryes: Will do.

tubular/git_repo.py
doctoryes: I think so - and it'd simplify the code. I'll try it out.

scripts/create_tag.py
doctoryes: The code uses the same datetime for the tag name and message. So I'd need to pass a datetime into whatever method/methods were created. And ultimately it'd be a single `strftime` call. If other code needed the same tagging, then I'd say yes. But for now, having it here is simpler.

doctoryes:
@macdiesel @cpennington This PR is ready for final review.

If you'd like to see the code in operation, check here:
https://gocd.tools.edx.org/go/pipelines/value_stream_map/edxapp_branch_cleanup/15

scripts/create_tag.py
macdiesel: 👍

doctoryes:
A note: I used the backoff Python module to perform the retries:
https://github.com/litl/backoff

It's more full-featured than our existing retry code, containing all these TODOS:
https://github.com/edx/tubular/blob/master/tubular/utils/retry.py#L61-L64

scripts/create_tag.py
```diff
@@ -0,0 +1,103 @@
...
+
+    if not commit_sha:
+        input_vars = yaml.safe_load(open(input_file, 'r'))  # pylint: disable=open-builtin
+        commit_sha = input_vars['sha']
+
+    # If no tag name was given, generate one using the date/time.
+    time_now = datetime.datetime.now(timezone('US/Eastern'))
+    if not tag_name:
+        tag_name = 'release-{}'.format(
+            time_now.strftime("%Y-%m-%d-%H.%M")
```
doctoryes, line 98: Note: I included the hour/minute at the end of the tag in order to ensure the tag won't be the same as previous tags. Everybody OK with this? Anybody think I need seconds also?

cpennington:
@doctoryes: a quick note on `backoff`: I just added as part of #108 the ability to use backoff to shoot for a total wait time (without having to just pick a number of attempts).

scripts/create_tag.py
cpennington: No, no need for seconds.

scripts/create_tag.py
cpennington: Minor note: you can actually specify the time format inside the `{}`, if you're passing in a `datetime` object: https://docs.python.org/2/library/datetime.html#datetime.date.__format__.

tubular/github_api.py
```diff
@@ -119,9 +160,76 @@ def get_head_commit_from_pull_request(self, pr_number):
...
+            github.GithubException.GithubException: If the response fails.
+            github.GithubException.UnknownObjectException: If the branch does not exist
+        """
+        return self.get_commit_combined_statuses(
+            self.get_head_commit_from_pull_request(pr_number)
+        ).state.lower() == 'success'
+
+    @backoff.on_predicate(
+        _constant_with_initial_wait,
+        lambda x: (x != 'success' and x != 'failure'),
```
cpennington: Nit: Maybe `x not in ('success', 'failure')`?

tubular/github_api.py
```diff
@@ -119,9 +160,76 @@ def get_head_commit_from_pull_request(self, pr_number):
...
+        max_tries=_envvar_get_int("MAX_PR_TEST_POLL_TRIES", MAX_PR_TEST_TRIES_DEFAULT),
+        initial_wait=_envvar_get_int("PR_TEST_INITIAL_WAIT_INTERVAL", PR_TEST_INITIAL_WAIT_INTERVAL_DEFAULT),
+        interval=_envvar_get_int("PR_TEST_POLL_INTERVAL", PR_TEST_POLL_INTERVAL_DEFAULT),
```
cpennington: Can these be specified via commandline arguments to the polling script? We're a bit inconsistent with what things can come through the environment and what through the commandline options.

cpennington (approved):

tubular/github_api.py
doctoryes: Better - will do.

tubular/github_api.py
doctoryes: The problem: These vars are read in the decorator at run-time. Is there a better way to pass parameters to a decorator at call-time?

doctoryes:
@cpennington @macdiesel Ready for a final review.

scripts/create_tag.py
```diff
@@ -0,0 +1,123 @@
...
+    Creates a tag at a specified commit SHA with a tag name/message.
+    The commit SHA is passed in using *one* of these ways:
+        - input_file: input YAML file containing a 'sha' key
+        - commit_sha: explicitly passed-in commit SHA
+        - branch_name: HEAD sha obtained from this branch name
+    """
+    github_api = GitHubAPI(org, repo, token)
+
+    # Check for one and only one of the mutually-exclusive params.
+    if (commit_sha, input_file, branch_name).count(True) != 1:
```
cpennington, line 78: Huh... I'm surprised this works.

cpennington (approved):

PR #102, opened 2017-01-17
Add logging about messages being posted to github

cpennington:


macdiesel (approved):

scripts/message_prs_in_range.py
```diff
@@ -84,6 +84,7 @@ def message_pull_requests(org,
 
     api = GitHubAPI(org, repo, token)
     for pull_request in api.get_pr_range(base_sha, head_sha):
+        LOG.info("Posting message type %r to %d.", message_type, pull_request.number)
```
doctoryes: Why not use Python's format here?

cpennington: Because it's a logging message. Logging will do the interpolation for you (and that's important in certain cases where you have a different logger).

doctoryes (approved):
Once quality fixed.

PR #105, opened 2017-01-18
Allow message_prs_in_range to read from an AMI tag yaml file

cpennington:


scripts/message_prs_in_range.py
```diff
@@ -34,10 +35,18 @@
...
-    required=True,
     help=u'The BASE SHA of the range',
 )
 @click.option(
+    u'--base_ami_tags',
+    help=u'A YAML file with tags for the base_ami should be used as the baseline for these messages',
+    type=click.File(),
+)
+@click.option(
+    u'--ami_tag_app',
```
macdiesel, line 46: shouldn't this be the same as the repo option?

cpennington: Well... there's a version listed for `edxapp`, too, which is really just edx_platform. Also, I'm not sure how edx_platform_private gets tagged.

macdiesel: That's right the tagging isn't consistent with the repository names. =(

macdiesel (approved):

scripts/message_prs_in_range.py
```diff
@@ -82,6 +95,14 @@ def message_pull_requests(org,
         u'rollback': u'message_pr_release_canceled'
     }
 
+    if base_sha is None and base_ami_tags and ami_tag_app:
+        ami_tags = yaml.safe_load(base_ami_tags)
+        tag = u'version:{}'.format(ami_tag_app)
+        version = ami_tags[tag]
+        repo_url, _, base_sha = version.partition(u' ')
```
doctoryes, line 102: Just curious: Why `partition` over `split` here?

cpennington: I tend to prefer partition when I know I'm only going to be splitting on the first occurance (rather than using something like `[repo_url, base_sha] = version.split(' ', 1)`)

PR #107, opened 2017-01-19
TE-1936 - Add method to generate a diff between two commits

davec-edx:
Added a method to generate a simple URL to do a diff (in github) against two commit SHAs.  If the
URL is not valid (invalid domain, space in one of the slugs, etc., the method will throw an InvalidUrlException.


davec-edx:
@macdiesel please review

macdiesel (approved):

PR #112, opened 2017-01-24
Clone from the target_branch, when merging branches

cpennington:


doctoryes (approved):

doctoryes:
👍  After tests are fixed....

PR #114, opened 2017-01-25
Convert Tubular to Python 3

cpennington:
Until edx-gomatic is converted to call the tubular scripts using the shebang, tubular should stay python2/3 compatible.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,16 @@
 [tox]
-envlist = py27, py34, py35
+envlist = py{27,34,35}-{test,quality}
 skip_install = True
 skipsdist = True
 
 [testenv]
+envdir =
+    py27: {toxworkdir}/py27
+    py34: {toxworkdir}/py34
+    py35: {toxworkdir}/py35
 deps=-rrequirements/test_requirements.txt
-commands=pytest {posargs:-n auto}
+commands=
+    test: pytest {posargs:-n auto}
+    quality: pep8 --config=.pep8 tubular scripts admin
+    py27-quality: pylint --py3k tubular scripts admin
+    quality: pylint tubular scripts admin
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,8 @@
+[tox]
+envlist = py27, py34, py35
+skip_install = True
+skipsdist = True
+
+[testenv]
+deps=-rrequirements/test_requirements.txt
+commands=pytest {posargs:-n auto}
--- a/tubular/tests/test_asgard.py
+++ b/tubular/tests/test_asgard.py
@@ -683,7 +683,7 @@ class TestAsgard(unittest.TestCase):
         self._mock_asgard_pending_delete(req_mock, [asg])
         self.assertTrue(asgard.is_asg_pending_delete(asg))
 
-        self.assertEquals(None, asgard.disable_asg(asg))
+        self.assertEqual(None, asgard.disable_asg(asg))
 
     def test_disable_asg_does_not_exist(self, req_mock):  # pylint: disable=unused-argument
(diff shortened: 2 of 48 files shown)


tubular/tests/test_asgard.py
```diff
@@ -29,336 +29,328 @@
...
+            "status": "deleted",
+            "launchingSuspended": False,
+            "desiredCapacity": 4,
+            "minSize": 4
+        },
+        "clusterName": "app_cluster"
+    }
+
+
+def deleted_asg_not_in_progress(name):
```
macdiesel, line 196: nice change. 👍

tubular/tests/test_asgard.py
```diff
@@ -381,228 +373,188 @@ def test_clusters_for_asgs(self):
...
 
-    @httpretty.activate
-    def test_new_asg(self):
+    def test_new_asg(self, req_mock):
         task_url = "http://some.host/task/1234.json"
         cluster = "loadtest-edx-edxapp"
         ami_id = "ami-abc1234"
 
-        def post_callback(request, uri, headers):  # pylint: disable=unused-argument
+        def post_callback(request, context):  # pylint: disable=unused-argument
```
macdiesel, line 650: I think you can remove the pylint annotation.

tubular/tests/test_asgard.py
```diff
@@ -649,216 +601,202 @@ def test_new_asg_failure(self,
...
             Callback method for POST.
             """
             self.assertEqual('POST', request.method)
             expected_request_body = {"name": [cluster], "imageId": [ami_id]}
             expected_querystring = {"asgardApiToken": ['dummy-token']}
 
-            self.assertEqual(expected_request_body, request.parsed_body)
-            self.assertEqual(expected_querystring, request.querystring)
-            return create_response_code, create_response_headers, create_response_body
+            parsed_body = six.moves.urllib.parse.parse_qs(request.text)
```
macdiesel: from six.moves import urllib to shorten this up?

tox.ini
```diff
@@ -0,0 +1,16 @@
...
+
+[testenv]
+envdir =
+    py27: {toxworkdir}/py27
+    py34: {toxworkdir}/py34
+    py35: {toxworkdir}/py35
+deps=-rrequirements/test_requirements.txt
+commands=
+    test: pytest {posargs:-n auto}
+	quality: pep8 --config=.pep8 tubular scripts admin
```
feanil: Is this spacing correct or relevant?  I don't see anything about it in the tox docs.

tox.ini
```diff
@@ -0,0 +1,16 @@
...
+	quality-py27: pylint --py3k tubular scripts admin
```
feanil: Am I correct in assuming this will only get run on python 2.7?  Does the order in the naming of the key not matter?

feanil (commented):
Generally looks good, just a couple of questions around tox.

tox.ini
cpennington: Whoah, strange.

tox.ini
cpennington: The order doesn't matter (it just does set comparison on the individual factors). But, having answered this question multiple times, I'm just going to use a consistent order.

tubular/confluence_api.py
```diff
@@ -0,0 +1,9 @@
+"""Stuff"""
```
doctoryes: Oops?

doctoryes (approved):
Once confluence_api.py is removed.

tubular/confluence_api.py
cpennington: Drat. Didn't mean to check that file in.

PR #117, opened 2017-01-30
Add retry attempts to ec2 calls when throttled.

macdiesel:
@edx/pipeline-team All ec2 methods have been decorated so they will be retried when throttling conditions are returned from AWS.

### DIFF
--- a/tubular/ec2.py
+++ b/tubular/ec2.py
@@ -5,11 +5,13 @@ when we deploy using asgard.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import os
 import logging
 import time
 from datetime import datetime, timedelta
+import backoff
 import boto
-from boto.exception import EC2ResponseError
+from boto.exception import EC2ResponseError, BotoServerError
 from boto.ec2.autoscale.tag import Tag
 from tubular.utils import EDP, WAIT_SLEEP_TIME
 from tubular.exception import (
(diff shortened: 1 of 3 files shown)


tubular/ec2.py
```diff
@@ -23,8 +25,26 @@
...
+    Args:
+        ex (boto.exception.BotoServerError):
+
+    Returns:
+        False if the throttling string is not found.
+    """
+    return not (str(ex.status) == "400" and ex.body and '<Code>Throttling</Code>' in ex.body)
+
+
+@backoff.on_exception(backoff.expo,
```
cpennington, line 45: It might be worth passing a scaling factor (`scale`, I think), so that we wait more than `1, 2, 4, 8, 16` seconds for the throttling to run out.

macdiesel: This will now retry at these intervals: `1.0, 3.0, 9.0, 27.0, 81.0`

tubular/ec2.py
```diff
@@ -411,6 +496,11 @@ def wait_for_in_service(all_asgs, timeout):
     raise TimeoutException("Some instances in the following ASGs never became healthy: {}".format(asgs_left_to_check))
 
 
+@backoff.on_exception(backoff.expo,
+                      BotoServerError,
+                      max_tries=MAX_ATTEMPTS,
+                      giveup=giveup_if_not_throttling,
+                      factor=RETRY_FACTOR)
 def wait_for_healthy_elbs(elbs_to_monitor, timeout):
```
doctoryes: This call already waits for 10 minutes IIRC. Should the MAX_ATTEMPTS perhaps be different for calls that have their own waiting/polling built-in?

macdiesel: good catch, I will remove the decorator from this method.

macdiesel:
@doctoryes I have either removed the retry decorator from functions that do not make calls to AWS or refactored code so only the portions that call to AWS are retried. PTAL.

tubular/ec2.py
```diff
@@ -516,14 +473,33 @@ def wait_for_healthy_elbs(elbs_to_monitor, timeout):
...
     """
+
+    @backoff.on_exception(backoff.expo,
+                          BotoServerError,
+                          max_tries=MAX_ATTEMPTS,
+                          giveup=giveup_if_not_throttling,
+                          factor=RETRY_FACTOR)
+    def _get_elb_health(selected_elb):
+        """
+        Moved This call to get_instance_health so it can be wrapped with a retry
```
doctoryes: Nit: More like a commit msg than a docstring explanation.

doctoryes (approved):

PR #119, opened 2017-02-01
Remove jitter from wait times when waiting for jenkins test runs

cpennington:


macdiesel (approved):

tubular/tests/test_jenkins.py
```diff
@@ -73,6 +76,25 @@ def test_max_timeout(self, base, factor, timeout, expected_max_tries, expected_w
...
+        call_count = 0
+
+        @backoff.on_predicate(
+            wait_gen,
+            max_tries=max_tries,
+            on_backoff=print,
+            jitter=None,
+        )
+        def falsy():
+            nonlocal call_count
```
jzoldak: tox is complaining about this syntax. 
Another approach would be to use a mock instead of an actual method, and assert on the call_count of the mock.

jzoldak (approved):

PR #120, opened 2017-02-02
Add a script to publish a release page

cpennington:
Accepts one or more pairs of old/new AMIs, and generates and publishes
Confluence pages.

[TE-1935]

### DIFF
--- a/tubular/confluence_api.py
+++ b/tubular/confluence_api.py
@@ -1,0 +1,305 @@
+u"""
+Functions for interacting with the Confluence API and for rendering
+Confluence pages.
+"""
+from __future__ import absolute_import
+
+from collections import namedtuple
+from enum import Enum
+import logging  # pylint: disable=wrong-import-order
+import re  # pylint: disable=wrong-import-order
+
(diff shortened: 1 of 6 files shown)


tubular/confluence_api.py
```diff
@@ -0,0 +1,253 @@
...
+        E.TABLE(
+            E.CLASS("wrapped"),
+            E.TBODY(
+                E.TR(
+                    E.TH(u"Merged By"),
+                    E.TH(u"Author"),
+                    E.TH(u"Title"),
+                    E.TH(u"PR"),
+                    E.TH(u"JIRA"),
+                    E.TH(u"Release Notes?"),
```
doctoryes, line 171: Will you also include the GoCD pipeline run instance link in here?

cpennington: I was considering including it in the AMI. But I wasn't sure it would actually be useful. Although, maybe a link to the overview from the pre-release pipeline?

cpennington: Alright, added that as an option.

doctoryes (approved):

PR #122, opened 2017-02-08
Fail jenkins_trigger_build.py if the build status isn't the expected …

cpennington:
…result


macdiesel (approved):

jzoldak (commented):
ah, I see now what it was. nice. 👍

PR #123, opened 2017-02-09
Use pbr to setup.py, and make all scripts available via installation

cpennington:
This should be backwards compatible with our current calling convention, but will let us switch to just pip installing the tubular package from git, and then calling the installed scripts. That, in turn, will let us get rid of the `scripts/*.py` symlinks, and also any `sys.path` hackery.


cpennington:
I should note that this method of installation will work whether or not we have a virtualenv.

macdiesel (approved):
:

PR #124, opened 2017-02-09
Cale/fix sys path hacks

cpennington:
This depends on #123. It also requires that we go through edx/edx-gomatic and scrub any references to `script/some_script.py` before merging it.

### DIFF
--- a/tubular/scripts/asgard_deploy.py
+++ b/tubular/scripts/asgard_deploy.py
@@ -9,16 +9,12 @@ from __future__ import unicode_literals
 from __future__ import print_function
 
 import io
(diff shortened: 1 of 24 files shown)


macdiesel (approved):
I would assume we would want to merge in edx-gomatic/pull/171 to go along with this with the corrected pathing?

cpennington:
Yeah, we need more prep work in edx-gomatic before this can land.

cpennington:
I don't think this has any more dependencies (I think we can merge it before we scrub all uses of `script/foo.py` in edx-gomatic).

PR #127, opened 2017-02-14
Use tubular.scripts rather than tubular.script when pointing to execu…

cpennington:
…table scripts


cpennington:
I'm going to go rogue and merge this w/o thumbs, to try and get a release to stage out before the end of the day.

PR #130, opened 2017-02-15
Bail on empty lists. Fix logging upon some errors.

doctoryes:
A bug was found in this GoCD run:
https://gocd.tools.edx.org/go/tab/build/detail/deploy-edx-ami/650/deploy_ami/1/deploy_ami_job
where an empty list of ELBs was sent to monitor - and the code behaved poorly.
Instead, skip monitoring upon empty lists of ASGs/ELBs.

Also, print exception backtraces when bare exceptions are caught. And add a log.

### DIFF
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -8,7 +8,6 @@ from datetime import datetime, timedelta
 import os
 import logging
 import time
-import traceback
 import copy
 from collections import defaultdict
 import requests
@@ -603,7 +602,7 @@ def deploy(ami_id):
         except:
             msg = "ASG creation failed for cluster '{}' but succeeded for cluster(s) {}."
             msg = msg.format(cluster, new_clustered_asgs.keys())
-            LOG.error(msg)
+            LOG.exception(msg)
             raise
 
     new_asgs = [asgs[0] for asgs in new_clustered_asgs.values()]
@@ -694,7 +693,7 @@ def _red_black_deploy(
                 try:
                     _disable_cluster_asg(cluster, asg)
                 except:  # pylint: disable=bare-except
-                    LOG.warning(failure_msg.format(asg))
+                    LOG.warning(failure_msg, asg, exc_info=True)
 
     elbs_to_monitor = []
     newly_enabled_asgs = defaultdict(list)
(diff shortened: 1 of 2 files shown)


tubular/asgard.py
```diff
@@ -604,6 +604,7 @@ def deploy(ami_id):
             msg = "ASG creation failed for cluster '{}' but succeeded for cluster(s) {}."
             msg = msg.format(cluster, new_clustered_asgs.keys())
             LOG.error(msg)
+            LOG.error(traceback.format_exc())
```
cpennington: Use `LOG.exception(msg)`, and it'll format the exception for you.

tubular/asgard.py
```diff
@@ -695,6 +696,7 @@ def _disable_clustered_asgs(clustered_asgs, failure_msg):
                     _disable_cluster_asg(cluster, asg)
                 except:  # pylint: disable=bare-except
                     LOG.warning(failure_msg.format(asg))
+                    LOG.error(traceback.format_exc())
```
cpennington: Here, you can use `LOG.warning(failure_msg.format(asg), exc_info=True)`.

Also, it's usually preferable to not format messages before passing them into the logger, and instead let the logger do the formatting.

tubular/asgard.py
```diff
@@ -725,6 +727,7 @@ def _disable_clustered_asgs(clustered_asgs, failure_msg):
         ec2.wait_for_healthy_elbs(elbs_to_monitor, 600)
     except:  # pylint: disable=bare-except
         LOG.info("Some ASGs are failing ELB health checks. Disabling traffic to all new ASGs.")
+        LOG.error(traceback.format_exc())
```
cpennington: As above, w/ `exc_info=True`.

tubular/ec2.py
```diff
@@ -308,6 +308,12 @@ def asgs_for_edp(edp, filter_asgs_pending_delete=True):
             if group_edp == edp:
                 matching_groups.append(group.name)
 
+    LOG.info("Returning {num} ASGs for EDP {env}-{dep}-{play}.".format(
+        num=len(matching_groups),
+        env=edp.environment,
+        dep=edp.deployment,
+        play=edp.play
+    ))
```
cpennington: As above re: formatting.

cpennington (approved):

cpennington:
(As a general note: you can use `git commit --fixup <some sha, like HEAD>`, and then `git rebase --autosquash`, and get fixup commits with squashing for nearly free.)

PR #131, opened 2017-02-15
Feanil/approve stage

feanil:
@edx/pipeline-team

### DIFF
--- a/tubular/scripts/approve_stage.py
+++ b/tubular/scripts/approve_stage.py
@@ -1,0 +1,63 @@
+#! /usr/bin/env python3
+
+"""
+Command-line script to click the "manual" gate in gocd.
+"""
+
+# pylint: disable=invalid-name
+from __future__ import absolute_import
+from __future__ import unicode_literals
+from __future__ import print_function
+
+import os
+import sys
+import click
+from yagocd import Yagocd
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+@click.command()
+@click.option(
+    '--go-server-url',
+    envvar='GOCD_SERVER_URL',
+    help='The URL for the GoCD server to interact with.',
+)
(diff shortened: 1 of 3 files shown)


tubular/scripts/approve_stage.py
```diff
@@ -0,0 +1,63 @@
...
+from __future__ import unicode_literals
+from __future__ import print_function
+
+import os
+import sys
+import click
+from yagocd import Yagocd
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
macdiesel, line 18: @feanil This sys pathing stuff can go and the scripts are now called directly because they are in the path and tubular is installed.

cpennington: You do need to add an entry to `setup.cfg` to make them callable, though.

tubular/scripts/approve_stage.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@click.command()
+@click.option(
+    '--go-server-url',
+    envvar='GOCD_SERVER_URL',
```
doctoryes: `envvar`: We've decided to use it as little as possible - really only for secrets and things we never want on the command-line. Seems like most all these `envvar`s should be removed.

doctoryes: If any of these params are required (and it seems like most of these are), they should be marked with `required=True`.

macdiesel: No magic variable passing! Huzzah!

macdiesel: I think it's fine if you pass it in via the command line if it's a secret, it will be obscured anyway.

feanil:
@edx/pipeline-team ready for re-review.

feanil:
@cpennington putting the pathing logic back for now until your new stuff fully lands.

doctoryes (approved):

PR #132, opened 2017-02-16
Convert to using pytest-pylint and pytest-pep8 for linting. Also, rem…

cpennington:
…ove testing of python2.7


doctoryes:
I'm hesitant to approve this PR until instructions exist (or at least are pointed to) on how to set up local `make test`ing of tubular. Currently, Python 2.7 testing is the only tox testing that works for me - I've tried setting up `pyenv` and such to get Python 3.{4,5} to work - but no luck yet.

cpennington:
Yeah, that makes sense. I'll add that documentation (or a link to documentation that already exists).

cpennington:
@doctoryes: Can you try the instructions I just added to test using python 3.4 and 3.5?

doctoryes (commented):
The instructions work. However `make test` does *not*. Could that be fixed as well?

cpennington:
@doctoryes: Yeah, should be fixed now.

doctoryes (approved):

PR #134, opened 2017-02-16
Use a sequence not a generator for *args

cpennington:
Should fix https://gocd.tools.edx.org/go/tab/build/detail/STAGE_edxapp_M-D/101/message_pr_stage/1/publish_wiki_job. Not sure why it wasn't flagged in tests.


cpennington:
Nevermind. The error is http://bugs.python.org/issue4806.

PR #142, opened 2017-02-22
TE-1987: Add tubular to add a remote to an existing repository

cpennington:


tubular/tests/test_git_repo.py
```diff
@@ -18,60 +18,69 @@ class GitRepoTestCase(TestCase):
...
+    @patch('tubular.git_repo.Repo', autospec=True)
+    def test_clone_failure(self, mock_repo, mock_rmtree):
+        """
+        Tests failing to merge a branch.
+        """
+        mock_repo.clone_from.side_effect = GitCommandError('cmd', 1)
+
+        with self.assertRaises(GitCommandError):
+            merge_branch('git@github.com:edx/tubular.git', 'foo', 'bar')
+            mock_rmtree.assert_not_called()
```
doctoryes: Does this line actually get called after the exception is raised above?

doctoryes (approved):

tubular/tests/test_git_repo.py
cpennington: Oops, no, it doesn't. Need to dedent it.

PR #143, opened 2017-02-23
Automate finding which pipeline instance to advance to production.

doctoryes:
https://openedx.atlassian.net/browse/TE-1984

- Remove old GoCD code.
- Add yagocd as a requirement.
- Add time to return from `default_expected_release_date()`.
- Add module to interface with GoCD API using yagocd.
- Add tests for methods added to GoCD module.


tubular/gocd_api.py
```diff
@@ -0,0 +1,97 @@
...
+        Given:
+            - the name of a pipeline to manually advance (the advancement pipeline)
+            - a datetime representing the current time (typically the production release time)
+        find the advancement pipeline that should be advanced/deployed to production.
+
+        The algorithm:
+        - Query the value stream map containing the upstream pipeline materials of the advancement pipeline.
+        - Find the initial upstream pipeline material.
+        - Check the time at which the first job of the first stage was triggered.
+        - If the job was triggered before the last release time relative to the passed-in time, found.
```
macdiesel, line 59: Currently the last release time is calculated, but it's possible that there could have been a release after this point (e.g. a hotfix late in the evening). How would this scenario effect this algorithm?

tubular/gocd_api.py
```diff
@@ -0,0 +1,97 @@
...
+        - If not, keep going backwards into pipeline history until found.
```
macdiesel, line 60: Should we put a cap on how far back in time the algorithm will look? Could this possibly trigger a pipeline that happened before the last release?

tubular/gocd_api.py
doctoryes: Good point. Our process doesn't currently account for this scenario. To support it, I'd need to add the ability for this code to look for a specific pipeline stage (or stages) and find the last (or most recent) time it (they) ran successfully, in order to determine the last successful production deploy. Do we think this feature is essential in this iteration?

tubular/gocd_api.py
doctoryes: If you pass in a past date, the code will go back in history in an attempt to find the pipeline that meets the timing criteria. The history goes backwards in time starting from the most recent - and the iteration stops when the pipeline is found. So I don't think it'll pick a release that happened before the last release. But the code could potentially walk the entire history of a pipeline, which wouldn't be great.

tubular/gocd_api.py
macdiesel: What would be the ramifications be if we did not add this? 

Perhaps the better thing to do is to look for the last successful stage deploy before the pipeline is suppose to open for the day?

cpennington: You could just stop iterating the advancement pipeline the first time you find a build that has already been advanced.

tubular/gocd_api.py
cpennington: Couldn't we just cut off the iteration as soon as we find a build that is too old?

tubular/gocd_api.py
```diff
@@ -0,0 +1,97 @@
...
+        relative_time = relative_to if relative_to else datetime.now(utc_zone)
+        last_release_time = default_expected_release_date(
+            (relative_time - timedelta(days=1)).astimezone(utc_zone)
+        )
+
+        # Go backwards in advancement pipeline history, starting with the most recent run.
+        for advancement_pipeline in self.client.pipelines.full_history(advance_pipeline_name):
+            # Get the full instance information for the initial pipeline determined by the value stream map.
+            vsm = advancement_pipeline.value_stream_map()
+            initial_pipeline_inst = self.client.pipelines.get(vsm[0].data.name, vsm[0].data.counter)
```
cpennington, line 91: How is the value-stream-map sorted? By date? I'm thinking about the possibility (in the future?) of having two different builds triggered from the same material, that then join into a single downstream... Which of the two `initial` pipelines would be first? Maybe the tubular code should be searching based on pipeline names?

tubular/tests/test_github.py
```diff
@@ -418,10 +419,14 @@ def test_extract_short(self, message, expected):
...
-        (datetime(2017, 2, 15, 4, 0, 0), date(2017, 2, 15)),  # on Wednesday morning, the release date is Wednesday
-        (datetime(2017, 2, 15, 11, 0, 0), date(2017, 2, 16)),  # on Wednesday afternoon, the release date is Thursday
+        # on Friday morning, the release date is Friday
+        (datetime(2017, 2, 17, 4, 0, 0), datetime.combine(datetime(2017, 2, 17), RELEASE_CUTOFF)),
+        # on Friday afternoon, the release date is Monday
+        (datetime(2017, 2, 17, 11, 0, 0), datetime.combine(datetime(2017, 2, 20), RELEASE_CUTOFF)),
+        # on Wednesday morning, the release date is Wednesday
+        (datetime(2017, 2, 15, 4, 0, 0), datetime.combine(datetime(2017, 2, 15), RELEASE_CUTOFF)),
+        # on Wednesday afternoon, the release date is Thursday
+        (datetime(2017, 2, 15, 11, 0, 0), datetime.combine(datetime(2017, 2, 16), RELEASE_CUTOFF)),
```
cpennington: Maybe put the `combine(..., RELEASE_CUTOFF)` inside the test function, since it doesn't vary across tests?

cpennington (approved):

tubular/gocd_api.py
doctoryes: @cpennington It does make sense to stop looking backward once we get to a production-advanced value-stream-map - I'll change the code to do that.

However, @macdiesel 's point was that if we do a hotfix between yesterday's release and today's, this code won't take it into account in its selection. Have we had to deal with this issue in the manual process so far? If so, how?

tubular/gocd_api.py
doctoryes: The code essentially does that now. But since any date can be passed-in, the date could be older than any of the pipeline instances. The code doesn't handle that smartly now - just continues back into history until history expires.

tubular/tests/test_github.py
doctoryes: Will do.

tubular/gocd_api.py
doctoryes: It'd be easy enough to support this now or in the future. I'm leaning towards keeping it simple for now and keeping it like this - but I'm open to doing it. This selection code will need further changes to support the new ECOM-style pipelines - perhaps we should just wait until then?

tubular/gocd_api.py
cpennington: In general, after a hotfix, I've had to force-push release-candidate back to whatever it was.

tubular/gocd_api.py
cpennington: I see.

doctoryes:
@cpennington @macdiesel Could I please get a re-review here with your thumbs *or* comments?

tubular/gocd_api.py
```diff
@@ -0,0 +1,133 @@
...
+            relative_to: Datetime relative to which the release should occur. If None, use the current datetime.
+
+        Returns:
+            yagocd.resources.pipeline.PipelineInstance: Instance of advancement pipeline to advance.
+        """
+        def has_advanced(pipeline_instance):
+            """
+            Check to see if a pipeline from a value stream map has been advanced.
+            """
+            return pipeline_instance.stage('manual_verification').data['result'] != 'Unknown'
```
cpennington: This probably doesn't need to change right now, but this stage name hard-coding will need to change if (say) ecommerce wants to use this on their new pipeline.

tubular/gocd_api.py
```diff
@@ -0,0 +1,133 @@
...
+
+            # Find the trigger timestamp from the first job in the first stage of the first pipeline.
+            # Trigger timestamp is in milliseconds since the epoch time - convert to seconds.
+            trigger_time = initial_pipeline_inst.stages()[0].jobs()[0].data.get('scheduled_date') / 1000
+
+            # Convert the trigger timestamp to a UTC datetime.
+            utc_trigger_time = datetime.utcfromtimestamp(trigger_time).replace(tzinfo=utc_zone)
+
+            # Was the initial pipeline in the value stream map was triggered before the last release time?
+            if utc_trigger_time < last_release_time:
```
macdiesel: Could we use a different variable name for last_release_time? This caused me some confusion as I was trying to parse the code. At first glance I thought you were looking for the "last release time" which in my mind was the last successful release. What you are really comparing here is the current expected release time.

Perhaps something like current_release_time or expected_release_time.

tubular/tests/test_gocd.py
```diff
@@ -0,0 +1,219 @@
...
+                            'manual_verification_edxapp_prod_early_ami_build'
+                        )
+                    self.assertEqual(found_pipeline.data.name, 'manual_verification_edxapp_prod_early_ami_build')
+                    self.assertEqual(found_pipeline.data.counter, 148)
+
+    INSTANCE_FIND_TEST_DATA = (
+        (VALID_JOB_TRIGGER_TIME_MS, ('Unknown', 'Unknown'), None),
+        (VALID_JOB_TRIGGER_TIME_MS, ('Passed', 'Passed'), AdvancementPipelineNotFound),
+        (VALID_JOB_TRIGGER_TIME_MS, ('Unknown', 'Passed'), AdvancementPipelineAlreadyAdvanced),
+        (INVALID_JOB_TRIGGER_TIME_MS, ('Unknown', 'Unknown'), AdvancementPipelineNotFound)
```
macdiesel: nit: should we add a case where there are no prior runs?

macdiesel (approved):
A few minor things.

tubular/gocd_api.py
doctoryes: Done.

tubular/tests/test_gocd.py
doctoryes: Done.

tubular/gocd_api.py
```diff
@@ -60,27 +60,30 @@ def fetch_pipeline_to_advance(self, advance_pipeline_name, relative_to=None):
...
             Check to see if a pipeline from a value stream map has been advanced.
             """
-            return pipeline_instance.stage('manual_verification').data['result'] != 'Unknown'
+            return pipeline_instance.stage(stage_name).data['scheduled']
 
         # Compute the last release time (in UTC) relative to the passed-in time -or- now.
         utc_zone = tz.gettz('UTC')
         relative_time = relative_to if relative_to else datetime.now(utc_zone)
-        last_release_time = default_expected_release_date(
+        current_release_time = default_expected_release_date(
```
cpennington: `previous_release_cutoff`?

tubular/gocd_api.py
```diff
@@ -60,27 +60,30 @@ def fetch_pipeline_to_advance(self, advance_pipeline_name, relative_to=None):
...
             (relative_time - timedelta(days=1)).astimezone(utc_zone)
```
cpennington: I don't think you need to use `astimezone` here. Because all times involved are timezone aware, python should just do the right thing.

tubular/gocd_api.py
```diff
@@ -60,27 +60,30 @@ def fetch_pipeline_to_advance(self, advance_pipeline_name, relative_to=None):
...
         """
-        def has_advanced(pipeline_instance):
+        def has_advanced(pipeline_instance, stage_name):
             """
             Check to see if a pipeline from a value stream map has been advanced.
             """
-            return pipeline_instance.stage('manual_verification').data['result'] != 'Unknown'
+            return pipeline_instance.stage(stage_name).data['scheduled']
 
         # Compute the last release time (in UTC) relative to the passed-in time -or- now.
```
cpennington: Update the comment to match the terminology used in the code? (last vs current vs previous)

doctoryes: Done.

tubular/gocd_api.py
doctoryes: Done.

tubular/gocd_api.py
doctoryes: Done.

cpennington (approved):

PR #148, opened 2017-03-06
Don't override the repo variable in merge_branch.py

cpennington:


cpennington:
I'm going to merge this w/o review to attempt to unblock the RC-cut pipeline.

PR #151, opened 2017-03-07
Only pass reference repo to git clone when it's available

cpennington:


cpennington:
Merging for great justice!

PR #152, opened 2017-03-07
Fix git tag fetching when tag is already created.

doctoryes:
Should fix this error:
https://gocd.tools.edx.org/go/tab/build/detail/edxapp_branch_cleanup/55/merge_rc_branch/1/tag_deployed_commit_job

@edx/pipeline-team @cpennington PTAL.


cpennington (approved):
Once tests are fixed...

macdiesel (approved):
File this whole object.sha type of thing as what I don't like about our current github api. Everything is one call too far down the tree.

PR #156, opened 2017-03-08
Don't create empty commits when there are no branches to octopus merge

cpennington:


macdiesel (approved):
I'm a little disappointed that you didn't call the branches to merge tentacles, but have my +1 anyway.

doctoryes (approved):
The edge cases!

PR #158, opened 2017-03-08
A differet approach to removing delete_on_ts from asgs

macdiesel:
This is the fix for TE-2019

### DIFF
--- a/tubular/ec2.py
+++ b/tubular/ec2.py
@@ -371,12 +371,14 @@ def remove_asg_deletion_tag(asg_name):
     Returns:
         None
     """
-    tag_to_delete = create_tag_for_asg_deletion(asg_name)
-    autoscale = boto.connect_autoscale()
-    if len(get_all_autoscale_groups([asg_name])) < 1:
+    asgs = get_all_autoscale_groups([asg_name])
+    if len(asgs) < 1:
         LOG.info("ASG {} no longer exists, will not remove deletion tag.".format(asg_name))
     else:
-        autoscale.delete_tags([tag_to_delete])
+        for asg in asgs:
+            for tag in asg.tags:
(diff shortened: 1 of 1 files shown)


feanil (approved):

tubular/ec2.py
```diff
@@ -371,12 +371,17 @@ def remove_asg_deletion_tag(asg_name):
...
+    if len(asgs) < 1:
         LOG.info("ASG {} no longer exists, will not remove deletion tag.".format(asg_name))
     else:
-        autoscale.delete_tags([tag_to_delete])
+        [
+            tag.delete()
+            for asg in asgs
+            for tag in asg.tags
+            if tag.key == ASG_DELETE_TAG_KEY
+         ]
```
cpennington: Here are a couple alternatives that might be less surprising than using the list comprehension for its side effects. I think I tend to prefer the straight-line loop.

```
tags = (
    tag
    for asg in asgs
    for tag in asg.tags
    if tag.key == ASG_DELETE_TAG_KEY
)

for tag in tags:
    tag.delet()
```

```
for asg in asgs
    for tag in asg.tags:
        if tag.key == ASG_DELETE_TAG_KEY:
            tag.delete()
```

```
tags = itertools.chain(asg.tags for asg in asgs)
tags = (tag for tag in tag if tag.key == ASG_DELETE_TAG_KEY)
for tag in tags:
    tag.delete()
```

macdiesel: I initially wrote the second choice and will go back to that, I would prefer not to add any additional iterations if not necessary.

cpennington: Actually, all of those snippets have the same number of iterations. In particular, they're all building generators, so in every case, they iterate over each tag once (and only once).

PR #160, opened 2017-03-15
Don't error out if the base_sha and head_sha are from different repos…

cpennington:
…. This allows messaging against both edx-platform and edx-platform-private

### DIFF
--- a/tubular/scripts/message_prs_in_range.py
+++ b/tubular/scripts/message_prs_in_range.py
@@ -102,9 +102,7 @@ def message_pull_requests(org,
         ami_tags = yaml.safe_load(base_ami_tags)
         tag = u'version:{}'.format(ami_tag_app)
         version = ami_tags[tag]
-        repo_url, _, base_sha = version.partition(u' ')
-        if not repo_url.endswith(u'github.com/{}/{}'.format(org, repo)):
-            raise ValueError(u"Reading base_sha from wrong repo, tag {!r} was {!r}".format(tag, version))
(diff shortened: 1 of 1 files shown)


cpennington:
I'm going to merge this aggressively, so that I can get a build to stage going.

tubular/scripts/message_prs_in_range.py
```diff
@@ -102,9 +102,7 @@ def message_pull_requests(org,
         ami_tags = yaml.safe_load(base_ami_tags)
         tag = u'version:{}'.format(ami_tag_app)
         version = ami_tags[tag]
-        repo_url, _, base_sha = version.partition(u' ')
-        if not repo_url.endswith(u'github.com/{}/{}'.format(org, repo)):
-            raise ValueError(u"Reading base_sha from wrong repo, tag {!r} was {!r}".format(tag, version))
+        _, _, base_sha = version.partition(u' ')
```
doctoryes, line 105: Fine - though I could see enforcing an `endswith('{org}/{repo})' or endswith('{org}/{repo}-private')` here.

doctoryes (approved):

macdiesel (approved):

macdiesel:
Frankenmerge.

tubular/scripts/message_prs_in_range.py
cpennington: My thinking was that if the base_sha doesn't exist in the repo, then it'll already throw an error lower down.

PR #162, opened 2017-03-20
remove decoding for python 3 strings

macdiesel:
TBH I'm not totally sure what to do here, It believe we no longer need to decode the string in python 3. Open to suggestions.


schenedx (approved):
👍 
Does this utility function gets used on other pipelines? I am hoping the scope of impact from this change is small

macdiesel:
The final place it's used is in the marketing pipeline, I had intended to depreciate this function on my first pass but this was still in use.

cpennington (approved):

PR #163, opened 2017-03-20
Allow specifying the variable for the version SHA when tagging

cpennington:


tubular/scripts/create_tag.py
```diff
@@ -68,6 +68,11 @@
     '--tag_message',
     help='Message to use when tagging.',
 )
+@click.option(
+    '--commit_sha_variable',
+    help='Name of the variable to read the SHA from from --input_file',
```
doctoryes: Nit: `from in --input_file`?

doctoryes (approved):

PR #164, opened 2017-03-20
Changes to enable CI checking from edxapp pipeline.

doctoryes:
https://openedx.atlassian.net/browse/TE-1738

Change the GitHub API calls to be consistent when checking test status.
Create directory for artifact when doesn't exist.

### DIFF
--- a/tubular/github_api.py
+++ b/tubular/github_api.py
@@ -285,7 +285,9 @@ class GitHubAPI(object):
             commit_sha (str): Commit SHA to check.
 
         Returns:
-            True if all tests have passed successfully, else False.
+            tuple(bool, dict):
+                bool: True if all tests have passed successfully, False otherwise
+                dict: Key/values of ci_context:ci_url
 
         Raises:
             github.GithubException.GithubException: If the response fails.
@@ -301,7 +303,9 @@ class GitHubAPI(object):
             pr_number (int): Number of PR to check.
 
         Returns:
-            True if all tests have passed successfully, else False.
+            tuple(bool, dict):
+                bool: True if all tests have passed successfully, False otherwise
+                dict: Key/values of ci_context:ci_url
 
         Raises:
             github.GithubException.GithubException: If the response fails.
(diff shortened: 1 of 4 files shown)


tubular/scripts/check_pr_tests_status.py
```diff
@@ -84,7 +93,7 @@ def check_tests(org,
         status_success = gh_utils.check_pull_request_test_status(pr_number)
         git_obj = 'PR #{}'.format(pr_number)
     elif commit_hash:
-        status_success = gh_utils.is_commit_successful(commit_hash)
```
feanil: Why add the extra function layer in front of the is_commit_successful function, it the wrapping function eventually going to do more?

doctoryes: Yeah - I did it for consistency in this method. But looking at it again - it's a needless extra level. I'll change it.

doctoryes: Wait - actually, I made wanted to keep the base `_is_commit_successful()` method "private" and expose the two `check_*_test_status` methods. Felt weird to have the method names so dissimilar. But - I'm open to changing it if you feel strongly about it.

doctoryes:
@edx/pipeline-team Now with output of each test status when checking the combined test status on GitHub. Please review.

tubular/github_api.py
```diff
@@ -246,6 +252,49 @@ def get_commit_combined_statuses(self, commit):
...
+        # Determine if the commit has passed all checks
+        if len(commit_status.statuses) < 1 or commit_status.state is None:
+            return (False, {})
+
+        return (
+            commit_status.state.lower() == 'success',
+            {cs.context: cs.target_url for cs in commit_status.statuses}
+        )
+
+    def check_commit_test_status(self, commit_sha):
```
macdiesel: could we change this to check_commit_combined_status?

macdiesel (approved):
one nit.

tubular/github_api.py
doctoryes: How about instead I change this to `check_combined_status_commit()` and change `check_pull_request_test_status()` to `check_combined_status_pull_request()`?

PR #166, opened 2017-03-21
Fix merge-approved-prs so that it correctly outputs repo and sha info…

cpennington:
…rmation


doctoryes (approved):
Seems obvious now! Perhaps we should start unit testing some of these scripts.

PR #169, opened 2017-03-22
Show changes across all named apps (even if they have the same repo) …

cpennington:
…in the release wiki page

You can see an example page here: https://openedx.atlassian.net/wiki/display/~cale/2017-03-23+Release

### DIFF
--- a/tubular/confluence_api.py
+++ b/tubular/confluence_api.py
@@ -17,7 +17,7 @@ from lxml.html.builder import E as element_maker
 import lxml.html
 
 SECTION = element_maker.section
-VersionDelta = namedtuple(u'VersionDelta', [u'base', u'new'])
+VersionDelta = namedtuple(u'VersionDelta', [u'app', u'base', u'new'])
 
 
 class Version(namedtuple(u'_Version', [u'repo', u'sha'])):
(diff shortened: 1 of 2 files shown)


tubular/confluence_api.py
```diff
@@ -152,9 +160,8 @@ def diff(base, new):
             They must also have the keys 'environment', 'deployment', 'play', and 'ami_id'.
     """
     diff_items = []
-    for delta in sorted(set(version_deltas(base, new)), key=lambda delta: delta.base or delta.new):
-        version = delta.new or delta.base
-        diff_items.append(E.LI(diff_link(version.repo, delta.base, delta.new)))
+    for delta in sorted(set(version_deltas(base, new)), key=lambda delta: (delta.app, delta.base or delta.new)):
```
feanil: NIT: this was hard to read.  Maybe a comment saying something like `# Add diff or commit links as a sorted list to the page.`

tubular/confluence_api.py
```diff
@@ -127,26 +130,25 @@ def diff_link(delta):
...
+        href = format_commit_url(delta.new)
     else:
         summary = u"{}...{}".format(delta.base.sha, delta.new.sha)
         href = u"{}/compare/{}...{}".format(
             delta.new.repo, delta.base.sha, delta.new.sha
         )
 
-    return E.A(
-        u"{} ({}): {}".format(
+    return (
```
doctoryes: Nit: Is this pair of parens still needed?

doctoryes (approved):
If it works, merge it.

tubular/confluence_api.py
macdiesel: I did have to go over this a few times myself.

cpennington: Yeah, it grew a bit incrementally. I'll tease it apart to make it more obvious.

tubular/confluence_api.py
cpennington: This is returning a tuple (I suppose it could return a list, as well... That would probably be clearer).

doctoryes: Right - I see that now. I suppose a list would be clearer.

PR #171, opened 2017-03-31
Cale/enhance pr messaging

cpennington:


macdiesel (approved):
If we are going to prefer - to _ could we make an entry in the readme file stating this preference so as new scripts are added the correct format is used?

PR #174, opened 2017-04-27
Scripts to merge private merged changes into public repo (and back).

doctoryes:
These two scripts will be used to:
- create a PR to merge any newly merged changes from the private repo over to the public repo
- push the merged public changes back over to the private repo to prepare for the next cycle

### DIFF
--- a/tubular/git_repo.py
+++ b/tubular/git_repo.py
@@ -11,6 +11,7 @@ import re
 from six.moves import urllib
 from git import Repo
 from git.util import rmtree
+from git.exc import GitCommandError
 
 
 LOGGER = logging.getLogger(__name__)
@@ -20,6 +21,13 @@ LOGGER.setLevel(logging.INFO)
 class InvalidGitRepoURL(Exception):
     """
     Raised when repo URL can't be parsed.
+    """
+    pass
+
+
+class FastForwardMergeImpossible(Exception):
+    """
+    Raised when attempting a fast-forward-only merge that can't be done.
     """
     pass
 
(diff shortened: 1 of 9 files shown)


tubular/git_repo.py
```diff
@@ -135,11 +176,12 @@ def force_branch_to(self, branch, commitish, remote=None):
             self.repo.heads[branch].reset(commitish, index=True)
 
     @contextmanager
-    def cleanup(self):
+    def cleanup(self, cleanup=True):
```
bmedx: I am confused about why you would use this context manager without cleanup=True. I don't see any usages in this PR where it is False?

doctoryes: I made this change as a mere convenience during debugging. I could just:
```
    with LocalGitAPI.clone(private_github_url, private_source_branch, reference_repo).cleanup(False) as local_repo:
```
to leave the cloned git repo around instead of changing it to this:
```
    local_repo = LocalGitAPI.clone(private_github_url, private_source_branch, reference_repo)
```
and changing all the other indentation beneath. But I don't wish to confuse - would you prefer I removed this change?

bmedx: If it's a change that you think will be useful going forward I'd suggest commenting it or perhaps renaming the "cleanup" param to indicate why you'd want to use it.

doctoryes: I'm going to remove it - I agree that it's confusing.

bmedx (approved):
I know very little about what's actually going on here, but nothing else stood out to me. 👍

PR #180, opened 2017-05-31
Improve rollback by AMI functionality and clarity

macdiesel:
Please take a look! This will fix the rollback by AMI functionality and make it more clear.

### DIFF
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -78,7 +78,6 @@ def clusters_for_asgs(asgs):
         u'test-edx-worker-v007',
     ]
 
-I want
     Returns:
         dict: A mapping of cluster names to asgs in the cluster.
 
@@ -519,8 +518,10 @@ def rollback(current_clustered_asgs, rollback_to_clustered_asgs, ami_id=None):
 
     Returns:
         dict(str, str, dict): Returns a dictionary with the keys:
-            'ami_id' - AMI id used to deploy the AMI, None if unspecified
+            'ami_id' - AMI id used to deploy the AMI, None if unspecified.
+            'current_ami_id' -  The AMI that is running in an environment after a rollback.
             'current_asgs' - Lists of current active ASGs, keyed by cluster.
+            'disabled_ami_id' - The AMI that was running in an environment before the rollback.
             'disabled_asgs' - Lists of current inactive ASGs, keyed by cluster.
 
     Raises:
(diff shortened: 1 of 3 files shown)


tubular/asgard.py
```diff
@@ -75,7 +78,7 @@ def clusters_for_asgs(asgs):
         u'test-edx-worker-v007',
     ]
 
-
+I want
```
feanil: Tell me what you want, what you really really want

feanil (approved):

tubular/asgard.py
macdiesel: 🌶 :woman_singer: 🎸

tubular/asgard.py
```diff
@@ -544,6 +548,16 @@ def rollback(current_clustered_asgs, rollback_to_clustered_asgs, ami_id=None):
                 LOG.info("Rollback ASG '{}' has been removed. Aborting rollback to ASGs.".format(asg))
                 rollback_ready = False
                 break
+    disabled_ami_id = None
+    try:
+        # fetch the currently deployed AMI_ID for the clusters for logging
+        if ami_id:
+            edp = ec2.edp_for_ami(ami_id)
+            disabled_ami_id = ec2.active_ami_for_edp(edp.environment, edp.deployment, edp.play)
+    except(MissingTagException, ImageNotFoundException, MultipleImagesFoundException):
```
doctoryes: Nit: Space after `except`.

tubular/asgard.py
```diff
@@ -552,15 +566,27 @@ def rollback(current_clustered_asgs, rollback_to_clustered_asgs, ami_id=None):
...
         return deploy(ami_id)
     else:
         LOG.info("No AMI id specified - so no deploy occurred during rollback.")
-        return {'ami_id': None, 'current_asgs': current_clustered_asgs, 'disabled_asgs': rollback_to_clustered_asgs}
+        return {
+            'ami_id': None,
+            'current_ami_id': ami_id,
+            'current_asgs': current_clustered_asgs,
+            'disabled_ami_id': disabled_ami_id,
+            'disabled_asgs': rollback_to_clustered_asgs
```
doctoryes, line 588: Ugh - feel like I didn't get the names quite right initially.

tubular/asgard.py
```diff
@@ -586,6 +614,9 @@ def deploy(ami_id):
     # Pull the EDP from the AMI ID
     edp = ec2.edp_for_ami(ami_id)
 
+    # fetch the active AMI_ID
+    disabled_ami_id = ec2.active_ami_for_edp(edp.environment, edp.deployment, edp.play)
```
doctoryes: Why didn't you add the try/except around this new call as well? Did you want the deploy to actually fail if this call failed?

doctoryes: I think all pipeline systems already have made this call back upstream to determine the base AMI on which a new AMI should be built. I wonder if that value could be used *or* maybe not used, but sanity-checked against.

doctoryes (commented):
Looks good - just nits and questions.

tubular/asgard.py
macdiesel: I think this is a good idea but out of scope for this fix.

macdiesel: Good point, I've added a try block around this.

macdiesel:
Comments addressed.

doctoryes (approved):

PR #182, opened 2017-07-13
[WIP] add configure_dynamic_scaling.py script

pwnage101:
WIP.

Add a `configure_dynamic_scaling.py` script which deploys cloudwatch alarms and scaling policies causing the edxapp ASG to dynamically scale.  This also introduces boto3 and makes `giveup_if_not_throttling()` compatible with both boto and boto3.

This script will be utilized by a change to the edxapp pipeline: https://github.com/edx/edx-gomatic/pull/345

### DIFF
--- a/tubular/ec2.py
+++ b/tubular/ec2.py
@@ -7,12 +7,15 @@ from __future__ import unicode_literals
 
 import os
 import logging
+import re
 import time
 from datetime import datetime, timedelta
 import backoff
 import boto
 from boto.exception import EC2ResponseError, BotoServerError
 from boto.ec2.autoscale.tag import Tag
+import boto3
+import botocore.exceptions  # use botocore.exceptions.ClientError instead of BotoServerError for boto3 exceptions
 from tubular.utils import EDP, WAIT_SLEEP_TIME
 from tubular.exception import (
     ImageNotFoundException,
@@ -29,17 +32,33 @@ MAX_ATTEMPTS = os.environ.get('RETRY_MAX_ATTEMPTS', 5)
 RETRY_FACTOR = os.environ.get('RETRY_FACTOR', 1.5)
 
 
+# TODO: determine if we need to handle the situation where max retry attempts was reached.
 def giveup_if_not_throttling(ex):
     """
-    Checks that a BotoServerError exceptions message contains the throttling string.
+    Checks if a BotoServerError/ClientError exception indicates non-throttling related error.
+
+    If this check returns True, it suggests that retry handlers should give up.
 
     Args:
-        ex (boto.exception.BotoServerError):
+        ex (boto.exception.BotoServerError or botocore.exceptions.ClientError):
 
     Returns:
-        False if the throttling string is not found.
+        boolean: True if the exception indicates non-throttling related error.
+
+    Raises:
+        ValueError: the exception parameter is not a boto error response.
     """
-    return not (str(ex.status) == "400" and ex.body and '<Code>Throttling</Code>' in ex.body)
+    if isinstance(ex, BotoServerError):
+        status_code = ex.status
+        message = ex.body
+    elif isinstance(ex, botocore.exceptions.ClientError):
+        status_code = ex.response['Error'].get('Code', 'Unknown')
+        message = ex.response['Error'].get('Message', 'Unknown')
+    else:
+        raise ValueError('Expecting the parameter to be either boto.exception.BotoServerError (boto2) or ' +
+                         'botocore.exceptions.ClientError (boto3), instead got {}'.format(type(ex)))
+
+    return not (str(status_code) == '400' and message and '<Code>Throttling</Code>' in message)
 
 
 @backoff.on_exception(backoff.expo,
@@ -262,16 +281,18 @@ def is_stage_ami(ami_id):
     return ami_for_stage
 
 
-def asgs_for_edp(edp, filter_asgs_pending_delete=True):
+def asgs_for_edp(edp, filter_asgs_pending_delete=True, regex_filter=None):
     """
     All AutoScalingGroups that have the tags of this play.
 
     A play is made up of many auto_scaling groups.
 
     Arguments:
-        EDP Named Tuple: The edp tags for the ASGs you want.
+        edp (EDP Named Tuple): The edp tags for the ASGs you want.
+        filter_asgs_pending_delete (bool): Do not include ASGs tagged for deletion.
+        regex_filter (str): Only include ASGs matching this regex.
     Returns:
-        list: list of ASG names that match the EDP.
+        list: list of ASG names that match the EDP and given filters.
     eg.
 
      [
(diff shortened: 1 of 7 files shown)


michaelroytman:
@edx/pipeline-team Could you review this PR to make sure this will accomplish adding a "configure_dynamic_scaling" stage to the edxapp pipeline? The relevant ticket is here: https://openedx.atlassian.net/browse/DEVOPS-6177.

tubular/ec2.py
```diff
@@ -560,3 +578,172 @@ def _get_elb_health(selected_elb):
...
+
+
+def edxapp_asg_for_environment(edx_environment):
+    """
+    Get the edxapp ASG name for the given environment.
+
+    The ASG name returned should represent an edxapp ASG managed/created by
+    Asgard.
+    """
+    edp = EDP(edx_environment, 'edx', 'edxapp')
```
macdiesel: asgs_for_edp performs this functionality. This function seems very specific to the use case. I would prefer we keep things like the play names and deployments out of this file.

tubular/ec2.py
```diff
@@ -560,3 +578,172 @@ def _get_elb_health(selected_elb):
...
+
+@backoff.on_exception(backoff.expo,
+                      botocore.exceptions.ClientError,
+                      max_tries=MAX_ATTEMPTS,
+                      giveup=giveup_if_not_throttling,
+                      factor=RETRY_FACTOR)
+def ensure_edxapp_scaling_alarm(asg, alarm_name_suffix, comparison_operator, threshold, scaling_policy_arn):
+    """
+    Arguments:
+        alarm_name_suffix (str): Name suffix for the alarm.
```
macdiesel: Please round out the docs here.

tubular/ec2.py
```diff
@@ -560,3 +578,172 @@ def _get_elb_health(selected_elb):
...
+
+    return edxapp_asgs_only[0]
+
+
+@backoff.on_exception(backoff.expo,
+                      botocore.exceptions.ClientError,
+                      max_tries=MAX_ATTEMPTS,
+                      giveup=giveup_if_not_throttling,
+                      factor=RETRY_FACTOR)
+def ensure_edxapp_scaling_alarm(asg, alarm_name_suffix, comparison_operator, threshold, scaling_policy_arn):
```
macdiesel: It would be better if this was a function to ensure_scaling_alarm and you passed in the alarm metric kwargs and passed them through to put_metric_alarm or provided more options for the alarm itself. 

Functions in this file shouldn't be specific to any particular E, D, or P.

tubular/ec2.py
```diff
@@ -560,3 +578,172 @@ def _get_elb_health(selected_elb):
...
+        ComparisonOperator=comparison_operator,
+    )
+
+
+@backoff.on_exception(backoff.expo,
+                      botocore.exceptions.ClientError,
+                      max_tries=MAX_ATTEMPTS,
+                      giveup=giveup_if_not_throttling,
+                      factor=RETRY_FACTOR)
+def ensure_edxapp_scaling_policy(asg, policy_name_suffix, scaling_adjustment_type, scaling_adjustment, cooldown=600):
```
macdiesel: Functions in this file shouldn't be specific to any particular E, D, or P.

tubular/ec2.py
```diff
@@ -560,3 +578,172 @@ def _get_elb_health(selected_elb):
...
+    This does immediately enable dynamic scaling, which means that if min and
+    max ASG capacity are non-equal, instances may suddenly launch or terminate.
+
+    Arguments:
+        asg_name (string): Name of an edxapp ASG.
+    """
+    upper_utilization_threshold = 60
+    # lower_utilization_threshold must be half of upper_utilization_threshold
+    # so that scaling from 2 to 1 instance does not cause thrashing.
+    lower_utilization_threshold = 30
```
macdiesel: These thresholds should be passed in to the function, this is too specific for this module.

tubular/ec2.py
```diff
@@ -560,3 +578,172 @@ def _get_elb_health(selected_elb):
...
+        PolicyName='{}-{}'.format(asg, policy_name_suffix),
+        PolicyType='SimpleScaling',
+        AdjustmentType=scaling_adjustment_type,
+        ScalingAdjustment=scaling_adjustment,
+        Cooldown=cooldown,
+    )
+    return response
+
+
+def configure_dynamic_scaling_for_edxapp_asg(asg_name):
```
macdiesel: This might be the function you want to pull out in to the scripts directory.

macdiesel (commented):
Looks like a good start, I left a few suggestions.

tubular/ec2.py
pwnage101: I'll move this to the script.

Something I noticed trying to use asgs_for_edp directly is that it returned a list when I only wanted one element from the list:

```
>>> tubular.ec2.asgs_for_edp(EDP('loadtest', 'edx', 'edxapp'))
['loadtest-edx-edxapp-v238', 'loadtest-edx-worker-v204']
```

Is there another way to get only `'loadtest-edx-edxapp-v238'` or should I just keep filtering the list?

macdiesel: There are 2 ASGs for that EDP, one is the worker the other is the web app. One could argue that these should be different play names or there should be another bit in that tuple, but that ship has sailed long ago.

If you wanted to add a filter to the adp_for_asg function that you pass in I think that would be fine. It could match on a regex for the items that you want to match. Just make sure it doesn't break existing tests/functionality.

pwnage101: went ahead and added a regex filter parameter for that function, and added a test.

pwnage101:
It's ready for another review pass.  I moved most edxapp-specific things into the script, and it now also does some minimal validation to make sure there aren't any extraneous policies or alarms (still, it does not error or exit if there are extraneous resources, just a warning).

tubular/ec2.py
```diff
@@ -290,11 +311,18 @@ def asgs_for_edp(edp, filter_asgs_pending_delete=True):
...
         if filter_asgs_pending_delete and ASG_DELETE_TAG_KEY in tags.keys():
-            LOG.info("filtering ASG: {0} because it is tagged for deletion on: {1}"
+            LOG.info('filtering out ASG "{0}" because it is tagged for deletion on "{1}"'
                      .format(group.name, tags[ASG_DELETE_TAG_KEY]))
+            filter_group = True
+        if regex_filter is not None and not re.match(regex_filter, group.name):
+            LOG.info('filtering out ASG "{0}" because it does not match the regex filter "{1}"'
+                     .format(group.name, regex_filter))
+            filter_group = True
+        if filter_group:
```
macdiesel: Could you explain filter_group to me?

pwnage101: It gets set to True when the current group in the iteration has been filtered-out for any reason.  Deferring the `continue` statement until after all filter checks allows all filter-related logging statements to get logged.

I also updated the code to clarify.

macdiesel (approved):

pwnage101:
:pizza:!

I won't merge at least until after testing and updating the commit message.

pwnage101:
I'm closing this now because manipulation of ASGs is currently somewhat incompatible with Asgard.  I can't remember the exact reason, something to do with Asgard renaming scaling policies or scaling alarms to its liking, and the names being unpredictable from tubular's perspective.

I'll leave this branch lingering so we can use it as a starting point or guideline for a post-asgard world (see PERF-436).

PR #184, opened 2017-08-10
[WIP] SplitMongo Pruning

eltoncarr-ms:
> ### This is work in progress

This PR introduces initial code for handling splitmongo pruning. 

**Core capabilities currently available**:
1. querying specified edxapp database active_versions and structures collections 
2. identifying all structures that can be removed
3. removing identified structures

All capabilities above are supported for both 
1. live (ie: actual database interaction) 
2. static (using test data)

**Although functional, do not run this against a live database.**

Additional capabilities that are in progress (or targeted):
1. addressing performance concerns about pulling all documents in active_versions and structures collections at once
2. enabling visualization of sample course/library structures before and after pruning

**Running against sample data**:
`python tubular/scripts/prune_modulestore.py --relink-structures True --test-data-file tubular/tests/test_prune_modulestore_data.json --output-file ~/pruned_output.json`

**Known Issues**:
1. ~~this script **doesn't** work against the **Ginkgo" release~~

### DIFF
--- a/tubular/modulestore.py
+++ b/tubular/modulestore.py
@@ -124,14 +124,7 @@ class ModuleStore(object):
 
         """
 
-        pruned_structures = []
-
-        for structure_doc in structures:
-
-            if structure_doc['_id'] not in structures_to_remove:
-                pruned_structures.append(structure_doc)
-
-        return pruned_structures
+        return [structure_doc for structure_doc in structures if structure_doc['_id'] not in structures_to_remove]
 
     def load_test_dataset(self, dataset_file):
 
@@ -148,7 +141,6 @@ class ModuleStore(object):
         # check if the specified file exists
         file_exists = os.path.isfile(dataset_file)
 
-        assert isinstance(file_exists, object)
         if not file_exists:
             raise IOError("The specified file doesn't exist: %s", dataset_file)
 
@@ -187,10 +179,8 @@ class ModuleStore(object):
 
         """
 
-        av_filter = {'$in': []}
-
-        for active_version_id in active_version_id_list.split(","):
-            av_filter['$in'].append(ObjectId(active_version_id.strip()))
+        filter_list = [ObjectId(active_version_id.strip()) for active_version_id in active_version_id_list.split(",")]
+        av_filter = {'$in': filter_list}
 
         # establish the query filter (respecting cases where no value is specified)
         return self.get_query_filter(av_filter)
@@ -211,15 +201,8 @@ class ModuleStore(object):
 
             # check for unknown versions
             unknown_versions = [
-                version
-                for
-                version
-                in
-                active_version['versions']
-                if
-                version
-                not in
-                self.target_active_versions_keys
+                version for version in active_version['versions']
+                if version not in self.target_active_versions_keys
             ]
 
             if unknown_versions:
@@ -455,9 +438,7 @@ class ModuleStore(object):
             self.log(message_template % (version_retention, self._minimum_version_retention), "info")
 
         # iterate the active versions and build the associated version tree
-        for active_version in active_versions:
-
-            counter += 1
+        for counter, active_version in enumerate(active_versions, start=1):
 
             for target_key in self.target_active_versions_keys:
 
--- a/tubular/modulestore.py
+++ b/tubular/modulestore.py
@@ -11,11 +11,28 @@ from pymongo import MongoClient
 
 class ModuleStore(object):
     """
-    Handles pruning operations for the edx module store structues
+    Handles pruning operations for the edx module store structures
+
+    Here are schemas for key objects referenced in this class:
+
+    active_version: dictionary representing an active_version document
+        ** '_id': an ObjectId (guid)
+        ** 'versions': dictionary of versions
+            ** 'draft-branch': draft branch version
+            ** 'published-branch': published branch version
+            ** 'library': library version
+
+    structure: dictionary representing a structure document:
+        ** '_id': an ObjectId (guid)
+        ** 'original_version': the original structure id in the previous_version relation
+        ** 'previous_version': the structure id from which this one was derived. For published courses, this
+                                points to the previously published version of the structure not the draft
+                                published to this.
+
     """
 
     # the dictionary keys to track for active versions
-    target_active_versions_keys = [u'library', u'draft-branch', u'published-branch']
+    target_active_versions_keys = ['library', 'draft-branch', 'published-branch']
 
     def __init__(self, logger=None):
 
@@ -23,13 +40,20 @@ class ModuleStore(object):
         self.logger = logger
         self.db = None
 
+        # retention minimum
+        self._minimum_version_retention = 3
+
     def initialize_database_connection(self, mongo_database_connection=None, mongo_database_name="edxapp"):
 
         """
         Initialize database connection
+
+        :param mongo_database_connection: connection string to mongo database. If none is specified, localhost is used.
+        :param mongo_database_name: name of the mongo database the application will interact with
+
         """
 
-        if mongo_database_connection is None:
+        if not mongo_database_connection:
             client = MongoClient()
         else:
             client = MongoClient(mongo_database_connection)
@@ -40,47 +64,85 @@ class ModuleStore(object):
 
         """
         Log a message
+
+        :param message: the message to log
+        :param message_type: the type of log message (info, warning or debug). If a message is not info or warning, it
+        falls back to debug
+
         """
 
-        if self.logger is not None:
+        if self.logger:
             if message_type == "info":
                 self.logger.info(message)
+
+            elif message_type == "warning":
+                self.logger.warning(message)
+
             else:
                 self.logger.debug(message)
 
-    def save_data_file(self, data, output_file):
+    def save_data_file(self, data, output_file=None, output_file_object=None):
 
         """
         Save the specified data file to disk
+
+        :param data: the active versions and/or structure data being saved
+        :param output_file: path to an output file to save the data
+        :param output_file_object: tempfile file-like object for saving temporary files
+
+        This function supports two options:
+        1. specifying the output file name - output will be saved at the path specified
+        2. a file-like object returned from tempfile.NamedTemporaryFile
+
+        Either output_file or output_file_object must be specified. If both are specified,
+        output_file will be preferred. If none is specified, an exception will be raised.
+
         """
 
-        self.log("Saving the purged dataset to {0}".format(output_file))
+        if output_file:
+            self.log("Saving the purged dataset to %s" % output_file)
 
-        # write the updated dataset
-        with open(output_file, 'w') as outfile:
-            json.dump(data, outfile)
+            # write the updated dataset
+            with open(output_file, 'w') as outfile:
+                json.dump(data, outfile)
 
-    def prune_structures_static_data(self, original_dataset, structures_to_remove):
+        elif output_file_object:
+            self.log("Saving the purged dataset using tempfile file-like object")
+            output_file_object.write(json.dumps(data).encode())
+
+        else:
+            raise ValueError("you must specify either the output_file or output_file_object")
+
+    def prune_structures_static_data(self, structures, structures_to_remove):
 
         """
         Prune the static test data and return the results
+
+        :param structures: list of structure dictionaries (see dictionary definition in class docstring)
+        :param structures_to_remove: list of ObjectIds representing Ids of structures that should be pruned
+        :return: list of structure dictionaries
+
         """
 
-        pruned_static_data = []
+        pruned_structures = []
 
-        for structure_doc in original_dataset[u'structures']:
+        for structure_doc in structures:
 
-            if structure_doc[u'_id'] not in structures_to_remove:
-                pruned_static_data.append(structure_doc)
+            if structure_doc['_id'] not in structures_to_remove:
+                pruned_structures.append(structure_doc)
 
-        original_dataset[u'structures'] = pruned_static_data
-
-        return original_dataset
+        return pruned_structures
 
     def load_test_dataset(self, dataset_file):
 
         """
-        Load the json dataset from the file specified
+        Load the json data set from the file specified
+
+        :param dataset_file: file containing the test data set
+            ** 'active_versions': list of active version dictionaries
+            ** 'structures': list of structure dictionaries
+        :return: dictionary representing the test data set
+
         """
 
         # check if the specified file exists
@@ -88,7 +150,7 @@ class ModuleStore(object):
 
         assert isinstance(file_exists, object)
         if not file_exists:
-            raise IOError("The specified file doesn't exist: {0}".format(dataset_file))
+            raise IOError("The specified file doesn't exist: %s", dataset_file)
 
         # load the file
         with open(dataset_file) as dataset:
@@ -100,10 +162,15 @@ class ModuleStore(object):
 
         """
         Generate a document filter for bulk querying
+
+        :param doc_filter: dictionary representing a mongo bulk query filter
+            ** '$in': list of document ids guids
+        :return: fully formed mongo query filter
+
         """
 
         # establish the query filter  (respecting cases where no value is specified)
-        query_filter = None
+        query_filter = {}
 
         if len(doc_filter['$in']) > 0:
             query_filter = {"_id": doc_filter}
@@ -114,6 +181,10 @@ class ModuleStore(object):
 
         """
         Generate document filter for bulk querying the active version collection
+
+        :param active_version_id_list: list of ids for active versions
+        :return: fully formed mongo query filter
+
         """
 
         av_filter = {'$in': []}
@@ -121,18 +192,39 @@ class ModuleStore(object):
         for active_version_id in active_version_id_list.split(","):
             av_filter['$in'].append(ObjectId(active_version_id.strip()))
 
-        # establish the query filter  (respecting cases where no value is specified)
+        # establish the query filter (respecting cases where no value is specified)
         return self.get_query_filter(av_filter)
 
-    def get_structures_filter(self, active_version_list):
+    def get_structures_filter(self, active_version_list=None):
 
         """
         Generate document filter for bulk querying the structures collection
+
+        :param active_version_list: list of active version dictionaries (see dictionary definition in class docstring)
+        :return: fully formed mongo query filter
+
         """
 
         structure_filter = {'$in': []}
 
         for active_version in active_version_list:
+
+            # check for unknown versions
+            unknown_versions = [
+                version
+                for
+                version
+                in
+                active_version['versions']
+                if
+                version
+                not in
+                self.target_active_versions_keys
+            ]
+
+            if unknown_versions:
+                message = "%s are not currently tracked for pruning" % unknown_versions
+                self.log(message, "warning")
 
             for target_key in self.target_active_versions_keys:
                 if target_key in active_version['versions']:
@@ -145,12 +237,15 @@ class ModuleStore(object):
 
         """
         Get all documents from the active_versions collection
+        :param active_version_list: list of active version dictionaries (see dictionary definition in class docstring)
+        :return: list of active version dictionaries
+
         """
 
         # establish the active version filter (if required)
         active_version_filter = None
 
-        if active_version_list is not None:
+        if active_version_list:
             active_version_filter = self.get_active_version_filter(active_version_list)
 
         fields = {
@@ -173,24 +268,28 @@ class ModuleStore(object):
             # collect all interesting docs: library & [draft|published]-branch active versions
             avdocs_versions = active_version_doc['versions']
 
-            if u'library' in avdocs_versions \
-                    or u'draft-branch' in avdocs_versions \
-                    or u'published-branch' in avdocs_versions:
+            if 'library' in avdocs_versions \
+                    or 'draft-branch' in avdocs_versions \
+                    or 'published-branch' in avdocs_versions:
                 active_versions.append(active_version_doc)
 
         # return the active versions
         return active_versions
 
-    def get_structures(self, filter_enabled, active_versions_list):
+    def get_structures(self, active_versions=None):
 
         """
         Get all documents from the structures collection
+
+        :param active_versions: list of active version dictionaries (see dictionary definition in class docstring)
+        :return: list of structure dictionaries (see dictionary definition in class docstring)
         """
 
         # use filters (if required)
         structure_filter = None
-        if filter_enabled:
-            structure_filter = self.get_structures_filter(active_versions_list)
+        if active_versions:
+            # TODO: implement structure filtering based on specified active versions
+            structure_filter = self.get_structures_filter()
 
         fields = {
             "_id": 1,
@@ -198,58 +297,53 @@ class ModuleStore(object):
             "original_version": 1
         }
 
-        structures_list = []
+        # Get a list of all structures (or those relevant to the active versions via specified filter)
+        structure_docs = self.db.modulestore.structures.find(structure_filter, fields)
 
-        # TODO: apply filters to support limiting the pruning operation to 1 or more courses / active versions
-        if structure_filter is None:
-            resultset = self.db.modulestore.structures.find({}, fields)
-        else:
-            resultset = self.db.modulestore.structures.find(structure_filter, fields)
-
-        for structure_doc in resultset:
-            # Get a list of all structures (or those relevant to the active versions via specified filter)
-            # this will give the list of Dictionary's with _id and previous_version
-            structures_list.append(structure_doc)
-
-        return structures_list
+        return list(structure_docs)
 
     def prune_structures(self, structures_to_remove):
 
         """
         Prune the specified documents from the structures collection
+
+        :param structures_to_remove: list of structure ids targeted for removal
+        :return: mongo WriteResult object representing the status of the operation
+
         """
 
         # TODO: (performance): chunk up the removals to minimize possible production impact
-        if len(structures_to_remove) == 0:
+        if not structures_to_remove == 0:
             return
 
-        structures_removal_filter = {'$in': []}
+        structures_removal_filter = {
+            '$in': [ObjectId(oid) for oid in structures_to_remove]
+        }
 
-        for structure_objectid_str in structures_to_remove:
-            structures_removal_filter['$in'].append(ObjectId(structure_objectid_str))
-
-        return self.db.modulestore.structures.remove({'_id': structures_removal_filter})
+        return self.db.modulestore.structures.remove(structures_removal_filter)
 
     def relink(self, structures):
 
         """
         Relink structures to their original version post pruning
+
+        :param structures: list of structure dictionaries to relink (see dictionary definition in class docstring)
+        :return: list of structure dictionaries relinked to their original versions
+
         """
 
         self.log("Relinking structures to their original versions")
 
-        index_position = 0
-        available_ids = []
-
         # build a list of all available ids
-        available_ids.extend([structure_doc['_id'] for structure_doc in structures])
+        available_ids = {structure_doc['_id'] for structure_doc in structures}
 
         # iterate structures and relink to original
-        for structure_doc in structures:
+        for index_position, structure_doc in enumerate(structures):
 
-            if structure_doc["previous_version"] is not None and structure_doc["previous_version"] not in available_ids:
+            if structure_doc["previous_version"] and structure_doc["previous_version"] not in available_ids:
 
-                self.log("{0} was not found in {1}".format(structure_doc["previous_version"], available_ids), "debug")
+                message = "%s was not found in %s" % (structure_doc["previous_version"], available_ids)
+                self.log(message, "debug")
 
                 to_be_linked_version_id = []
                 original_version_id = []
@@ -257,11 +351,11 @@ class ModuleStore(object):
                 to_be_linked_version_id.append(structure_doc['_id'])
                 original_version_id.append(structure_doc['original_version'])
 
-                self.log("{0} version is being linked to {1}".format(to_be_linked_version_id, original_version_id[0]),
-                         "debug")
+                message = "%s version is being linked to %s" % (to_be_linked_version_id, original_version_id[0])
+                self.log(message, "debug")
 
                 # TODO: refactor to support bulk updates for performance concerns
-                if self.db is not None:
+                if self.db:
 
                     # this is a live update session
                     self.db.modulestore.structures.update(
@@ -271,74 +365,94 @@ class ModuleStore(object):
                 else:
 
                     # this is working against static dataset
-                    structures[index_position][u'previous_version'] = original_version_id[0]
+                    structures[index_position]['previous_version'] = original_version_id[0]
 
             else:
-                self.log("Nothing to link for structure: {0}".format(structure_doc['_id']), "debug")
+                self.log("Nothing to link for structure: %s" % structure_doc['_id'], "debug")
 
             # advance the index position
             index_position += 1
 
         return structures
 
-    def find_previous_version(self, lookup_key, lookup_value, structures_list):
+    def find_previous_version(self, lookup_key, lookup_value, structures):
 
         """
         This function searches all structure documents for the one specified
+
+        :param lookup_key: dictionary key to use for lookups
+        :param lookup_value: expected value of the dictionary field to establish a match
+        :param structures: list of structure dictionaries (see dictionary definition in class docstring)
+        :return: structure dictionary with matching key/value or None
+
         """
 
+        # TODO: Perf improvements needed here (ie: using lookup dictionary instead)
         # it is more efficient to use a structures list
         # instead of separate db calls (which may be necessary for supporting
         # pruning individual active versions)
-        for structure_doc in structures_list:
+        for structure_doc in structures:
 
             if structure_doc[lookup_key] == lookup_value:
                 return structure_doc
 
-    def build_activeversion_tree(self, active_version, structures):
+    def build_active_version_ancestry(self, active_version, structures):
 
         """
-        Build a tree representing the active_version and its tree of ancestors
-        from structures
+        Build a tree representing the active_version and its list of ancestors from structures
+        :param active_version: an active version dictionary (see dictionary definition in class docstring)
+        :param structures: list of structure dictionaries (see dictionary definition in class docstring)
+        :return: list of structures representing an active version lineage
         """
 
         # link the active version to its base version in structures
         structure_doc = self.find_previous_version('_id', active_version, structures)
 
-        # map the tree for the identified structure
-        version_tree = []
+        # map the ancestors for the identified structure
+        version_ancestry = []
 
         # recursively identify the ancestors
-        if structure_doc is not None:
+        if structure_doc:
 
-            # build the tree
-            version_tree.append(str(structure_doc['_id']))
+            # build the ancestral list
+            version_ancestry.append(str(structure_doc['_id']))
 
-            while structure_doc[u'previous_version'] is not None:
+            while structure_doc['previous_version']:
 
                 # search for the parent structure doc
-                structure_doc = self.find_previous_version('_id', structure_doc[u'previous_version'], structures)
+                structure_doc = self.find_previous_version('_id', structure_doc['previous_version'], structures)
 
-                # build the tree - recursively
-                if structure_doc is None:
-                    # end of the tree (original version)
+                # build the ancestral list - recursively
+                if not structure_doc:
+                    # start of the lineage (original version)
                     break
 
-                version_tree.append(str(structure_doc['_id']))
+                version_ancestry.append(str(structure_doc['_id']))
 
-        return version_tree
+        return version_ancestry
 
-    def get_structures_to_delete(self, active_versions, structures=None, version_retention=2, relink_structures=False):
+    def get_structures_to_delete(self, active_versions, structures=None, version_retention=3):
 
         """
         Generate a list of structures that meet the conditions for pruning and associated visualization
+
+        :param active_versions: list of active version dictionaries (see dictionary definition in class docstring)
+        :param structures: list of structure dictionaries (see dictionary definition in class docstring)
+        :param version_retention: number of structures to retain after purge excluding the original version
+        :return: set of structures to be removed and their ancestry list
         """
 
         # initialize key variables
-        version_trees = []
+        version_ancestry_lists = []
         versions_to_retain = []
         versions_to_remove = []
         counter = 0
+
+        # defensive: ensure that version_retention >=3
+        if version_retention < self._minimum_version_retention:
+
+            message_template = "Version retention of %s is below the minimum allowed and is being updated to %s"
+            self.log(message_template % (version_retention, self._minimum_version_retention), "info")
 
         # iterate the active versions and build the associated version tree
         for active_version in active_versions:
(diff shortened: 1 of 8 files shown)


ormsbee:
I realize this is a WIP, but I'm curious -- what in particular makes this incompatible with Ginkgo?

eltoncarr-ms:
@ormsbee , regarding Ginkgo, we had done an initial run about 1 month ago, saw the failure but did not explore it further. We instead opted for getting the script fully operational for Ficus. We re-validated this morning and had no issues. I will remove the limitation.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+)
+@click.option(
+    u'--active-version-filter',
+    default=None,
+    help=u'comma-separated list of objectIds to target for pruning'
+)
+@click.option(
+    u'--database-name',
+    default=u'edxapp',
+    help=u'name of the edx database to prune'
```
doctoryes: `Name of the Open edX MongoDB database to prune`? The current help is vague without needed context.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+@click.option(
+    u'--version-retention',
+    default=2,
+    help=u'Number of versions to retain for a course/library'
+)
+@click.option(
+    u'--relink-structures',
+    default=False,
+    help=u'boolean indicator of whether or not to relink the structures to the original version after pruning'
+)
```
doctoryes: Is this option needed? I think we'd always want to re-link the structures.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+@click.option(
+    u'--database-name',
+    default=u'edxapp',
+    help=u'name of the edx database to prune'
+)
+@click.option(
+    u'--test-data-file',
+    default=None,
+    help=u'file path containing a json representation of test data to use for pruning validation'
+)
```
doctoryes: A better test strategy would be to move the util code that performs the pruning out of the actual click script. Then test that separated util code separately with test data. It'd keep the script below having a single purpose. Many of the other scripts in this repo do the same - the script itself handles validating parameters and then calls the specific utility code in other modules. This strategy also makes the utility code callable from other useful locations.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+@click.option(
+    u'--output-file',
+    default=u'pruned_dataset.json',
+    help=u'output file of the prune structures for test purposes'
+)
```
doctoryes: Ditto.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    for active_version_doc in resultset:
+
+        # collect all interesting docs: library & [draft|published]-branch active versions
+        avdocs_versions = active_version_doc['versions']
+
+        if u'library' in avdocs_versions \
+                or u'draft-branch' in avdocs_versions \
+                or u'published-branch' in avdocs_versions:
+            active_versions.append(active_version_doc)
```
doctoryes: Why this looping check? Is it possible that none of these `fields` will be present?

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    return db.modulestore.structures.remove({'_id': structures_removal_filter})
+
+
+# TODO: get this fully operational. Test against static data is ready.
+def relink(db, structures):
+
+    """
+    There are ongoing discussions about the need to support relinking modulestore structures
+    to their original version.
+
```
doctoryes: Ah - I see. Unless I hear a convincing case from someone, I vote for re-linking.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    else:
+        resultset = db.modulestore.structures.find(structure_filter, fields)
+
+    for structure_doc in resultset:
+
+        # Get a list of all structures (or those relevant to the active versions via specified filter)
+        # this will give the list of Dictionary's with _id and previous_version
+        structures_list.append(structure_doc)
+
+    return structures_list
```
doctoryes: `return list(resultset)`?

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+
+def prune_structures(db, structures_to_remove):
+
+    """
+    Prune the specified documents from the structures collection
+    """
+
+    # TODO: (performance): chunk up the removals to minimize possible production impact
```
doctoryes: I agree. The script should ultimately be tested on a large course with many edits/structures in its history.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+                # print(active_version)
+                version_tree = build_activeversion_tree(active_version['versions'][target_key], structures)
+
+                # only add the tree if it has 1+ element
+                tree_length = len(version_tree)
+                if tree_length > 0:
+
+                    status_message = "Processing Active Version {0} | {3}: {1} version with a {2}-version tree"
+
+                    LOG.debug(status_message.format(
```
doctoryes: Nit: Most of your log.debugs should be log.info - they're important! 😄

doctoryes (commented):
@eltoncarr Thanks for submitting this code! It's headed in the right direction.

ormsbee:
I'm starting on this review now. Is the plan to get the basic functionality merged in first and then address performance and visualization issues later? Or is it your intention for this PR to eventually cover all that functionality? I have a mild preference for the former, but either is fine -- I'm just trying to make sure I understand the intended scope here.

Thank you.

tubular/tests/test_prune_modulestore.py
```diff
@@ -0,0 +1,57 @@
...
+    input_file = u"test_prune_modulestore_data.json"
+
+    def remove_output_file(self):
+        """
+        Removes residual test output
+        """
+        # check if the specified file exists
+        file_exists = os.path.isfile(self.output_file)
+
+        assert isinstance(file_exists, object)
```
ormsbee: What is the purpose of this assertion?

tubular/scripts/prune_modulestore.py
eltoncarr-ms: We've had ongoing discussions about the need for re-linking structures to origin. Given the conversations, we made this optional. Additionally, there is an additional overhead for re-linking (primarily the batch update needed post-pruning). If necessary, we can default it to true instead.

- [x] Default relink structures option to True

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Yes. Libraries don't have [draft|published]-branch for example.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: structures_list is already a list

tubular/scripts/prune_modulestore.py
eltoncarr-ms: That's the target of our next iteration. First focus was correctness. Accounting for performance is next.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Agreed. See comments above

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Certainly. 

- [x]  Switch to INFO logging where it is appropriate

eltoncarr-ms:
The essence of this PR was to get validation of the approach, just covering a very basic case. With that initial indication of correctness. We assumed there are other aspects we may not be privy to and have not accounted for (examples: how pruning impacts e-commerce, insight, mysql "snapshots" or related other components). 

With confirmation of the most basic case, we'd iterate further wrt performance and other concerns that may come up. Therefore, it seems easier to start with the basic concerns first (first round of review) and layer on performance, visualization, enabling single course/library pruning (in a subsequent review). This is my current preference. 

However, like you, I am also open to adding the additional pieces and get them all reviewed once and pushed once as one complete feature. If you feel otherwise, do let me know. If not, let's proceed with the basics first and follow up with other concerns.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+def get_structures(db, filter_enabled, active_versions_list):
+
+    """
+    Get all documents from the structures collection
+    """
+
+    # use filters (if required)
+    structure_filter = None
+    if filter_enabled:
+        structure_filter = get_structures_filter(active_versions_list)
```
ormsbee: Is this redundant? Do we ever send `active_versions_list` but have `filter_enabled` be `False`?

tubular/scripts/prune_modulestore.py
ormsbee: I think what @doctoryes means that that:

```python
structures_list = []
for structure_doc in resultset:
    structures_list.append(structure_doc)

return structures_list
```

Is equivalent to:

```python
return list(resultset)
```

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+def prune_structures(db, structures_to_remove):
+
+    """
+    Prune the specified documents from the structures collection
+    """
+
+    # TODO: (performance): chunk up the removals to minimize possible production impact
+    if len(structures_to_remove) == 0:
+        return
+
```
ormsbee: Nitpick: Python idiom is to use the falsiness of empty lists, and do something more like:

```python
if not structures_to_remove:
    return
```

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    structures_removal_filter = {'$in': []}
+
+    for structure_objectid_str in structures_to_remove:
+        structures_removal_filter['$in'].append(ObjectId(structure_objectid_str))
```
ormsbee: Nitpick: Might be a little clearer as a comprehension.

```python
structures_removal_filter = {
    '$in': [ObjectId(oid) for oid in structures_to_remove]
}

File: tubular/scripts/prune_modulestore.py (hunk 1 above)
Comment (eltoncarr-ms) [reply tubular/scripts/prune_modulestore.py]: - [x] Improve option message for the database name option

File: tubular/tests/test_prune_modulestore.py [hunk 14]
```diff
@@ -0,0 +1,57 @@
...
+            # there is residual output
+            os.remove(self.output_file)
+
+    def setUp(self):
+        # make sure we remove any residual output file
+        # this is defensive
+
+        super(TestModuleStorePruning, self).setUp()
+
+        self.remove_output_file()
```
ormsbee: FWIW, the [`tempfile`](https://docs.python.org/2/library/tempfile.html) package can help with the management of short lived files for testing (or other purposes).

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    else:
+        # we are using live data
+        # establish database connection
+        LOG.debug("Establishing database connection")
+        db_client = get_database(connection, database_name)
+
+        # get the data: active versions (courses/library) and accompanying structures
+        # get a dictionary listing all active versions
+        active_versions = get_active_versions(db_client, active_version_filter)
+        LOG.debug("{0} active versions identified.".format(len(active_versions)))
```
ormsbee: Nitpick: With logging, the preferred way to do string formatting would look like:

```python
LOG.debug("%s active versions identified", len(active_versions))
```

This allows certain smart log aggregators to group log messages together, where using the `"{0}"` style of format string does the text substitution before the logging system can see it.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+        # get the accompanying structures
+        filter_enabled = (active_version_filter is not None and len(active_versions) > 0)
```
ormsbee: Nitpick: Using truthiness of values, I think you can write this out as:

```python
filter_enabled = active_version_filter and active_versions
```

tubular/scripts/prune_modulestore.py
ormsbee: Sorry, I don't quite understand that. What active_versions_docs would be excluded by this if statement?

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    """
+    Get all documents from the active_versions collection
+    """
+
+    # establish the active version filter (if required)
+    active_version_filter = None
+
+    if active_version_list is not None:
+        active_version_filter = get_active_version_filter(active_version_list)
```
ormsbee: Nitpick: This might read more clearly if `get_active_version_filter` returned `{}` when it got a `None` for a `active_version_list`. So the above three lines becomes:

```python
    active_version_filter = get_active_version_filter(active_version_list)
```

And the actual call for the result set loses the conditional and just becomes:

```python
    resultset = db.modulestore.active_versions.find(active_version_filter, fields)
```

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    # establish the query filter  (respecting cases where no value is specified)
+    return get_query_filter(av_filter)
+
+
+def get_structures_filter(active_version_list):
+
+    """
+    Generate document filter for bulk querying the structures collection
+    """
```
ormsbee: What is the intended behavior of this method if an empty list is passed as the argument? Should it return a filter that will get everything? Nothing? Error?

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+import click
+import click_log
+from bson.objectid import ObjectId
+from pymongo import MongoClient
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+LOG = logging.getLogger(__name__)
+
+# the dictionary keys to track for active versions
+TARGET_ACTIVE_VERSIONS_KEYS = [u'library', u'draft-branch', u'published-branch']
```
ormsbee: Does it matter what the names of these branches are? Shouldn't we just preserve all the structure docs pointed to in the values of the versions dict of an active_versions doc?

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    structure_prune_data = None
+    testmode_data = None
+    operation_status = 0
+
+    # ensure that version_rention 2+
+    if version_retention < 2:
+        raise ValueError("Version rention must be at at least 2: origin and active version")
+
+    # Support loading sample dataset from file system for test purposes
+    if test_data_file is not None:
```
ormsbee: Suggestion: This general pattern of switching call behavior to invoke different methods depending on whether a `test_data_file` is passed in or not is prone to error creeping in later -- it's easy to add a piece of controller logic on one of the branches that isn't present in the other and harder to make sure the two paths are actually equivalent. It would be valuable to encapsulate the I/O logic into separate classes so that the `test_data_file` switch looks more like:

```
if test_data_file:
    backend = StaticModuleStoreBackend(test_data_file)
else:
    backend = MongoModuleStoreBackend(connection, database_name)

# Later calls are all the same and look like:
active_versions = backend.get_active_versions(active_version_filter)

#... etc.
```

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    # initialize the key variables
+    db_client = None
+    structure_prune_data = None
+    testmode_data = None
+    operation_status = 0
+
+    # ensure that version_rention 2+
+    if version_retention < 2:
+        raise ValueError("Version rention must be at at least 2: origin and active version")
```
ormsbee: Nitpicks:

* typo: "rention" -> "retention"
* You can use a [`Range option`](http://click.pocoo.org/6/options/#ranges) to do this validation where you specify the `click.option` decorator for this parameter.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+Options 2 & 3 support removing all structures or keeping a number of older structures (in support of
+user-specified retention policy)
+
+See more details regarding module store at
+http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/modulestores/split-mongo.html
+
+See additional details regarding the growth problem with the modulestore at
+https://openedx.atlassian.net/browse/PLAT-697
+
+"""
```
ormsbee: General note: A lot of this code assumes a level of familiarity with what these documents look like. The only detailed documentation that I'm aware of for that is in the module docstring for [xmodule.modulestore.split_mongo.split](https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/modulestore/split_mongo/split.py) in edx-platform. A pointer to that documentation, or a corrected version of it (if it's out of date) in this module might help people to better understand and later make fixes/contributions to this code.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    else:
+
+        # we are pruning the live data
+        prune_structures(db_client, structure_prune_candidates)
+
+        if relink_structures:
+            relink(db_client, structures)
+
+        operation_status = 1
```
ormsbee: We hardcode failure for this method?

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    """
+    Prune the static test data and return the results
+    """
+
+    pruned_static_data = []
+
+    for structure_doc in original_dataset[u'structures']:
+
+        if structure_doc[u'_id'] not in structures_to_remove:
+            pruned_static_data.append(structure_doc)
```
ormsbee: I believe the argument `structures_to_remove` here is currently a list. For large values of `structures_to_remove` for both this and the other `prune_structures()` method, it would be potentially much faster for large groups of deletes to pass in a set instead.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    original_dataset[u'structures'] = pruned_static_data
```
ormsbee: Please return a new data structure instead of mutating the old. This can really complicate debugging.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    # write the updated dataset
+    with open(output_file, 'w') as outfile:
+        json.dump(data, outfile)
+
+
+def prune_structures_static_data(original_dataset, structures_to_remove):
+
+    """
+    Prune the static test data and return the results
+    """
```
ormsbee: Please document the data structures being passed in here.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    LOG.debug("Relinking structure to original version")
+
+    index_position = 0
+    available_ids = []
+
+    # build a list of all available ids
+    available_ids.extend([structure_doc['_id'] for structure_doc in structures])
+
+    # iterate structures and relink to original
+    for structure_doc in structures:
```
ormsbee: Whenever you need a counter over a for loop, you can use [`enumerate`](https://docs.python.org/2/library/functions.html#enumerate). So it would look like:

```python
for i, structure_doc in enumerate(structures):
    # code here
```

If you need to start at 1 for the counter, you can use the optional `start` arg like:

```python
for i, structure_doc in enumerate(structures, start=1):
    # code here
```

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    """
+
+    LOG.debug("Relinking structure to original version")
+
+    index_position = 0
+    available_ids = []
+
+    # build a list of all available ids
+    available_ids.extend([structure_doc['_id'] for structure_doc in structures])
```
ormsbee: If `available_ids` is only being used for set inclusion operations, you can make a set out of it like:

```python
available_ids = {structure_doc['_id'] for structure_doc in structures}
```

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    # iterate structures and relink to original
+    for structure_doc in structures:
+
+        if structure_doc["previous_version"] is not None and structure_doc["previous_version"] not in available_ids:
+
+            LOG.debug("{0} was not found in {1}".format(structure_doc["previous_version"], available_ids))
```
ormsbee: I realize we're addressing perf issues later, but could we flag this log message? From my understanding of how this works, this message could get really large.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    This function searches all structure documents for the one specified
+    """
+
+    # it is more efficient to use a structures list
+    # instead of separate db calls (which may be necessary for supporting
+    # pruning individual active versions)
+    for structure_doc in structures_list:
+
+        if structure_doc[lookup_key] == lookup_value:
+            return structure_doc
```
ormsbee: For future perf: This will likely get expensive for large values of `structures_list`. We'll probably need to make a lookup dict to map previous versions so that the check can be done in constant time instead of O(n) on each lookup.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Move support functions into utility class

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+
+def build_activeversion_tree(active_version, structures):
+
+    """
+    Build a tree representing the active_version and its tree of ancestors
+    from structures
+    """
```
ormsbee: This says it's building a tree, but it's always a flat list of ancestors, right? I realize that's still technically a tree, but it might cause confusion.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+    # map the tree for the identified structure
+    version_tree = []
+
+    # recursively identify the ancestors
+    if structure_doc is not None:
+
+        # build the tree
+        version_tree.append(str(structure_doc['_id']))
+
+        while structure_doc[u'previous_version'] is not None:
```
ormsbee: Has this been tested with course re-runs?

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+                    # if the tree exceeds the minimum number of elements,
+                    # identify tree elements that should be removed
+                    if tree_length > version_retention:
+
+                        # track the required version: first & last
+                        versions_to_retain.extend(version_tree[:2])
+
+                        # if relinking is not required, it is ok to remove the original version
+                        if not relink_structures:
+                            versions_to_retain.append(version_tree[-1])
```
ormsbee: I don't think we ever want to remove the original version, even if it's a meaningless empty course. I thought that the process of relinking meant that we would be modifying the "previous" pointer of the oldest preserved structure to point to "original". The original is used to determine things like common ancestry, and due to the way things are cached, we can get weird results if we change the original and do updates.

For instance, we might still have a cached current structure doc that points to original_1. If the pruning process deletes original_1 and updates the structure docs in MongoDB to point to original_2 as their new "original", then looking for ancestors (e.g. split's `find_ancestor_structures()`) against the cached original will silently fail in a hard to debug way.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,559 @@
...
+
+    # All trees have been processed.
+    # We now have a final list of structures to retain & remove
+
+    # remove duplicates from the versions_to_retain
+    versions_to_retain = list(set(versions_to_retain))
+
+    # remove structures on the versions_to_remove list that are on the versions_to_retain list
+    # this supports course re-runs and related links
+    versions_to_remove = list(set(versions_to_remove) - set(versions_to_retain))
```
ormsbee: Why cast to lists instead of leaving them as sets?

ormsbee (changes requested):
Thank you so much for taking this on! I'm done with my first pass. Please note that the items marked nitpicks are intended to be optional -- they're mostly small suggestions to simplify the code or make it more idiomatically Python.

Some high level concerns:

1. The specific MongoDB serialized representation bleeds through into the functions, making it harder to understand.

I realize in some cases this is unavoidable. `get_active_verisions()` needs to understand MongoDB serialization internals. But it currently returns the serialized active_versions docs rather than just the data it needs, so other methods have to know that it's a dictionary with a `versions` key, and inside of that it has one of the `TARGET_ACTIVE_VERSIONS_KEYS`...

The abstract problem space we're dealing with is making a tree and pruning it. I think that it would greatly help future maintainers understand what's going on if we deserialize from MongoDB (and the static file) in as small and isolated a layer as possible so we can immediately get back those simple objects where we do all our interesting work. Once we've determined the list of structure IDs that need deleting or relinking, we can send a simple list of those IDs back to the backend layer, and have it worry about the details of the serialized format.

2. We have some places where it looks like we're doing O(n)^2 operations on the number of structures, which could cause us trouble down the road.

I realize that we're not aiming to be performant with this PR, so I'm not going to gate this PR on addressing these issues. I just want to flag a few cases where I'm seeing this, to inform future work.

Thank you!

ormsbee:
Sorry for the delay in review. I'm afraid that almost all of my comments are marked "outdated" because I didn't realize you had done the refactoring of moving all the logic to a new file. Doing a quick glance, it looks like much of the code stayed the same after the move though, so I hope it's still useful to you.

pwnage101:
It occurs to me that this repo is python 3, so you should also be able to safely drop all the `u` prefixes before string literals.

eltoncarr-ms:
@pwnage101, great idea. I was hoping to maintain the status quo.

- [x] Update string literals

eltoncarr-ms:
@ormsbee , yes, I made an update with the available suggested changes. I will review your items and respond/provide updates as required

tubular/tests/test_prune_modulestore.py
eltoncarr-ms: It's the equivalence of a null check. Thinking more about it, it doesn't seem necessary. 

- [x] Remove the isinstance assertion in remove_output_file

tubular/tests/test_prune_modulestore.py
eltoncarr-ms: The initial though on this was to use this output file as an input for driving visualization. Inside the UT, it will certainly be short lived and the tempfile package will prove handy.

- [x] Leverage tempfile package for handling output files in the UT

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Add link to [xmodule.modulestore.split_mongo.split](https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/modulestore/split_mongo/split.py) in the prune_modulestore.py module docstring

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Initially, we were only privy to those types. My major concern was whether or not other types would require extra handling. Without a definitive answer to that, we took a defensive posture of limiting to what we could explicitly account for instead of making a blanket assumption.

@ormsbee , here are some follow up questions:
1. are there other versions that should be supported? 
2. is there documentation on these versions? 
3. better still, can we confirm that all versions follow the same structures pattern?

- [x]  Follow up on removing the version restrictions

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [ ] ~~Clean up retention typo~~ (not necessary. already done)
- [x] Leverage click range option to restrict version_retention value

tubular/scripts/prune_modulestore.py
eltoncarr-ms: I've simplified by limiting the test data path to UT.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Update logging format to old style

tubular/scripts/prune_modulestore.py
eltoncarr-ms: That would be an interesting proposition... It's inadvertent. Apparently, I hadn't noticed these were meant to be treated as exit codes. This has been refactored.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Document the data structures passed (where applicable)

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Leverage sets for handling large lists

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] return a new data set with updated output

tubular/scripts/prune_modulestore.py
eltoncarr-ms: ~~Yes, it is redundant.~~

On closer look, this seems to be a residue of an early attempt to enable active version course/library filtering for structures. It has been held back. I will remove the traces until next phase when we enable it.

- [ ] ~~Switch to evaluating only the active_versions_list filter~~
- [x] Remove structure filtering option

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Yes. Course re-runs was key scenario we did a lookout for. In fact, IIRC, course re-runs is what's driving the lower limit of the retention.

- [x] Augment test dataset with course re-run scenario

tubular/scripts/prune_modulestore.py
eltoncarr-ms: It seems +2 for not making re-linking optional.

- [x] Remove option for re-linking, making it mandatory

tubular/scripts/prune_modulestore.py
eltoncarr-ms: The subsequent operation is to iterate the items. Generally, lists are better at iteration. We could flag this in the perf iteration for evaluation.

- [x] Flag for performance evaluation

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Fallback to python "truthiness"

tubular/scripts/prune_modulestore.py
eltoncarr-ms: This is residue of initial work we did to support user-specified course/library for purging. I pulled that back for next iteration. The intended behavior is to return None if an empty list is passed and the caller is expected to take the appropriate action.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Force get_query_filter to return { } when no user filter is provided and simplify query

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Got it.

- [x]  Simplify generation of structures list with list conversion of the returned mongodb query cursor

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Fallback to python "truthiness"

tubular/scripts/prune_modulestore.py
eltoncarr-ms: So far, we have accounted only for active versions that have draft-branch, published-branch and library versions. It is a defensive posture to only allow what we have accounted for when considering that there may be others we are not aware of and their handling may be different. Remember, we are talking about pruning structures here. By our calculation, it would be better to err on the side of not including some versions for pruning and only pruning those we explicitly understood and count account for.

So part of this review is to draw on your contextual expertise to callout other versions we should factor in and confirm if they all follow same/similar structures pattern.

tubular/scripts/prune_modulestore.py
ormsbee: Okay. I think conservative is fine.

1. Those are the only ones we expect. Maybe we can just log if we find a key that we don't anticipate? That way if we do add a new type later on, we'll at least log something so we don't silently skip pruning -- while at the same time preserving the conservative behavior.
2. There is no documentation that I'm aware of aside from the link to source code comments I already mentioned.
3. If you mean for the three types already listed, I can test locally on my devstack if you'd like to help verify, but I'm guessing you folks have already done that for those three. If not, I can do that on Friday (I'm really booked up tomorrow).

Thank you!

tubular/scripts/prune_modulestore.py
ormsbee: FWIW, lists generally give you predictable ordering (though not here, since you're creating it from a set), but they aren't any faster at iteration than sets. I'm fine with doing this in the perf iteration.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Agreed. 

- [x] switch to inline list generation for the removal filter creation

tubular/scripts/prune_modulestore.py
eltoncarr-ms: This is good.

- [x] Switch to enumerate for iterating structures

tubular/scripts/prune_modulestore.py
eltoncarr-ms: It could, but that'd be in debug mode only.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Yes, I will flag this.

- [x] flag for performance improvement

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] clarify ancestral list building

doctoryes:
@eltoncarr One more thing: It's known that simultaneous editing in Studio by multiple users can cause two (or more) child structures from a single parent due to a race condition. When this situation happens, one of the children will become the next parent structure while any other children will be dead-end children - never to be used again. All those dead-end children can be safely deleted.

Your script doesn't necessarily need to deal with those dead-end child structures. But I wanted to make you aware of the possibility so that your clean-up code can be written to still function properly even when this situation is encountered.

eltoncarr-ms:
@doctoryes, thanks for the heads-up. Where applicable, we will take that into account. 

@all, we've covered all the review items and currently running validations. I will be updating the PR in 1-2 days.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] switch to inline set creation

tubular/scripts/prune_modulestore.py
eltoncarr-ms: - [x] Adding warming log line for an unexpected active version (ie: not library, draft-branch or published-branch)

eltoncarr-ms:
Hopefully, all review items have been addressed.

ormsbee:
Looking this over now.

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+        if output_file:
+            self.log("Saving the purged dataset to %s" % output_file)
+
+            # write the updated dataset
+            with open(output_file, 'w') as outfile:
+                json.dump(data, outfile)
+
+        elif output_file_object:
+            self.log("Saving the purged dataset using tempfile file-like object")
+            output_file_object.write(json.dumps(data).encode())
```
ormsbee, line 111: Nit: If this is a file object, doesn't `json.dump(data, output_file_object)` work here?

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+        """
+
+        pruned_structures = []
+
+        for structure_doc in structures:
+
+            if structure_doc['_id'] not in structures_to_remove:
+                pruned_structures.append(structure_doc)
+
+        return pruned_structures
```
ormsbee: Nit: Can be collapsed to a comprehension:

```python
    def prune_structures_static_data(self, structures, structures_to_remove):
        return [sdoc for sdoc in structures if sdoc['_id'] not in structures_to_remove]
```

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+            ** 'active_versions': list of active version dictionaries
+            ** 'structures': list of structure dictionaries
+        :return: dictionary representing the test data set
+
+        """
+
+        # check if the specified file exists
+        file_exists = os.path.isfile(dataset_file)
+
+        assert isinstance(file_exists, object)
```
ormsbee: This assertion will always pass (`os.path.isfile` always returns a boolean, which is always an object).

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+
+        """
+
+        av_filter = {'$in': []}
+
+        for active_version_id in active_version_id_list.split(","):
+            av_filter['$in'].append(ObjectId(active_version_id.strip()))
+
+        # establish the query filter (respecting cases where no value is specified)
+        return self.get_query_filter(av_filter)
```
ormsbee: Nit: Whenever you have an empty list that you then append to in a for-loop, it's often clearer as a comprehension.

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+                version
+                for
+                version
+                in
+                active_version['versions']
+                if
+                version
+                not in
+                self.target_active_versions_keys
+            ]
```
ormsbee: Nit: It's unusual to see a comprehension spread out this much. A more typical form would be:

```python
            unknown_versions = [
                 version for version in active_version['versions']
                 if version not in self.target_active_versions_keys
             ]
```

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        # initialize our active versions dictionary
+        active_versions = []
+
+        # TODO: apply filters here in support of a user limiting the pruning operation to 1+ course/active versions
+        if active_version_filter is None:
+            resultset = self.db.modulestore.active_versions.find({}, fields)
+        else:
+            resultset = self.db.modulestore.active_versions.find(active_version_filter, fields)
```
ormsbee, line 247: Nit: This conditional is no longer necessary, right?

tubular/tests/test_prune_modulestore.py
```diff
@@ -0,0 +1,243 @@
...
+
+        """
+
+        found = False
+        for structure_doc in structures:
+            if structure_doc['_id'] == key:
+                found = True
+                break
+
+        return found
```
ormsbee: Nit: You can use `any` and `all` for situations like this where you're iterating a sequence and returning a boolean from it:

```python
def find_structure(self, structures, key):
    return any(structure_doc['_id'] == key for structure_doc in structures)
```

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+
+        # defensive: ensure that version_retention >=3
+        if version_retention < self._minimum_version_retention:
+
+            message_template = "Version retention of %s is below the minimum allowed and is being updated to %s"
+            self.log(message_template % (version_retention, self._minimum_version_retention), "info")
+
+        # iterate the active versions and build the associated version tree
+        for active_version in active_versions:
+
```
ormsbee: Nit: You can use `enumerate(activer_versions, start=1)` to get the counter.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # prune structures
+    structure_prune_candidates = structure_prune_data['versions_to_remove']
+    LOG.info("%s structures identified for removal", len(structure_prune_candidates))
+
+    # we are pruning the live data
+    module_store.prune_structures(structure_prune_candidates)
+
+    # relinking is mandatory
+    module_store.relink(structures)
```
ormsbee, line 117: If the script stops running partway through the pruning (or re-linking) parts, do we ever end up in a broken state? Or will re-running the script just work?

ormsbee (approved):
Some nits (optional) and an open question about how we do failure recovery. But nothing blocking this first version from merging.

tubular/scripts/prune_modulestore.py
```diff
@@ -0,0 +1,122 @@
...
+
+    """
+    Main script entry point for pruning the edxapp modulestore structures
+    """
+
+    # initialize the key variables
+    module_store = modulestore.ModuleStore(logger=LOG)
+    structure_prune_data = None
+
+    # we are using live data
```
doctoryes: Nit: Is this comment necessary?

tubular/modulestore.py
```diff
@@ -0,0 +1,527 @@
...
+        """
+
+        if not mongo_database_connection:
+            client = MongoClient()
+        else:
+            client = MongoClient(mongo_database_connection)
+
+        self.db = client[mongo_database_name]
+
+    def log(self, message, message_type="info"):
```
doctoryes, line 63: Nit: A global `LOG` could be used in both the script and the tests to remove the need for this class-based logging function.

doctoryes (approved):
Just a couple of nits. The script may need some iteration once run against production data. But I'm fine with this version as a starting point. Thanks, @eltoncarr !

tubular/modulestore.py
eltoncarr-ms: No, it doesn't.

tubular/modulestore.py
eltoncarr-ms: Love it

tubular/modulestore.py
eltoncarr-ms: It cost just as much to go either way. However, I though of making this option available in case someone wanted to target a single active_version instead of pruning everything. Additionally, the visualization option we had doesn't deal very well with large data sets. Therefore, this condition primarily supports a validation scenario.

Either way works for me but I lean more towards keeping it and a validation option. Keeping it means a bit more work. Removing it means less work but less opportunity for others to validate.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: As is, it will end up in a "broken" state. My thought was to use a snapshot of the data saved to file system after the complete pull from mongo or in between iterations. With this snapshot, yes, it will be completely resumable.

tubular/scripts/prune_modulestore.py
eltoncarr-ms: Not really.

tubular/modulestore.py
eltoncarr-ms: Wrapping it provide a bit more control. I personally prefer staying away from use of global variables where possible.

PR #185, opened 2017-08-31
Perf POC for Split Pruning

ormsbee:
@eltoncarr, @macdiesel, @feanil, @doctoryes: I think I've got this cleaned up to a state where it's reviewable. This version separates the plan generation and pruning into separate commands. We've run it on loadtest with about 5 million documents with the following results:

* 1.7 TB deleted (almost all the structures)
* Plan creation: 2 hours, almost entirely MongoDB query time
* Pruning: 8.5 hours for a pruning rate that's safe for running on prod
* Pruning settings: --batch-size=1000 --delay=5000

This should require no downtime to run. It can be resumed if interrupted (we tested this on the loadtest machine). It should be safe to run old plan files  as well -- e.g. create the plan file tonight, run pruning tomorrow.

Please let me know what you think.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,13 +3,14 @@ backoff>=1.3.2,<1.4.0
 boto==2.43.0
 click>=6.2,<7.0
 click-log==0.1.8
+edx-opaque-keys==0.4.0
 enum34 >=1.1.6, <2.0; python_version < '3.4'
 freezegun==0.3.8
 GitPython==2.1.1
 jenkinsapi==0.3.3
 lxml >= 3.7, <3.8
 PyGithub==1.29
-pymongo==3.5.0
+pymongo==3.5.1
 pytz==2016.10
 pyyaml==3.12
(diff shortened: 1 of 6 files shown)


ormsbee:
FWIW, the biggest constraint is memory usage. 5M structures amounts to about 1.3GB of RAM on my machine. Mostly, it's about preferring tuples to dicts and using generators for intermediate data structures. I chose 5M because that's about how many we have in edx.org -- not sure if others have bigger databases they need to prune.

ormsbee:
I somehow broke logging in a non-obvious way (no errors, it's just no longer displaying). Should be okay to review otherwise though. :(

ormsbee:
@jdmulloy: This is the pruning script we were testing.

ormsbee:
Okay, logging fixed.

macdiesel (approved):
Looks good.

eltoncarr-ms:
It does look great. I will take this for a run.

PR #190, opened 2017-10-11
Add delay and batch to structures.py make_plan.

ormsbee:
This is to allow us to throttle reads which were causing MongoDB instability.

@feanil, @jdmulloy: Review please?

### DIFF
--- a/tubular/scripts/structures.py
+++ b/tubular/scripts/structures.py
@@ -84,8 +84,25 @@ def cli(ctx, connection, database_name):
         "to 2. Put 0 here if you want to prune as much as possible."
     )
 )
+@click.option(
+    '--delay',
+    default=5000,
+    type=click.IntRange(0, None),
+    help=(
+        "Delay in milliseconds between queries to fetch structures from MongoDB "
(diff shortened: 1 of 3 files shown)


ormsbee:
Note: This does batch/delay for structure doc reads, where the vast majority of time is spent iterating over millions of results. It does not batch the active structures query (which is one query that returns thousands of results).

feanil (approved):

jdmulloy (approved):

ormsbee:
@eltoncarr: Our loadtest server handled the `make_plan` command okay without batching/throttling, but we encountered load problems when running it on one of our production machines. This PR adds batch and delay settings `make_plan` (we're not sure about what the defaults should be yet). It's been merged to master, so pulling the latest tubular will get you those extra flags.

eltoncarr-ms:
Got it. Thanks for the update.

ormsbee:
@eltoncarr: One last update -- after this change, we ran pruning on prod without encountering any further performance issues. The `make_plan` command was run with the default batch and delay settings.

PR #191, opened 2017-10-17
OPS-2372 Retry the message prs in range query if it fails

coryleeio:


tubular/scripts/message_prs_in_range.py
```diff
@@ -135,7 +136,18 @@ def message_pull_requests(org,
...
+
+    number_of_tries = 10
+    pull_requests = []
+    while number_of_tries > 0:
+        if number_of_tries == 0:
+            LOG.error('Was not able to retrieve PR range from GitHub')
+            sys.exit(1)
+        try:
+            pull_requests = api.get_pr_range(base_sha, head_sha)
+        except GithubException:
```
feanil: Are all github exception due to rate limits?  Also, we should print the error before retrying so we know how often we are having to retry.

feanil: We should also probably back-off exponentially(sleep for more and more time) so that if the services is having issues we don't just hammer it more during the issue.

tubular/scripts/message_prs_in_range.py
```diff
@@ -135,7 +138,25 @@ def message_pull_requests(org,
...
+            LOG.info('Got PR Range')
+            break
+        except RateLimitExceededException:
+            number_of_tries = number_of_tries - 1
+            time_until_next_try += time_until_next_try
+            LOG.info("Failed to retrieve PR range will try again in {} seconds".format(time_until_next_try))
+            sleep(time_until_next_try)
+    for pull_request in pull_requests:
         LOG.info(u"Posting message type %r to %d.", message_type, pull_request.number)
         getattr(api, methods[message_type])(pr_number=pull_request, extra_text=extra_text)
```
jdmulloy: I think this line is calling the github api to post the messages to the PRs. Do we need to wrap this with an exception handler and retry logic as well?

jdmulloy (changes requested):

tubular/scripts/message_prs_in_range.py
jdmulloy: @feanil Sometimes it's a network timeout. The failure that motivated the ticket is such a case. I think we also need to handle network timeouts on the API calls.

jdmulloy (changes requested):
This looks like a good solution to me.

Can we also add the timeout exceptions to the backoff decorator? Sometimes the calls to the GitHub API timeout instead of returning a rate limit error.

tubular/scripts/message_prs_in_range.py
```diff
@@ -6,14 +6,18 @@
 from __future__ import absolute_import
 from os import path
 import sys
+import socket
 import logging
 import click
+import backoff
```
feanil, line 12: This is a new package right?  In that case it should be added to the requirements file for this repo so it gets correctly installed on the server.

feanil (commented):
Other than the one comment, this generally looks good.

tubular/scripts/message_prs_in_range.py
coryleeio: Nope, it's already in there, on line 2

feanil (approved):

jdmulloy (approved):

PR #195, opened 2017-10-27
Coryleeio/fix retry

coryleeio:
To summarize - The old retry logic catches all exceptions and retries, so I left the retry logic in places where it was before, the main changes in behaviour were a specific few methods.  these methods were rollback, deploy, and wait_for_task_completion, and new_asg

I have removed the additional retry logic around rollback and deploy, as it can cause these steps to repeat during red_black_deploy and additional infrastructure to be created.  I left the logic around new_asg, as it only repeats when the new rate limiting exception.  For wait_for_task_completion I now only retry if you are throttled, not if you have timed out, as timeouts are often a task failure.

### DIFF
--- a/tubular/asgard.py
+++ b/tubular/asgard.py
@@ -581,12 +581,6 @@ def elbs_for_asg(asg):
     return elbs
 
 
-@backoff.on_exception(backoff.expo,
-                      (RateLimitedException,
-                       TimeoutException,
-                       BackendError,
-                       ASGDoesNotExistException),
-                      max_tries=MAX_ATTEMPTS)
 def rollback(current_clustered_asgs, rollback_to_clustered_asgs, ami_id=None):
     """
     Rollback to a particular list of ASGs for one or more clusters.
@@ -670,12 +664,6 @@ def rollback(current_clustered_asgs, rollback_to_clustered_asgs, ami_id=None):
         }
 
 
-@backoff.on_exception(backoff.expo,
-                      (RateLimitedException,
-                       TimeoutException,
-                       BackendError,
-                       ASGDoesNotExistException),
-                      max_tries=MAX_ATTEMPTS)
 def deploy(ami_id):
     """
     Deploys an AMI as an auto-scaling group (ASG) to AWS.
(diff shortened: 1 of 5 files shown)


tubular/asgard.py
```diff
@@ -176,6 +212,9 @@ def wait_for_task_completion(task_url, timeout):
     raise TimeoutException("Timed out while waiting for task {}".format(task_url))
 
 
+@backoff.on_exception(backoff.expo,
```
feanil, line 210: Rather than putting a retry around this whole function, is it possible to put the retries on just the calls we expect to be non-sideeffecting?  Like reads of status?

feanil: Also, it looks like you are already wrapping the wait_for_task completion, so the only reason to wrap here is if we get a rate limit exception on the call to new_asg?

coryleeio: Actually i've now got wrapped everywhere that method is called, so doing it again just leads to redundant waiting, so thanks for pointing that out, I am removing it.

tubular/asgard.py
```diff
@@ -176,6 +209,9 @@ def wait_for_task_completion(task_url, timeout):
...
+                      (RateLimitedException),
```
feanil: If this is meant to be a tuple, I think you need a trailing `,` with just one item.

feanil (commented):
It looks like every call to handle_throttoling is preceded by _parse_json.  Why not just do the throttle handling within that function?

tubular/asgard.py
coryleeio: I'll add the trailing comma.

The reason I didn't put it in _parse_json is because I felt like it was out of scope, as that makes the function do more than parse JSON, but considering this is a parse_json which is local to asgard it is probably alright.  I will change the name of this function and move it in there to be more DRY

feanil: You can also change the name to make it more clear.  Maybe `_process_request`?

tubular/asgard.py
```diff
@@ -49,8 +50,27 @@
 
 LOG = logging.getLogger(__name__)
 
+MAX_ATTEMPTS = int(os.environ.get('RETRY_MAX_ATTEMPTS', 5))
 
-def _parse_json(url, response):
+
+def handle_throttling(json_response):
```
feanil: In python a leading underscore is a convention for methods internal to a module.  I think we should prefix this function name with a `_`

feanil (approved):

PR #205, opened 2017-12-22
Adding in all other acquia environments. Changing language to make it…

Dillon-Dumesnil:
… more clear we can deploy a branch. Adding in test file to use for new pipelines until we can deprecate the standard drupal_deploy file. Removing format_rsa_key.

@robrap


tubular/scripts/drupal_deploy.py
```diff
@@ -16,6 +16,7 @@
 from tubular import drupal  # pylint: disable=wrong-import-position
 
 
+# TODO: LEARNER-3645 REMOVE THIS FILE and RENAME drupal_deploy_test.py to drupal_deploy.py
```
robrap: For the purpose of history, you may want to just copy the contents of drupal_deploy_test.py into this file and then delete drupal_deploy_test.py.

tubular/tests/test_drupal.py
```diff
@@ -220,26 +220,27 @@ def test_deploy_failure(self, mock):
         Tests deploy raises BackendError when status != 200
         """
         mock.post(
-            drupal.DEPLOY_URL.format(env=ACQUIA_ENV, tag=TEST_TAG),
+            drupal.DEPLOY_URL.format(env=ACQUIA_ENV, branch_or_tag=TEST_TAG),
```
robrap, line 223: How do the tests all working with the newer code, or does it not matter?

tubular/scripts/drupal_deploy_test.py
```diff
@@ -0,0 +1,38 @@
...
+@click.option("--env", help="The environment to deploy code in.", type=str, required=True)
+@click.option("--username", help="The Acquia username necessary to run the command.", type=str, required=True)
+@click.option("--password", help="The Acquia password necessary to run the command.", type=str, required=True)
+@click.option("--branch_or_tag", help="The branch or tag name to be deployed to the environment.",
+              type=str, required=True)
+def deploy(env, username, password, branch_or_tag):
+    """
+    Deploys a given tag to the specified environment.
+
+    Args:
```
robrap: We're supposed to use "Arguments:".  See http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings

tubular/drupal.py
```diff
@@ -33,13 +33,27 @@
...
+        "acceptance-edx-mktg-webview.edx.org",
+        "acceptance.edx.org",
+    ],
+    "dev": [
+        "edxdev.prod.acquia-sites.com",
+        "dev-edx-mktg-backend.edx.org",
+        "dev-edx-mktg-edx.edx.org",
+        "dev-edx-mktg-webview.edx.org",
+        "dev.edx.org",
+        "www.dev.edx.org",
```
robrap, line 49: No change needed on this PR, but...
Not sure why we need www supported for any of these environments?
I know this question is out of scope of your change (sort of), but inconsistency just causes confusion.  Why does tubular need this whole list anyway?  Could we:
1. Clean up our aliases so we have all www supported?
2. Would there be any harm just removing www from valid environments for all of these?
I'd prefer #1, but that could be a separate PR/story.

setup.cfg
```diff
@@ -28,9 +28,9 @@ console_scripts =
     drupal_backup_database.py = tubular.scripts.drupal_backup_database:backup_database
     drupal_clear_varnish.py = tubular.scripts.drupal_clear_varnish:clear_varnish_cache
     drupal_deploy.py = tubular.scripts.drupal_deploy:deploy
+    drupal_deploy_test.py = tubular.scripts.drupal_deploy_test:deploy
     drupal_fetch_deployed_tag.py = tubular.scripts.drupal_fetch_deployed_tag:fetch_deployed_tag
     find_and_advance_pipeline.py = tubular.scripts.find_and_advance_pipeline:find_and_advance_pipeline
-    format_rsa_key.py = tubular.scripts.format_rsa_key:format_rsa_key
```
robrap, line 33: Please confirm, this was part of the other clean-up?

tubular/drupal.py
```diff
@@ -22,7 +22,7 @@
 CLEAR_CACHE_URL = "{root}/sites/{realm}:{site}/envs/{{env}}/domains/{{domain}}/cache.json".format(
     root=ACQUIA_ENDPOINT, realm=REALM, site=SITE
 )
-DEPLOY_URL = "{root}/sites/{realm}:{site}/envs/{{env}}/code-deploy.json?path={{tag}}".format(
+DEPLOY_URL = "{root}/sites/{realm}:{site}/envs/{{env}}/code-deploy.json?path={{branch_or_tag}}".format(
```
robrap, line 25: Where is this file getting used, and is it ok to switch this without breaking legacy?

tubular/tests/test_drupal.py
```diff
@@ -220,26 +220,27 @@ def test_deploy_failure(self, mock):
...
             json=DEPLOY_RESPONSE_WAITING,
         )
         mock.get(
             drupal.CHECK_TASKS_URL.format(id="2"),
             json=DEPLOY_RESPONSE_DONE
         )
-        self.assertTrue(drupal.deploy(env=ACQUIA_ENV, username=TEST_USERNAME, password=TEST_PASSWORD, tag=TEST_TAG))
+        self.assertTrue(drupal.deploy(
+            env=ACQUIA_ENV, username=TEST_USERNAME, password=TEST_PASSWORD, branch_or_tag=TEST_TAG))
```
robrap: I learned to place closing characters on a new line, aligned with the start of the line with the opening characters.  For example:
```
self.assertTrue(drupal.deploy(
    env=ACQUIA_ENV, username=TEST_USERNAME, password=TEST_PASSWORD, branch_or_tag=TEST_TAG
))
```

setup.cfg
Dillon-Dumesnil: Yes, this is part of clean-up Brian asked me to do since we don't format any rsa keys anymore

tubular/drupal.py
Dillon-Dumesnil: Every drupal file calls drupal.py. This change is fine because it is only a format variable and we are just feeding in other variables.

tubular/drupal.py
Dillon-Dumesnil: Those aliases come from what we have listed on acquia. Only some environments have www so I just made it account for all sites available for each environment.

tubular/tests/test_drupal.py
Dillon-Dumesnil: It doesn't matter since all that changed here is the format variable name

robrap (approved):
Still need tests to pass.  Let me know if anything major changes to make that happen.  Also, not sure if you want a reviewer who knows more about tubular, but this seems pretty basic.

PR #208, opened 2018-01-22
Retry search query failures on github

coryleeio:


coryleeio:
The short of this is factoring out all the places that call search_issues into a single method and wrapping that method with backup, catching the generic Github exception thrown by GithubPy that is thrown when the search endpoint returns a rate limit response, and bubbling that up as a non generic exception that we catch with backoff so that we can retry the method.  I also set the number of retries to a larger amount than usual so it should back off sufficiently to get more retries.  Note that this method has a special retry limit that is different from the other Github API endpoints and it throws a generic exception as opposed to the normal exception thrown for rate limiting.  As far as I can tell this shouldn't result in a change in behaviour aside from retrying the query(read op) until it succeeds

tubular/github_api.py
```diff
@@ -541,6 +548,42 @@ def merge_pull_request(self, pr_number):
...
+                See: https://help.github.com/articles/searching-issues-and-pull-requests/
+            github_type (string): optional legal values are 'pr' or 'issue'
+            base (string): optional Base branch
+            user (string): optional Github user
+            repo (string): optional Github repo
+
+        Raises:
+            GithubException: For unknown errors
+            SearchRateLimitError: If we have retried the search, and it has failed the specified number of times
+        """
```
jibsheet: Looks like we have one other raise command in the code, and I think GithubException is "unknown errors from github" - that was slightly clearer to me, but I don't have strong opinions on it all.

tubular/github_api.py
```diff
@@ -541,6 +548,42 @@ def merge_pull_request(self, pr_number):
         pull_request = self.get_pull_request(pr_number)
         pull_request.merge()
 
+    # We run this a few extra times than normal bc it may need to backoff up to a minute or more
+    @backoff.on_exception(backoff.expo,
+                          (SearchRateLimitError),
+                          max_tries=12)
```
jibsheet: This may be my failure to be able to read the backoff docs, but do we know this reaches a minute with 12 retries?

jibsheet (approved):
Small doc nit, code looks good

tubular/github_api.py
coryleeio: Heh, it's actually quite a bit more now that I looked it up

1s
2s
4s
8s
16s
32s
64s
124s
248s

on 9 steps, 
Here is the code

https://github.com/litl/backoff/blob/4f66972d80c90012eee1375374f3303db2a699e6/backoff/_wait_gen.py

I'm going to set a max value so that it'll retry a few times at 2 minutes before failing, since other jobs may consume some of our searches as well...

At the moment it'd probably get up to like 8 or 16 minutes, so that's probably too much

tubular/github_api.py
coryleeio: I'll update that

PR #211, opened 2018-01-25
Attempted to move args around to more closely mimic the original beha…

coryleeio:
Attempted to move args around to more closely mimic the original behaviour. Added logging in case iteration is necessary


jibsheet:
@feanil I looked at the diff between branches and this change makes sense to me.  I'll leave merging/running a gocd pipeline with it for you as I'll be out for a few hours today.

jibsheet (approved):

PR #213, opened 2018-03-08
Pull tags from the base AMI even when building from scratch.

feanil:
Otherwise downsteam jobs that expect the info about the current base AMI fail.

Example failures:

- https://gocd.tools.edx.org/go/tab/build/detail/STAGE_edxapp_M-D/812/message_pr_stage/1/publish_wiki_job
- https://gocd.tools.edx.org/go/tab/build/detail/STAGE_edxapp_M-D/812/message_pr_stage/1/message_pr_job

### DIFF
--- a/tubular/scripts/retrieve_base_ami.py
+++ b/tubular/scripts/retrieve_base_ami.py
@@ -61,10 +61,11 @@ def retrieve_base_ami(environment, deployment, play, override, out_file):
         sys.exit(1)
 
     try:
+        edp_ami_id = ec2.active_ami_for_edp(environment, deployment, play)
         if override:
             ami_id = override
         else:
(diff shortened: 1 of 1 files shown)


cpennington (approved):

jibsheet (approved):

jibsheet:
@feanil Unfortunately, this fails on builds from scratch for IDAs (xqueue in this case).
retrieve_base_ami.py is called without the edp when the override is specified, so I think we need to fix the gomatic script to pass those (so we can pull tags off)
Failure https://gocd.tools.edx.org/go/tab/build/detail/stage-xqueue/14/build_ami/1/prod_edx
I'll see if I can have a tiny PR to fix this

PR #216, opened 2018-04-26
Script to retrieve learners to retire from LMS.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2095

Just the basics for now - but this version works fine:
```
(jenkins_py_script) John-Eskews-MacBook-Pro-3:gdpr jeskew$ python get_learners_to_retire.py --lms_hostname https://gdpr-sq.sandbox.edx.org --client_id <client_id> --client_secret <client_secret>
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gdpr-sq.sandbox.edx.org
DEBUG:urllib3.connectionpool:https://gdpr-sq.sandbox.edx.org:443 "POST /oauth2/access_token HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gdpr-sq.sandbox.edx.org
DEBUG:urllib3.connectionpool:https://gdpr-sq.sandbox.edx.org:443 "GET /api/user/v1/accounts/retirement_queue/?cool_off_days=7&states=PENDING&states=COMPLETE HTTP/1.1" 200 2
```

To get this script to work on a devstack, first run *this* new script:
```
./provision-retirement-user.sh retirement retirement_service_worker
```
The client_id and client_secret will be logged to your terminal. Use them in the command shown:
```
python get_learners_to_retire.py --lms_hostname http://localhost --lms_port 18000 --client_id <client_id> --client_secret <client_secret>
```


tubular/scripts/get_learners_to_retire.py
```diff
@@ -0,0 +1,125 @@
...
+        'Authorization': 'JWT {}'.format(jwt)
+    }
+
+    # Request the learners to retire.
+    learners_url = '{hostname}/api/user/v1/accounts/retirement_queue/'.format(
+        hostname=lms_hostname
+    )
+    params = {
+        'cool_off_days': 7,
+        'states': ['PENDING', 'COMPLETE']
```
bmedx: I think the call receiving this is just looking for a comma-separated string of states (for better or worse). Might want to test that this actually behaves as expected and the call isn't just getting the last state in the list or something weird.

doctoryes:
Reviewers: I've pushed changes to make this script usable in our new Jenkins plan. Please review/re-review.

tubular/edx_api.py
```diff
@@ -0,0 +1,61 @@
...
+            lms_base_url, api_base_url, client_id, client_secret
+        )
+
+    def learners_to_retire(self, cool_off_days=7):
+        """
+        Retrieves a list of learners awaiting retirement.
+        """
+        params = {
+            'cool_off_days': cool_off_days,
+            'states': ['PENDING', 'COMPLETE']
```
bmedx: Are these just placeholder states? I'd think that we would want `PENDING` and all of the intermediate `*_COMPLETE `states in these list, but not `COMPLETE`.

tubular/jenkins.py
```diff
@@ -17,6 +19,34 @@
...
+        directory (str): Directory in which to create the properties files.
+    """
+    _recreate_directory(directory)
+
+    for learner in learners:
+        learner_name = learner['original_username'].lower()
+        filename = os.path.join(directory, 'learner_retire_{}'.format(learner_name))
+        with open(filename, 'w') as learner_prop_file:
+            for prop in learner.keys():
+                learner_prop_file.write('{}={}\n'.format(prop.upper(), learner[prop]))
```
bmedx: Is there a reason for going with this format instead of YAML or JSON? I think the returned user properties are going to be nested deeper than this due to the DRF serialization.

tubular/scripts/get_learners_to_retire.py
```diff
@@ -0,0 +1,85 @@
...
+                           output_dir):
+    """
+    Retrieves a JWT token as the retirement service user, then calls the LMS
+    endpoint to retrieve the list of learners awaiting retirement.
+    """
+    if config_file:
+        # If a config file is present, it overrides all passed-in params.
+        with io.open(config_file, 'r') as config:
+            config_yaml = yaml.load(config)
+        client_id = config_yaml['client_id']
```
bmedx, line 69: Will the client id and secret be the same across IDAs?

tubular/scripts/get_learners_to_retire.py
```diff
@@ -0,0 +1,85 @@
...
+        client_secret = config_yaml['client_secret']
+        lms_base_url = config_yaml['base_urls']['lms']
+
+    api = LmsApi(client_id, client_secret, lms_base_url, lms_base_url)
+
+    # Retrieve the learners to retire and export them to separate Jenkins property files.
+    export_learner_job_properties(
```
bmedx: Is there anything we need to do here logging-wise for failures or do you think Jenkins will be able to surface errors appropriately?

tubular/scripts/get_learners_to_retire.py
```diff
@@ -0,0 +1,85 @@
...
+        api.learners_to_retire(cool_off_days),
+        output_dir
+    )
+
+
+if __name__ == "__main__":
+    # pylint: disable=unexpected-keyword-arg, no-value-for-parameter
+    # If using env vars to provide params, prefix them with "RETIREMENT_", e.g. RETIREMENT_CLIENT_ID
```
bmedx, line 84: This is cool 👍

tubular/edx_api.py
doctoryes: Ack - good point! Hadn't put thought into it. I'll change these.

tubular/jenkins.py
doctoryes: This format is required for the Jenkins properties file that creates each slave job. Each file turns out looking something like this:
```
ORIGINAL_USERNAME=barr
CURRENT_STATE={'state_name': 'PENDING', 'state_execution_order': 1, 'id': 1}
RETIRED_USERNAME=
MODIFIED=2018-04-28T13:26:16.858383Z
ORIGINAL_NAME=
CREATED=2018-04-28T13:26:15.362395Z
ID=1
ORIGINAL_EMAIL=
LAST_STATE={'state_name': 'PENDING', 'state_execution_order': 1, 'id': 1}
USER={'profile': {'name': 'barr', 'id': 12}, 'email': 'barr@example.com', 'username': 'barr', 'id': 12}
RETIRED_EMAIL=
```

tubular/scripts/get_learners_to_retire.py
doctoryes: They should be the same per environment using the same LMS as an OAuth identity provider. All IDAs in that environment would use the same client id/secret.

tubular/scripts/get_learners_to_retire.py
doctoryes: I don't know at the moment. It could be that we need to add more error logging. Guess I was waiting to see the Jenkins integration result first.

tubular/scripts/get_learners_to_retire.py
doctoryes: Agreed! I stumbled across this feature somewhere.

iloveagent57 (approved):
Brian already brought the hard-hitting journalism, here.  The rest of this LGTM.

doctoryes:
@bmedx @pwnage101 Requested changes are in-place. Re-review?

bmedx (approved):
👍 Is there a plan to add tests?

doctoryes:
@bmedx Tests are now added - along with Python3.6 testing and I *think* local tox testing will work now.

tubular/edx_api.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    LMS API client with convenience methods for making API calls.
+    """
+    def learners_to_retire(self, cool_off_days=7):
+        """
+        Retrieves a list of learners awaiting retirement actions.
+        """
+        params = {
+            'cool_off_days': cool_off_days,
+            'states': [
```
macdiesel, line 47: Interesting. Do we have to send an exhaustive list of states to the server?

doctoryes: We implemented the endpoint to be flexible and filter on whatever states are sent to it. Since the LMS knows about all states, we *could* make another endpoint that knows about which states to send back for learners to advance towards retirement, which would remove the need to send all these states.

pylintrc
```diff
@@ -80,7 +80,10 @@ disable =
 	too-many-arguments,
 	too-many-locals,
 	unused-wildcard-import,
-	duplicate-code
+	duplicate-code,
+    no-else-return,
```
pwnage101: nit: inconsistent indentation

tubular/edx_api.py
pwnage101: Just to be clear, if the open source community wanted to remove a state, they don't need to remove it from this list, right?

Don't let this be a blocker, but it seems that at minimum the open source community is going to have to touch code in four different places and run a management command just to add a new state.  The docs will be their friend.

pwnage101 (approved):
just a nit

tubular/edx_api.py
doctoryes: True. But there's really no way that the Jenkins scripts can't know about the states. We *could* pass this list into the method and keep the states centralized somewhere in the Jenkins configuration. Maybe that'll become clearer as we add more methods in here.

PR #219, opened 2018-05-09
Fixes to issues in retire_one_learner.py and edx_api.py

bmedx:
- Better error when user is already retired
- Catch cases where forum retirement returns a 404 for unknown user
- Remove unnecessary and error prone log statement

### DIFF
--- a/tubular/edx_api.py
+++ b/tubular/edx_api.py
@@ -3,7 +3,7 @@ edX API classes which call edX service REST API endpoints using the edx-rest-api
 """
 import logging
 
-from slumber.exceptions import HttpClientError
+from slumber.exceptions import HttpClientError, HttpNotFoundError
 
 from edx_rest_api_client.client import EdxRestApiClient
 
@@ -118,8 +118,15 @@ class LmsApi(BaseApiClient):
         """
         Performs the forum retirement step of learner retirement
         """
+        # api/discussion/
         params = {'data': {'username': learner['original_username']}}
-        return self._client.api.user.v1.accounts.retire_forum.post(**params)
+        try:
+            return self._client.api.discussion.v1.accounts.retire_forum.post(**params)
+        except HttpNotFoundError as exc:
+            if "User not found" in str(exc):
+                print("User not found in forums, this is expected sometimes. Overriding to success.")
+                return True
+            raise
 
     def retirement_retire_mailings(self, learner):
         """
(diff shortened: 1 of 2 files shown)


tubular/edx_api.py
```diff
@@ -118,8 +118,15 @@ def retirement_retire_forum(self, learner):
...
         Performs the forum retirement step of learner retirement
         """
+        # api/discussion/
         params = {'data': {'username': learner['original_username']}}
-        return self._client.api.user.v1.accounts.retire_forum.post(**params)
+        try:
+            return self._client.api.discussion.v1.accounts.retire_forum.post(**params)
+        except HttpNotFoundError as exc:
+            if "User not found" in str(exc):
+                print("User not found in forums, this is expected sometimes. Overriding to success.")
```
schenedx, line 127: @bmedx I thought "User Not Found" is also addressed on cs_comment_services by @doctoryes. Is this necessary if that's the case? 
That said, I am fine handling the case here as well.

doctoryes (approved):

tubular/scripts/retire_one_learner.py
```diff
@@ -157,10 +158,13 @@ def _setup_or_fail(username, client_id, client_secret, lms_base_url, ecommerce_b
...
-        learner = APIS['LMS'].get_learner_retirement_state(username)
-        learner_state_index = _get_learner_state_index_or_fail(learner)
-
-        return learner, learner_state_index
+        try:
+            learner = APIS['LMS'].get_learner_retirement_state(username)
+            learner_state_index = _get_learner_state_index_or_fail(learner)
+            return learner, learner_state_index
+        except HttpNotFoundError:
+            _fail(ERR_BAD_LEARNER, 'Learner {} not found. '
```
schenedx: The message `Learner {} not found` is confusing to me. Can we be a bit more prescriptive?
`Learner {} is not found in PendingRetirement status. Please check if the learner is not present, or learner is already retired.`

schenedx (commented):
Minor comments

tubular/edx_api.py
bmedx: The changes in cs_comments_service will return a 404 status code instead of the current 500 when the user doesn't exist in forums. This change catches the 404 and transforms it into a success (since it's an expected condition) instead of bubbling the error up and stopping the retirement job with an error.

schenedx: Ah, that make sense. Thank you for the explanation

bmedx:
@schenedx updated the error message. @doctoryes if / when tests pass can you merge? We're still sorting out permissions on this repo.

doctoryes:
@bmedx Merged.

PR #223, opened 2018-05-15
Fix an issue with handling retiring users without forum accounts

bmedx:
https://openedx.atlassian.net/browse/PLAT-2141
Forums will 404, which will trigger an LMS 404. The string checking
in place previously didn't work, likely a Py2 vs Py3 thing, this just
simplifies the check to make sure this case gets caught.


doctoryes (approved):
Yep - 404s should just be eaten. Nothing to do...

PR #226, opened 2018-05-23
Add retry logic around mailing retirement call.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2152

### DIFF
--- a/tubular/edx_api.py
+++ b/tubular/edx_api.py
@@ -3,8 +3,9 @@ edX API classes which call edX service REST API endpoints using the edx-rest-api
 """
 import logging
 
+import backoff
 from six import text_type
-from slumber.exceptions import HttpClientError, HttpNotFoundError
+from slumber.exceptions import HttpClientError, HttpServerError, HttpNotFoundError
 
 from edx_rest_api_client.client import EdxRestApiClient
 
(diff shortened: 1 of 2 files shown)


pwnage101 (approved):

tubular/edx_api.py
```diff
@@ -126,6 +127,25 @@ def retirement_retire_forum(self, learner):
...
+        Simple logging handler for when timeout backoff occurs.
+        """
+        LOG.info('Trying again in {wait:0.1f} seconds after {tries} tries calling {target}'.format(**details))
+
+    @staticmethod
+    def _not_a_timeout(exc):
+        """
+        Return True if the exception was *not* caused by a timeout.
+        """
+        return not (exc.response.status_code == 500 and 'timed out' in exc.content)
```
bmedx: Seems ok, but I have had weird experiences in Tubular with the Py2/3 issues and strings in these exceptions not matching up as expected. Might be worth wrapping both in a `text_type` to make sure they're always the same thing.

bmedx (approved):

doctoryes:
OK - I've made the change suggested by @bmedx and have tested the happy path on Jenkins. The retry logic itself has *not* been tested, which means it's still suspect. But I'm going to go ahead and merge this to be expedient. The turn-around time on fixing a retry bug should be small.

PR #230, opened 2018-05-30
Remove deletion tags from ASGs that should not be removed.

feanil:
Sometimes the tag remains on an active ASG, and we have to go in and manually delete it.  This should automate that process.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,9 +1,12 @@
 CHANGES
 =======
 
-* Update
-* Update tests
-* Update
+* Remove deletion tags from ASGs that should not be removed
+* Relax Sailthru retire retry to be any non-500 error
+* Fix hard coded states to fetch retirement users for (#228)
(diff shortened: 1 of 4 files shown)


ChangeLog
```diff
@@ -1,9 +1,12 @@
 CHANGES
 =======
 
-* Update
-* Update tests
-* Update
+* Remove deletion tags from ASGs that should not be removed
+* Relax Sailthru retire retry to be any non-500 error
```
jibsheet, line 5: Are most of these platform changelogs?

feanil: The change log in this repo is generated but someone failed to generate it for changes they made here so I'm just getting it up to date.

jibsheet: Oh, fascinating, they were just weird to me

jibsheet (approved):

PR #231, opened 2018-05-30
Consolidate retirement states to make updates easier

bmedx:
- Also forces get_learners_to_retire.py to use a config file instead of allowing command line or env var overrides

### DIFF
--- a/tubular/scripts/get_learners_to_retire.py
+++ b/tubular/scripts/get_learners_to_retire.py
@@ -30,19 +30,6 @@ LOG = logging.getLogger(__name__)
     help='File in which YAML config exists that overrides all other params.'
 )
 @click.option(
-    '--client_id',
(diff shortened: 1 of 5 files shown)


tubular/scripts/get_learners_to_retire.py
```diff
@@ -53,28 +40,31 @@
...
-        # If a config file is present, it overrides all passed-in params.
-        with io.open(config_file, 'r') as config:
-            config_yaml = yaml.load(config)
-        client_id = config_yaml['client_id']
-        client_secret = config_yaml['client_secret']
-        lms_base_url = config_yaml['base_urls']['lms']
+    if not config_file:
+        click.echo('A config file is required.')
+
+    # If a config file is present, it overrides all passed-in params.
```
pwnage101: This comment seems outdated.

pwnage101 (approved):
the rest looks good to me.

PR #234, opened 2018-05-31
Minimal tests to smoke out simple errors in get_learners_to_retire.py

bmedx:


tubular/tests/test_get_learners_to_retire.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        with open('test_config.yml', 'w') as f:
+            fake_config_file(f)
+        result = runner.invoke(
+            get_learners_to_retire,
+            args=['--config_file', 'test_config.yml', '--cool_off_days', cool_off_days, '--output_dir', output_dir])
+        print(result)
+        print(result.output)
```
pwnage101, line 30: Are these intentional, or leftover from debugging?  BTW, you may need to make this python2/3 compatible.

bmedx: They're intentional, they generally hold the only actually relevant information in a failure and only show up when a test fails. Shouldn't need to be 2/3 safe as we only test against 3.

pwnage101 (approved):
right, forgot that tests don't need to be python2/3

PR #236, opened 2018-05-31
Error if an unexpectedly high number of users are being retired

bmedx:


pwnage101 (approved):
if we're getting about 5 requests per hour, 200 per hour threshold might be excessive, but we're currently running daily so lets cross that bridge when we get to it.

PR #239, opened 2018-06-15
PLAT-2177: Use different backoff for 504 gateway timeout errors.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2177

Test both 500 and 504 errors. Update Changelog.

### DIFF
--- a/tubular/edx_api.py
+++ b/tubular/edx_api.py
@@ -2,6 +2,7 @@
 edX API classes which call edX service REST API endpoints using the edx-rest-api-client module.
 """
 import logging
+from contextlib import contextmanager
 
 import backoff
 from six import text_type
@@ -13,6 +14,14 @@ from edx_rest_api_client.client import EdxRestApiClient
 LOG = logging.getLogger(__name__)
 
 OAUTH_ACCESS_TOKEN_URL = "/oauth2/access_token"
+
+
+class EdxGatewayTimeoutError(Exception):
+    """
+    Exception used to indicate a 504 server error was returned.
+    Differentiates from other 5xx errors.
+    """
+    pass
 
 
 class BaseApiClient(object):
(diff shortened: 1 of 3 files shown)


tubular/edx_api.py
```diff
@@ -161,34 +206,40 @@ def retirement_retire_forum(self, learner):
...
         params = {'data': {'username': learner['original_username']}}
         try:
-            return self._client.api.discussion.v1.accounts.retire_forum.post(**params)
+            with correct_exception():
+                return self._client.api.discussion.v1.accounts.retire_forum.post(**params)
         except HttpNotFoundError:
             return True
 
-    @_retry_lms_api(retry_statuses=[500, 504])
+    # @_retry_lms_api(retry_statuses=[500, 504])
```
bmedx: Looks like a couple of comments got left behind.

bmedx (approved):
I'd be slightly more comfortable with a test that explicitly checks that `_wait_one_minute()` got called in the 504 case, but not the 500. I'm not sure how complicated that is off hand. Looks good otherwise. 👍

tubular/edx_api.py
```diff
@@ -74,32 +83,72 @@ def inner(exc):  # pylint: disable=missing-docstring
...
+        yield
+    except HttpServerError as err:
+        if err.response.status_code == 504:  # pylint: disable=no-member
+            # Differentiate gateway errors so different backoff can be used.
+            raise EdxGatewayTimeoutError(text_type(err))
+        else:
+            raise err
+    except HttpClientError as err:
+        try:
+            LOG.error("API Error: {}".format(err.content))
```
pwnage101, line 134: This works under python2 because the literal "API Error: {}" is a bytestring, and so is err.content, but I don't think it will under python 3 since the types are different:

```
$ python3
>>> print('API Error: {}'.format('‡'.encode('utf-8')))
API Error: b'\xe2\x80\xa1'
```

You may need to explicitly decode err.content to get a text_type, then explicitly add the unicode literal prefix to the format string, upgrading everything to unicode.  E.g. `LOG.error(u"API Error: {}".format(err.content.decode('utf-8')))`

I dabbed with this in a separate area of the code where my real errors were being gobbled up by unicode decode errors:
https://github.com/edx/tubular/pull/237/files

tubular/edx_api.py
doctoryes: Thanks - fixed.

tubular/edx_api.py
```diff
@@ -64,42 +73,70 @@ def _backoff_handler(details):
...
-        return exc.response.status_code not in statuses
-    return inner
+    return backoff.constant(interval=60)
 
 
-def _retry_lms_api(retry_statuses=None):
+def _exception_not_internal_svr_error(exc):
     """
-    Decorator which enables retries with sane backoff defaults for LMS APIs.
+    Giveup method that gives up backoff upon any non-500 and 504 server errors.
```
pwnage101: nit: "non-5xx"

tubular/edx_api.py
```diff
@@ -64,42 +73,70 @@ def _backoff_handler(details):
     LOG.info('Trying again in {wait:0.1f} seconds after {tries} tries calling {target}'.format(**details))
 
 
-def _exception_not_like(statuses=None):
+def _wait_one_minute():
     """
-    Parameterized callback for backoff's "giveup" argument which checks that the exception does NOT have any of the
-    given statuses.
+    Backoff handler that waits for 60 seconds.
```
pwnage101: nit: technically this is a "backoff generator" rather than a "backoff handler".

pwnage101 (approved):
just comment nits

tubular/edx_api.py
doctoryes: I took this code from its existing location and moved it to this context manager. Did the incompatibility you describe exist before as well?

tubular/edx_api.py
doctoryes: Correct - will fix.

tubular/edx_api.py
doctoryes: Will fix.

doctoryes:
@bmedx I've tried and failed so far to write a successful test of `_wait_one_minute()` being called in one case but not the other - mostly because of the complexities of backoff and such. I plan to merge as-is and will view the upcoming GDPR user runs, verifying 504 behavior.

tubular/edx_api.py
pwnage101: Yes, I believe it existed before as well, so feel free to punt this until a later PR.

Basically, `requests` already solved this problem by providing an automatically decoded .text attribute in the response object, but slumber does not pass this attribute through to the exception (an oversight, IMO) which forces us to do the work that requests has already done.

PR #240, opened 2018-06-18
add google_api.py

pwnage101:
PLAT-2173

Testing:

```
>>> client.create_file_in_folder('root', 'testfile.txt', StringIO('hello world'), 'text/plain')
'1cvPXmuoLgWbDwRbgb-vnpLAC4gbisSP8'
>>> [f['name'] for f in client._client.files().list().execute()['files']]
['testfile.txt', 'photo.jpg', 'photo.jpg', 'photo.jpg', 'photo.jpg', 'photo.jpg']
>>> client.delete_files(['1cvPXmuoLgWbDwRbgb-vnpLAC4gbisSP8'])
>>> [f['name'] for f in client._client.files().list().execute()['files']]
['photo.jpg', 'photo.jpg', 'photo.jpg', 'photo.jpg', 'photo.jpg']
>>>
```

### DIFF
--- a/tubular/google_api.py
+++ b/tubular/google_api.py
@@ -1,0 +1,141 @@
+"""
+Helper API classes for calling google APIs.
+
+DriveApi is for managing files in google drive.
+"""
+import logging
+
+import backoff
+
+from googleapiclient.discovery import build
+from googleapiclient.errors import HttpError
+# I'm not super happy about this since the function is protected with a leading
+# underscore, but the next best thing is literally copying this ~40 line
+# function verbatim.
+from googleapiclient.http import _should_retry_response
+from googleapiclient.http import MediaIoBaseUpload
+from google.oauth2 import service_account
+
+LOG = logging.getLogger(__name__)
+
+
+class BaseApiClient(object):
+    """
+    Base API client for google services.
+
+    To add a new service, extend this class and override these class variables:
+
+      _api_name  (e.g. "drive")
+      _api_version  (e.g. "v3")
+      _api_scopes
+    """
+    _api_name = None
+    _api_version = None
+    _api_scopes = None
+
+    def __init__(self, client_secrets_file_path, **kwargs):
+        self.build_client(client_secrets_file_path, **kwargs)
+
+    def build_client(self, client_secrets_file_path, **kwargs):
+        """
+        Build the google API client, specific to a single google service.
+        """
+        credentials = service_account.Credentials.from_service_account_file(
+            client_secrets_file_path, scopes=self._api_scopes)
+        self._client = build(self._api_name, self._api_version, credentials=credentials, **kwargs)
+
+
+def _backoff_handler(details):
+    """
+    Simple logging handler for when timeout backoff occurs.
+    """
+    LOG.info('Trying again in {wait:0.1f} seconds after {tries} tries calling {target}'.format(**details))
+
+
+def _should_retry_google_api(exc):
+    """
+    General logic for determining if a google API response is retryable.
+
+    Args:
+        exc (googleapiclient.errors.HttpError): The exception thrown by googleapiclient.
+
+    Returns:
+        bool: True if the caller should retry the API call.
+    """
+    retry = False
+    if hasattr(exc, 'resp') and exc.resp:  # bizzare and disappointing that sometimes `resp` doesn't exist.
+        retry = _should_retry_response(exc.resp.status, exc.content)
+    return retry
+
+
+class DriveApi(BaseApiClient):
+    """
+    Google Drive API client.
+    """
+    _api_name = 'drive'
+    _api_version = 'v3'
+    _api_scopes = ['https://www.googleapis.com/auth/drive.file']
+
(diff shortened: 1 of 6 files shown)


tubular/google_api.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    _api_name = 'drive'
+    _api_version = 'v3'
+    _api_scopes = ['https://www.googleapis.com/auth/drive.file']
+
+    @backoff.on_exception(
+        backoff.expo,
+        HttpError,
+        max_time=600,  # 10 minutes
+        giveup=lambda e: not _should_retry_google_api(e),
```
doctoryes, line 83: This whole mechanism seems weird - but if it works, fine.

tubular/google_api.py
```diff
@@ -0,0 +1,141 @@
...
+        def callback(identifier, response, exception):  # pylint: disable=unused-argument,missing-docstring
+            if exception:
+                LOG.error(exception)
+            else:
+                LOG.info("Deleted file ID: {}".format(identifier))
+
+        batched_requests = self._client.new_batch_http_request(callback=callback)  # pylint: disable=no-member
+        for file_id in file_ids:
+            batched_requests.add(self._client.files().delete(fileId=file_id))  # pylint: disable=no-member
+        batched_requests.execute()
```
doctoryes, line 141: I assume with the `callback` that a missing file will just be continued over, deleting the files that are present.

tubular/tests/test_google_api.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        Test normal case for uploading a file.
+        """
+        fake_file_id = 'fake-file-id'
+        http_mock_sequence = HttpMockSequence([
+            # First, a request is made to the discovery API to construct a client object for Drive.
+            ({'status': '200'}, self.mock_discovery_response_content),
+            # Then, a request is made to upload the file.
+            ({'status': '200'}, '{{"id": "{}"}}'.format(fake_file_id)),
+        ])
```
doctoryes, line 37: That's certainly helpful!

tubular/tests/test_google_api.py
```diff
@@ -0,0 +1,132 @@
...
+        test_client = DriveApi('non-existent-secrets.json', http=http_mock_sequence)
+        response = test_client.create_file_in_folder(
+            'fake-folder-id',
+            'Fake Filename',
+            StringIO('fake file contents'),
+            'text/plain',
+        )
+        assert response == fake_file_id
```
doctoryes: Shouldn't you check to see if `mock_backoff_handler` is called once here?

tubular/tests/test_google_api.py
```diff
@@ -0,0 +1,132 @@
...
+            ({'status': '200'}, self.mock_discovery_response_content),
+            # Then, a request is made to delete files.
+            ({'status': '200', 'content-type': 'multipart/mixed; boundary="batch_foobarbaz"'}, batch_response),
+        ])
+        test_client = DriveApi('non-existent-secrets.json', http=http_mock_sequence)
+        output = StringIO()
+        with capture_stdout(output):
+            test_client.delete_files(fake_file_ids)
+        for line, fileid in zip(output, fake_file_ids):
+            assert fileid in line
```
doctoryes: This test doesn't test the "no file present to delete" case - but might be too difficult for the value?

tubular/tests/discovery-drive.json
```diff
@@ -0,0 +1,3083 @@
+{
```
doctoryes, line 1: JFTR - I *did not* review this file. 😄

tubular/google_api.py
pwnage101: yea, I just found myself reading the function and starting to copy it until I realized that I would have just copied the whole function had I finished.

tubular/tests/test_google_api.py
pwnage101: Ah, I had introduced `mock_backoff_handler`, then decided that it wasn't needed.

The retry is implicitly checked because the response could not possibly have been equal to `fake_file_id` otherwise (due to the HttpMockSequence being constructed such that only the last response contains that data).

I'll add a comment.

tubular/tests/discovery-drive.json
pwnage101: Hah, I wanted to add a comment but JSON doesn't support comments, so then I was going to create a `discovery-drive.json.README` file instead but fogot.

tubular/google_api.py
pwnage101: yes, missing files get a message printed to the console, nothing more:

```
<HttpError 404 when requesting https://www.googleapis.com/drive/v3/files/2vfmBb6VPH-TtxvosKTGiq2-UEh4SLO9f? returned "File not found: 2vfmBb6VPH-TtxvosKTGiq2-UEh4SLO9f.">
```

bmedx (approved):
Not going to pretend I understand it all, but if there are issues when we test in stage we can add more tests for them then.

PR #241, opened 2018-06-20
Backoff after hitting API rate limit when messaging PRs

isabelgk:


isabelgk:
tox tests passed on latest commit

coryleeio (changes requested):
If you backoff on this method and the final call to message_prs fails (in the loop) throws an exception it is going to resend every notification before it.  You need to wrap a further down method that does only one thing.  As a general rule if the method writes to a anything it should only be called once in the method you are wrapping. You may even have to extract the singular writing network call into a new method and wrap that.

coryleeio (approved):

PR #247, opened 2018-07-06
add DriveApi.list_subfolders() for listing partner folders

pwnage101:
PLAT-2200

---

Manual testing:

```
>>> client.list_subfolders('17E1UoFlhXr3iu8D9yjPBc0pojVZ3hpfZ')
[{'name': 'edX', 'id': '11_eUyyv058zJA7NgHpkwfRnpg1U4jXp3'}]
```

`17E1UoFlhXr3iu8D9yjPBc0pojVZ3hpfZ` is the id of the top level folder for stage partner reporting, and the response is correct since there is only one subfolder.

Also, tests pass locally under python 2.7.

### DIFF
--- a/tubular/google_api.py
+++ b/tubular/google_api.py
@@ -101,7 +101,7 @@ class DriveApi(BaseApiClient):
         Returns: file ID (str).
 
         Throws:
-            apiclient.errors.HttpError:
+            googleapiclient.errors.HttpError:
                 For some non-retryable 4xx or 5xx error.  See the full list here:
                 https://developers.google.com/drive/api/v3/handle-errors
         """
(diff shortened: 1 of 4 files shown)


tubular/google_api.py
```diff
@@ -144,3 +144,47 @@ def callback(request_id, response, exception):  # pylint: disable=unused-argumen
...
+    @backoff.on_exception(
+        backoff.expo,
+        HttpError,
+        max_time=600,  # 10 minutes
+        giveup=lambda e: not _should_retry_google_api(e),
+        on_backoff=lambda details: _backoff_handler(details),  # pylint: disable=unnecessary-lambda
+    )
+    def list_subfolders(self, top_level, file_fields='id, name'):
+        """
+        List all subfolders of a tiven top level folder
```
doctoryes: Typo.

tubular/google_api.py
```diff
@@ -144,3 +144,47 @@ def callback(request_id, response, exception):  # pylint: disable=unused-argumen
...
+        extra_kwargs = {}  # only used for carrying the pageToken.
+        results = []
+        while True:
+            resp = self._client.files().list(  # pylint: disable=no-member
+                q="mimeType = 'application/vnd.google-apps.folder' and '{}' in parents".format(top_level),
+                fields='nextPageToken, files({})'.format(file_fields),
+                **extra_kwargs
+            ).execute()
+            results.extend(resp.get('files', []))
+            if 'nextPageToken' in resp and resp['nextPageToken']:
```
doctoryes: Nit: Maybe add a leading `if results and 'nextPageToken'....`?

tubular/tests/test_google_api.py
```diff
@@ -191,3 +201,88 @@ def test_delete_file_with_nonexistent_file(self, mock_from_service_account_file)
...
+            (
+                {'status': '200'},
+                self.mock_discovery_response_content,
+            ),
+            # Then, a request is made to list files.  The response contains a nextPageToken suggesting there are more
+            # pages.
+            (
+                {'status': '200', 'content-type': 'application/json'},
+                json.dumps({'files': fake_files_part_1, 'nextPageToken': 'fake-next-page-token'}).encode('utf-8'),
+            ),
```
doctoryes, line 252: What if a rate-limiting error or any other error is thrown between these two page fetches?

pwnage101: Then the request will start over from the first page.  This is not as efficient, but helps keep the code simple.

doctoryes: That's totally acceptable.

doctoryes (approved):

tubular/google_api.py
pwnage101: Do you mean `if resp and 'nextPageToken'....`?  The emptiness of `results` is not relevant, I'd argue, so long as a `nextPageToken` is provided (but definitely weird if the first page is empty).

doctoryes: I just meant that if `results` is empty, then no results where returned and no next page should ever be checked. In theory, `nextPageToken` should correspond accordingly to empty `results` - this would be just a fail-safe. I'm not insisting on this change - mere nit.

pwnage101: i went ahead and pushed a slight upgrade to your recommendation and checked `page_results` instead of the aggregate `results` for emptiness.

PR #250, opened 2018-07-16
Add script to delete expired partner reports.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2183

Adds implementations to both the Google API code and a click script for deleting the old GDPR partner reports, along with tests for it all.

### DIFF
--- a/tubular/google_api.py
+++ b/tubular/google_api.py
@@ -9,6 +9,7 @@ import logging
 from six import iteritems
 import backoff
 
+from dateutil.parser import parse
 from google.oauth2 import service_account
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
@@ -24,6 +25,9 @@ LOG = logging.getLogger(__name__)
 
 # Maximum number of requests per batch, according to the google API docs.
 GOOGLE_API_MAX_BATCH_SIZE = 100
+
+# Mimetype used for Google Drive folders.
+FOLDER_MIMETYPE = 'application/vnd.google-apps.folder'
 
 
 class BaseApiClient(object):
@@ -152,6 +156,26 @@ class DriveApi(BaseApiClient):
             batched_requests.add(self._client.files().delete(fileId=file_id))  # pylint: disable=no-member
         batched_requests.execute()
 
+    def delete_files_older_than(self, top_level, delete_before_dt, mimetype=None, prefix=None):
+        """
+        Delete all files beneath a given top level folder that are older than a certain datetime.
+        Optionally, specify a file mimetype and a filename prefix.
+
+        Args:
+            top_level (str): ID of top level folder.
+            delete_before_dt (datetime.datetime): Datetime to use for file age. All files created before this datetime
+                will be permanently deleted. Should be timezone offset-aware.
+            mimetype (str): Mimetype of files to delete. If not specified, all non-folders will be found.
+            prefix (str): Filename prefix - only files started with this prefix will be deleted.
+        """
+        all_files = self.walk_files(top_level, 'id, name, createdTime', mimetype)
+        file_ids_to_delete = []
+        for file in all_files:
+            if (not prefix or file['name'].startswith(prefix)) and parse(file['createdTime']) < delete_before_dt:
+                file_ids_to_delete.append(file['id'])
+        if file_ids_to_delete:
+            self.delete_files(file_ids_to_delete)
+
     @backoff.on_exception(
         backoff.expo,
         HttpError,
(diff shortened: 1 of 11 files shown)


tubular/google_api.py
```diff
@@ -147,6 +148,22 @@ def callback(request_id, response, exception):  # pylint: disable=unused-argumen
...
+            top_level (str): ID of top level folder.
+            delete_before_dt (datetime.datetime): Datetime to use for file age. All files created before this datetime
+                will be permanently deleted. Should be timezone offset-aware.
+        """
+        all_files = self.list_subfolders(top_level, 'id, createdTime')
+        file_ids_to_delete = []
+        for file in all_files:
+            if parse(file['createdTime']) < delete_before_dt:
+                file_ids_to_delete.append(file['id'])
+        self.delete_files(file_ids_to_delete)
```
pwnage101: list_subfolders will only list folders, not, for example, csv files.  You could modify list_subfolders to be more general and take a mimetype argument.  For example, `mimetype='application/vnd.google-apps.folder'` would be used for listing subfolders, and `mimetype='application/csv'` would be used for listing csv files.

bmedx: I'm not sure of the folder structures in play here, but afaik we're uploading to a shared folder that may have non-partner-report data in it. We might also need to take a wildcard filename as well as a mimetype to make sure we aren't deleting more than intended?

tubular/scripts/delete_expired_partner_gdpr_reports.py
```diff
@@ -0,0 +1,148 @@
...
+    # if there is no current exception, so we eat that here.
+    try:
+        _log(traceback.format_exc())
+    except AttributeError:
+        pass
+
+    exit(code)
+
+
+def _fail_exception(code, exc, message):
```
bmedx: If we're going to be reusing these we should probably pull them out into a shared library and make them consistent.

tubular/tests/test_delete_expired_reports.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def test_bad_secrets():
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        with open(TEST_CONFIG_FILENAME, 'w') as config_f:
+            fake_config_file(config_f)
+
+        with open(TEST_GOOGLE_SECRETS_FILENAME, 'w') as config_f:
+            config_f.write('{this is bad json')
```
bmedx, line 164: This line seems broken? Maybe broken in the original too?

bmedx: Oh nevermind I remember why I did this. Not very readable of me.

tubular/scripts/delete_expired_partner_gdpr_reports.py
doctoryes: Agreed - I meant to and then got lazy. Will do.

tubular/google_api.py
doctoryes: Agreed on all. I'll attempt to mirror the created filenames with a wildcard.

tubular/scripts/delete_expired_partner_gdpr_reports.py
```diff
@@ -0,0 +1,148 @@
...
+ERR_BAD_SECRETS = -4
+ERR_DELETING_REPORTS = -5
+ERR_BAD_AGE = -6
+
+
+def _log(info):
+    """
+    Convenience method to log text. We're currently not sending these to, ex, Splunk, but the
+    prepended text would make finding these entries easier if we go that route.
+    """
```
pwnage101: update this copypasta docstring to remove the suggestion that we don't log to splunk---we do send all jenkins console logs to splunk.

tubular/tests/test_google_api.py
```diff
@@ -201,6 +203,45 @@ def test_delete_file_with_nonexistent_file(self, mock_from_service_account_file)
...
+            for idx in range(2, 10, 2)
+        ]
+        fake_files = fake_newish_files + fake_old_files
+        http_mock_sequence = HttpMockSequence([
+            # First, a request is made to the discovery API to construct a client object for Drive.
+            (
+                {'status': '200'},
+                self.mock_discovery_response_content,
+            ),
+            # Then, a request is made to list files.
```
pwnage101, line 237: update this comment to reflect the request to delete files

doctoryes: Turns out - this request actually *is* a list request.

doctoryes:
Ready for re-review!

tubular/google_api.py
```diff
@@ -147,50 +150,108 @@ def callback(request_id, response, exception):  # pylint: disable=unused-argumen
...
+                LOG.debug("walk_files: Returned %s results.", len(page_results))
+
+                # Examine returned results to separate folders from non-folders.
+                for result in page_results:
+                    LOG.debug("walk_files: Result: %s", result)
+                    # Folders contain files - and get special treatment.
+                    if result['mimeType'] == FOLDER_MIMETYPE:
+                        if recurse and result['id'] not in visited_folders:
+                            # Add any undiscovered folders to the list of folders to check.
+                            folders_to_visit.append(result['id'])
```
pwnage101, line 246: love how this function doesn't technically recurse despite claiming to :)

tubular/google_api.py
```diff
@@ -147,50 +150,108 @@ def callback(request_id, response, exception):  # pylint: disable=unused-argumen
...
+                    # Determine if this result is a file to return.
+                    if result['id'] not in found_ids and (not mimetype or result['mimeType'] == mimetype):
+                        found_ids.append(result['id'])
+                        # Return only the fields specified in file_fields.
+                        results.append(dict([(k.strip(), result.get(k.strip(), None)) for k in file_fields.split(',')]))
```
pwnage101: Why not just append `result` at this point?  My assumption is that would be functionally equivalent since google should only return the fields we asked for.

bmedx (approved):
I have not even a nit on this. 🥇

tubular/google_api.py
bmedx: I can answer that one since I had the same thought. Back on line 225 it adds in `mimetype, parents` to the fields, so I think this is to prevent returning those fields if they weren't requested.

pwnage101: Oh, that makes sense.  Then I still have a nit: can this be simplified to use dict comprehension instead of passing a list comprehension into dict()?

Otherwise, approved.

doctoryes: Correct!

pwnage101 (approved):
see nit about the dict comprehension.

tubular/google_api.py
doctoryes: @pwnage101 I'll try out a dict comprehension - and go with whichever one is more readable. :)

doctoryes: The dict comprehension was *much* better! ☄️

PR #252, opened 2018-07-17
Correct the search query for finding approved open PRs.

feanil:
This search must have work at some point because it has been like this
for a while.  The search is incorrect because it queries for two
different values of `state` that contradict eachother.

It must have previously taken the first value and done the right thing.
However, it now seems to take the last value and so does not find the
intended open PRs that have been approved.  This change should remove
the ambiguity and correct the search behavior.


robrap (approved):
Don’t know what merged was for, so seems ok to me. I imagine another search is used for merged PRs to go public?

coryleeio (approved):

PR #255, opened 2018-07-20
Added scripts used when deploying frontend apps.

douglashall:
Related PRs:
https://github.com/edx/edx-gomatic/pull/427
https://github.com/edx/edx-internal/pull/500
https://github.com/edx/configuration/pull/4694


robrap (commented):
You may want someone like John or Brian to review.  Looks good to me, but not sure of the best pattern to follow.  

* Are you missing simlinks in the scripts directory?  Not sure.
* Looking at the drupal scripts, they all just point to a drupal file.  Should frontend follow that pattern?  What pattern/files did you follow?
Thanks.

tubular/scripts/frontend_build.py
```diff
@@ -0,0 +1,100 @@
...
+from os import path
+
+import click
+import yaml
+from git import Repo
+
+from tubular.git_repo import LocalGitAPI
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
```
doctoryes: This `sys.path` line needs to go before the `from tubular` statement above - it's there to allow the scripts to run without tubular being installed as a module.

tubular/scripts/frontend_build.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Uses the provided environment-specific configuration file to pass
+    environment variables to the frontend build.
+
+    Args:
+        env_config_file (str): Path to a YAML file containing environment configuration variables.
+        app_name (str): Name of the frontend app.
+    """
+    if not env_config_file:
+        LOG.error("Environment config file was not specified - failing.")
```
doctoryes: FYI - we've established a new script pattern recently for logging/exiting. To see it, search for LOG/FAIL/FAIL_EXCEPTION in this file:

https://github.com/edx/tubular/blob/master/tubular/scripts/retirement_partner_report.py#L27-L32

You might/might not want to use it.

tubular/scripts/frontend_build.py
```diff
@@ -0,0 +1,100 @@
...
+        sys.exit(1)
+    if not app_name:
+        LOG.error("App name was not specified - failing.")
+        sys.exit(1)
+
+    try:
+        env_vars = yaml.safe_load(io.open(env_config_file, 'r'))
```
doctoryes: Nit: Just to be sure to get the file closed ASAP, I typically use this pattern:
```
with io.open(env_config_file, 'r') as contents:
    env_vars = yaml.safe_load(contents)
```

tubular/scripts/frontend_build.py
```diff
@@ -0,0 +1,100 @@
...
+        LOG.warning("Frontend app config variables do not exist.")
+
+    proc = subprocess.Popen(['npm install'], cwd=app_name, shell=True)
+    return_code = proc.wait()
+    if return_code != 0:
+        LOG.error("Could not run `npm install` for app %s - failing.", app_name)
+        sys.exit(return_code)
+
+    env_vars = ['{}={}'.format(k, v) for k, v in app_config.items()]
+    proc = subprocess.Popen(
```
macdiesel: FWIW you probably could have skipped this python script and just called the NPM build directly from the command line in the gomatic code though perhaps the JSON version doc might be a little easier to create via python.

tubular/scripts/frontend_deploy.py
```diff
@@ -0,0 +1,82 @@
...
+        sys.exit(1)
+
+    bucket_name = env_vars.get('S3_BUCKET_NAME')
+    if not bucket_name:
+        LOG.error("No S3 bucket name configured for %s - failing.", app_name)
+        sys.exit(1)
+
+    bucket_uri = 's3://{}'.format(bucket_name)
+
+    proc = subprocess.Popen(
```
macdiesel: Same here, the s3 CLI tools are installed on the box, not sure you are getting much benefit from the python script.

tubular/scripts/frontend_build.py
douglashall: Yeah, I originally had it that way, but then we decided to pass environment variables to `npm run build`, so I needed to be able to access `edx-internal`.

tubular/scripts/frontend_deploy.py
douglashall: Need to read the environment-specific `S3_BUCKET_NAME` from `edx-internal`.

tubular/scripts/frontend_build.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    return_code = proc.wait()
+    if return_code != 0:
+        LOG.error("Could not run `npm run build` for app %s - failing.", app_name)
+        sys.exit(return_code)
+
+    # Add version.json file to build.
+    version = {
+        'repo': app_name,
+        'commit': LocalGitAPI(Repo(app_name)).get_head_sha(),
```
nasthagiri, line 92: Is there a race condition here where the latest commit on master changes between when it's captured here and then when the code is actually taken?  Should the commit that is captured here be passed to the pipeline?  Or vice versa?

tubular/scripts/frontend_build.py
```diff
@@ -0,0 +1,100 @@
...
+        'created': datetime.now().strftime('%Y-%m-%d %H:%M:%S'),
+    }
+    try:
+        json.dump(version, io.open('{}/dist/version.json'.format(app_name), 'w'))
```
doctoryes: Shouldn't this path be a passed-in param for flexibility in where the script actually runs from?

tubular/scripts/frontend_deploy.py
```diff
@@ -0,0 +1,82 @@
...
+import io
+from os import path
+import subprocess
+import sys
+import logging
+import click
+import yaml
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
```
doctoryes, line 19: This line isn't needed if you don't import anything from tubular.

tubular/scripts/frontend_build.py
douglashall: No, the build and version file are produced from the same local version of the frontend repo that gets cloned.

tubular/scripts/frontend_deploy.py
```diff
@@ -0,0 +1,82 @@
...
+    if not bucket_name:
+        LOG.error("No S3 bucket name configured for %s - failing.", app_name)
+        sys.exit(1)
+
+    bucket_uri = 's3://{}'.format(bucket_name)
+
+    proc = subprocess.Popen(
+        ' '.join(['aws s3 sync', app_dist, bucket_uri, '--delete']),
+        shell=True
+    )
```
doctoryes, line 72: The way we've accomplished things like this so far in tubular scripts is to code the AWS functionality in a separate file and use the code from there. For an example, see: https://github.com/edx/tubular/blob/master/tubular/ec2.py

All the calls using subprocess.Popen() are likely better off as a bash script - unless you need Python functionality wrapped around it - like retry/backoff and such.

doctoryes (commented):
The added scripts need a symlink from the top-level `scripts` directory.

tubular/scripts/frontend_build.py
douglashall: Ah, does [this](https://github.com/edx/edx-gomatic/pull/427/files#diff-04d7dd935a664e66be45d3624e1fd819R17) mean I don't need this line at all then?

tubular/scripts/frontend_build.py
douglashall: Thanks!

tubular/scripts/frontend_build.py
douglashall: Cool. Me too, this was a copy/paste that I was too lazy to change. :) I'll take off my lazy hat now.

tubular/scripts/frontend_build.py
douglashall: Yeah, again, I'm lazy. I will update.

tubular/scripts/frontend_deploy.py
douglashall: Cool. Thanks.

macdiesel (approved):
FWIW I'm ok with the using of these scripts to load in the data from the YAML files at runtime. Typically we would store these items in gomatic-secure and populate the pipeline with these configuration values at pipeline creation time. The advantage here is that these configuration items will be available for anyone at edx to view.

tubular/scripts/frontend_deploy.py
douglashall: Is there a place for bash scripts in this repo?

doctoryes: We haven't done that in the past because we've mainly used tubular with GoCD. If we needed bash scripts, we coded them into the gomatic code itself - here's an example:
https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/tasks/common.py#L1583-L1611

tubular/scripts/frontend_build.py
doctoryes: There's a reason for the sys.path addition to be there that I can't quite remember - except for enabling the possibility of running the scripts locally with having to install tubular while you're working on it.

tubular/scripts/frontend_deploy.py
douglashall: I think the issue is that I need to read a file from the `edx-internal` repo to grab the `S3_BUCKET_NAME` value. I talked about this with @macdiesel and there didn't seem to be a clear way to do that with a pure gomatic bash script.

I'm thinking it's overkill at this juncture to create a `/tubular/s3.py`. Let me know if you feel strongly the other way.

doctoryes: Are these scripts meant to be run within a GoCD pipeline? If so, our usual pattern for what you describe would be to make a GoCD pipeline env var containing the S3 bucket name to which the script would have access at run-time via a cmd-line arg.

But it's fine at this simple stage to do this as in this PR. If multiple scripts are doing this sync - or multiple steps are added - or error-handling and retry logic is needed, then I'd advocate for a `s3.py`.

doctoryes (approved):

douglashall:
FYI @georgebabey

PR #257, opened 2018-07-27
Add deletion_completed field to Partner Reporting

macdiesel:
The date a deletion was completed is a required field for partner reporting. This PR adds this field to the report using the 'created' field returned from the Retirement Partner report API in the LMS.

### DIFF
--- a/tubular/scripts/retirement_partner_report.py
+++ b/tubular/scripts/retirement_partner_report.py
@@ -127,6 +127,12 @@ def _get_orgs_and_learners_or_exit(config):
         # learners can appear in more than one dict.
         for learner in learners:
             usernames.append({'original_username': learner['original_username']})
+
+            # Use the datetime upon which the record was 'created' in the partner reporting queue
+            # as the approximate time upon which user retirement was completed ('deletion_completed')
+            # for the record's user.
+            learner['deletion_completed'] = learner['created']
+
             for org in learner['orgs']:
                 try:
                     reporting_org = config['org_partner_mapping'][org]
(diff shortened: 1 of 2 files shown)


tubular/tests/test_retirement_partner_report.py
```diff
@@ -121,8 +127,10 @@ def _fake_retirement_report_user(seed_val, user_orgs=None):
     return {
         'original_username': 'username_{}'.format(seed_val),
         'original_email': 'user_{}@foo.invalid'.format(seed_val),
-        'original_name': '阿碧 {}'.format(seed_val),
-        'orgs': user_orgs
+        'original_name': '{} {}'.format(UNICODE_NAME_CONSTANT, seed_val),
+        'orgs': user_orgs,
+        'created': DELETION_TIME,
+        'modified': DELETION_TIME,
```
pwnage101: we don't explicitly return the modified field in the API, so it's probably best leave this line out for consistency.

tubular/scripts/retirement_partner_report.py
```diff
@@ -127,6 +127,10 @@ def _get_orgs_and_learners_or_exit(config):
         # learners can appear in more than one dict.
         for learner in learners:
             usernames.append({'original_username': learner['original_username']})
+
+            # Rename the 'created' field for the partner reports
```
doctoryes: Nit: It'd be nice for the comment here to explain the situation better, something like:
```
Use the datetime upon which the record was 'created' in the partner reporting queue
as the approximate time upon which user retirement was completed ('deletion_completed')
for the record's user.
```

doctoryes (commented):
LGTM after comments addressed.

PR #259, opened 2018-07-30
Purge CloudFlare cache when frontend app is deployed.

douglashall:
Related PR: https://github.com/edx/edx-gomatic/pull/434

### DIFF
--- a/tubular/scripts/frontend_deploy.py
+++ b/tubular/scripts/frontend_deploy.py
@@ -13,6 +13,7 @@ from functools import partial
 from os import path
 
 import click
+import CloudFlare
 import yaml
 
 # Add top-level module path to sys.path before importing tubular code.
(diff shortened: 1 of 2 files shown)


tubular/scripts/frontend_deploy.py
```diff
@@ -74,6 +75,22 @@ def frontend_deploy(env_config_file, app_name, app_dist):
     if return_code != 0:
         FAIL(1, 'Could not sync app {} with S3 bucket {}.'.format(app_name, bucket_uri))
 
+    # Purge the Cloudflare cache for the frontend hostname.
+    # Frontend S3 buckets are named by hostname.
+    # Cloudflare zones are named by domain.
+    # Assumes the caller's shell has the following environment
+    # variables set to enable Cloudflare API auth:
+    # CF_API_EMAIL
+    # CF_API_KEY
```
doctoryes, line 84: The pattern we established for these types of config:

- Set the env vars as GoCD "Secure Variables" via edx-gomatic pipeline creation.
- Explicitly pass the secure variables to the tubular script, as in `--cf_api_email ${CF_API_EMAIL} --cf_api_key ${CF_API_KEY}`
- Pass the cmd-line params into the object that needs it, like `CloudFlare.CloudFlare(cf_api_email, cf_api_key)` in this case.

Why? We decided that explicit parameters are much better than relying on assumptions about what is and isn't defined in the environment. And we had cases where some scripts used env vars and some used cmd-line params - and it all became confusing.

tubular/scripts/frontend_deploy.py
```diff
@@ -74,6 +75,22 @@ def frontend_deploy(env_config_file, app_name, app_dist):
...
+    zone_name = bucket_name.split('.')[-2:]
+    data = {'hosts': [bucket_name]}
+    cloudflare_client = CloudFlare.CloudFlare()
+    try:
+        zone_id = cloudflare_client.zones.get(params={'name': zone_name})[0]['id']  # pylint: disable=no-member
+        cloudflare_client.zones.purge_cache.post(zone_id, data=data)  # pylint: disable=no-member
+    except (CloudFlare.exceptions.CloudFlareAPIError, IndexError, KeyError):
+        FAIL(1, 'Failed to purge the Cloudflare cache for hostname {}.'.format(bucket_name))
```
doctoryes, line 92: Nit: The code sends the same error code (1) for every failure currently. A simple enum and differentiated error codes might be appropriate for a script with this many failure conditions now.

doctoryes: Also, in what condition is the front-end in if the s3 sync worked above - and the cloudflare cache purge failed? Should any other action be taken by this script?

tubular/scripts/frontend_deploy.py
douglashall: Makes sense, but this script already depends on `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` being set as environment variables. To me it would be better to keep things consistent within this script and use environment variables for both.

tubular/scripts/frontend_deploy.py
douglashall: How would the differentiated error codes be used? Seems like a premature optimization in this case.

Eventually we will want to have a rollback procedure for frontend deployments. We're planning to take an iterative approach to get these pipelines up and running. For now, we will monitor pipeline failures and manually intervene should we happen to end up in an inconsistent state.

doctoryes (approved):
I'm satisfied with the answers given - just mentioned them as things to think about.

PR #262, opened 2018-07-31
Tag partners in Drive comment notifications

pwnage101:
At a high level:
* Implements a new DriveApi function list_permissions_for_files.
* Refactors create_comments_for_files to allow different comments per file.
* Updates the retirement_partner_report script to add tags to comments.
* Adds "future" to requerements, one small backstep for mankind.

PLAT-2265

### DIFF
--- a/tubular/google_api.py
+++ b/tubular/google_api.py
@@ -6,7 +6,7 @@ DriveApi is for managing files in google drive.
 from itertools import count
 
 import logging
-from six import iteritems
+from six import iteritems, text_type
 import backoff
 
 from dateutil.parser import parse
@@ -28,6 +28,13 @@ GOOGLE_API_MAX_BATCH_SIZE = 100
 
 # Mimetype used for Google Drive folders.
 FOLDER_MIMETYPE = 'application/vnd.google-apps.folder'
+
+
+class BatchRequestError(Exception):
+    """
+    Exception which indicates one or more failed requests inside of a batch request.
+    """
+    pass
 
 
 class BaseApiClient(object):
@@ -267,7 +274,7 @@ class DriveApi(BaseApiClient):
         giveup=lambda e: not _should_retry_google_api(e),
         on_backoff=lambda details: _backoff_handler(details),  # pylint: disable=unnecessary-lambda
     )
-    def _create_comments_for_files(self, file_ids, content, fields='id'):
+    def _create_comments_for_files(self, file_ids_and_content, fields='id'):
         """
         Retryable helper function for create_comments_for_files()
 
@@ -281,9 +288,10 @@ class DriveApi(BaseApiClient):
 
         # Generate arbitrary (but unique in this batch request) IDs for each file, so that we can recall the
         # corresponding file_id for a batch response item.
+        file_ids, _ = zip(*file_ids_and_content)
         file_id_to_request_id = dict(zip(
             file_ids,
-            (str(n) for n in count()),
+            (text_type(n) for n in count()),
         ))
         # Create a flipped mapping for convenience.
         request_id_to_file_id = {v: k for k, v in iteritems(file_id_to_request_id)}
@@ -297,7 +305,7 @@ class DriveApi(BaseApiClient):
                 LOG.info('Successfully created comment on file "{}".'.format(file_id))
 
         batched_requests = self._client.new_batch_http_request(callback=callback)  # pylint: disable=no-member
-        for file_id in file_ids:
+        for file_id, content in file_ids_and_content:
             batched_requests.add(
                 self._client.comments().create(fileId=file_id, body={u'content': content}, fields=fields),  # pylint: disable=no-member
                 request_id=file_id_to_request_id[file_id]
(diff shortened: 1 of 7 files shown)


tubular/scripts/retirement_partner_report.py
```diff
@@ -248,12 +248,31 @@ def _add_comments_to_files(config, file_ids):
...
+
+    partner_folders_to_permissions = drive.list_permissions_for_files(config['partner_folder_mapping'].values())
+
+    permissions = {
+        partner: partner_folders_to_permissions[config['partner_folder_mapping'][partner]]
+        for partner in file_ids
+    }
+
+    reader_emails = {
+        partner: [perm['emailAddress'] for perm in permissions[partner] if perm['role'] == 'reader']
```
bmedx: Are we sure that the primary role for all partner contacts will always be `reader`?

tubular/google_api.py
```diff
@@ -327,6 +327,83 @@ def create_comments_for_files(self, file_ids, content, fields='id'):
...
+        file_id_to_request_id = dict(zip(
+            file_ids,
+            (str(n) for n in count()),
+        ))
+        # Create a flipped mapping for convenience.
+        request_id_to_file_id = {v: k for k, v in iteritems(file_id_to_request_id)}
+
+        def callback(request_id, response, exception):  # pylint: disable=unused-argument,missing-docstring
+            file_id = request_id_to_file_id[request_id]
+            responses[file_id] = response['permissions']
```
bmedx: Will the `permissions` key be present if there is an exception? Similarly  for the callback in `_create_comments_for_files` - though the responses seem to eventually be thrown out there, we're just logging exception responses and ignoring them?

tubular/tests/test_google_api.py
```diff
@@ -641,4 +646,45 @@ def test_comment_files_with_duplicate_file(self, mock_from_service_account_file)
         ])
         test_client = DriveApi('non-existent-secrets.json', http=http_mock_sequence)
         with self.assertRaises(ValueError):
-            test_client.create_comments_for_files(fake_file_ids, 'some comment message')
+            test_client.create_comments_for_files(list(zip(fake_file_ids, cycle(['some comment message']))))
+
+    @patch('tubular.google_api.service_account.Credentials.from_service_account_file', return_value=None)
+    def test_list_permissions_success(self, mock_from_service_account_file):  # pylint: disable=unused-argument
+        """
+        Test normal case for listing permissions on files.
```
bmedx, line 654: Is there a case for exercising the error conditions to make sure they're being eaten and/or returned / raised correctly?

tubular/google_api.py
pwnage101: I forgot to do any reasonable error handling, reworked this section to hard fail on any error.

tubular/tests/test_google_api.py
pwnage101: added a new test to cover the an error case (file not found).

tubular/scripts/retirement_partner_report.py
pwnage101: I'll hopefully have an answer for this later.

pwnage101: the answer: filter out emails ending with '@edx.org'

pwnage101:
Ready for re-review. Just added email filtering based on whether or not the domain is "edx.org".

I know we discussed we may not even be doing these notifications, but may as well merge it since it's all here now.

tubular/google_api.py
```diff
@@ -327,6 +334,85 @@ def create_comments_for_files(self, file_ids, content, fields='id'):
                 For some non-retryable 4xx or 5xx error.  See the full list here:
                 https://developers.google.com/drive/api/v3/handle-errors
         """
+        file_ids, _ = zip(*file_ids_and_content)
+        if len(set(file_ids)) != len(file_ids):
+            raise ValueError('Duplicates detected in the file_ids_and_content list.')
```
doctoryes, line 339: Nit: Would it be worth including the duplicates here in the exception message? Looks like you could use:
https://docs.python.org/2/library/stdtypes.html#frozenset.symmetric_difference

tubular/scripts/retirement_partner_report.py
```diff
@@ -207,12 +213,40 @@ def _add_comments_to_files(config, file_ids):
...
+        partner: partner_folders_to_permissions[config['partner_folder_mapping'][partner]]
+        for partner in file_ids
+    }
+
+    # throw out all edx.org addresses, and flatten the permissions dicts to just the email:
+    external_emails = {
+        partner: [
+            perm['emailAddress']
+            for perm in permissions[partner]
+            if not perm['emailAddress'].endswith('@edx.org')
```
doctoryes: Nit: I don't *think* it should be an issue here, but perhaps:
`if not perm['emailAddress'].lower().endswith('@edx.org')` ?

bmedx (approved):

doctoryes (approved):

tubular/scripts/retirement_partner_report.py
pwnage101: hah

tubular/google_api.py
pwnage101: wouldn't work due to operator only working with sets.  I'm sure there's something else out there.

PR #267, opened 2018-08-09
encode all strings going out to the console

pwnage101:
This prevents UnicodeEncode errors on Build Jenkins by encoding all text before being written to `sys.stdout`.  This workaround is essentially a spot treatment because we only run some scripts under these limitations (python 2 on build jenkins).  Other related retirement scripts were not changed because they either only import from `tubular/scripts/helpers.py` which contains the workaround, or they aren't expected to print non-ascii characters.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 CHANGES
 =======
 
+* encode all strings going out to the console
 * Switch to unicodecsv for partner reporting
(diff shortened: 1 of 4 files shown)


bmedx (approved):
lgtm

doctoryes (approved):

PR #269, opened 2018-08-14
Fail when listing partner folders retuns zero results.

pwnage101:


bmedx (approved):
Nit: it'd be nice to have a test for this code path, but I wouldn't gate on it

doctoryes (approved):

PR #270, opened 2018-08-15
Perform Google API retries correctly

pwnage101:
Primarily, this new iteration of the code does not assume that
rate limiting can only happen at the HTTP level, but in fact it may
happen additionally at the logical request level.

Also, this adds chunking to the delete_files() method using this same
retry framework.

PLAT-2286

### DIFF
--- a/tubular/google_api.py
+++ b/tubular/google_api.py
@@ -43,6 +43,13 @@ class BatchRequestError(Exception):
     pass
 
 
+class TriggerRetryException(Exception):
+    """
+    Exception which indicates one or more throttled requests inside of a batch request.
+    """
+    pass
+
+
 class BaseApiClient(object):
     """
     Base API client for google services.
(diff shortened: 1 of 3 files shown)


tubular/google_api.py
```diff
@@ -58,16 +65,112 @@ class BaseApiClient(object):
...
+            # If the API throttled some requests, batch_callback would have populated the retry queue.  Reset the
+            # try_requests queue and indicate to backoff that there are requests to retry.
+            if retry_requests:
+                del try_requests[:]
+                try_requests.extend(retry_requests)
+                raise TriggerRetryException()
+
+        # func()'s side-effect is that it indirectly calls batch_callback which populates the responses dict.
+        func()
+        return responses
```
doctoryes, line 172: That's quite a method!

tubular/google_api.py
```diff
@@ -157,17 +254,32 @@ def delete_files(self, file_ids):
...
+            for file_id in file_ids_batch:
+                request_objects.append(self._client.files().delete(fileId=file_id))  # pylint: disable=no-member
+
+            # this generic helper function will handle the retry logic
+            responses_batch = self._batch_with_retry(request_objects)
+
+            responses.update(responses_batch)
+
+        if len(responses) != len(file_ids):
+            raise BatchRequestError('Error deleting one or more files/folders.')
```
doctoryes, line 282: Since we know which responses succeeded here, would it be worthwhile to tell the user which ones succeeded/failed? Or at least numbers of each? Or is that info discoverable in the other logging?

doctoryes: It's too bad that these response-gathering sections aren't quite the same enough to wrap them up into the other common code. Oh well...

doctoryes (approved):

tubular/google_api.py
pwnage101: That info is discoverable via logging from the google python API client.  It logs each outgoing sub-request and each sub-responses too.

pwnage101: > It's too bad that these response-gathering sections aren't quite the same enough to wrap them up into the other common code. Oh well...

At least it's better than before, in that there's now some common code at all!

tubular/google_api.py
```diff
@@ -58,16 +65,112 @@ class BaseApiClient(object):
...
+        # Retry on API throttling at the BATCH ITEM request level.
+        @backoff.on_exception(
+            backoff.expo,
+            TriggerRetryException,
+            max_time=600,  # 10 minutes
+            on_backoff=lambda details: _backoff_handler(details),  # pylint: disable=unnecessary-lambda
+        )
+        def func():
+            """
+            Core function which contitutes the retry loop.  It has no inputs or outputs, only side-effects which
```
bmedx: Nit: "constitutes"

bmedx (approved):
Minor nit, otherwise 👍

PR #271, opened 2018-08-16
bjh/learner_6167_new_message_types

inventhouse:
Added messages for stage and prod deployment failures; refactored to use fewer "magic strings", smaller API surface, and adding new message types is now just one line of code change.


macdiesel (approved):
Seems reasonable. Please squash before merging.

tubular/github_api.py
```diff
@@ -803,109 +816,35 @@ def _not_duplicate(pr_messages, new_message):
...
-        Returns:
-            github.IssueComment.IssueComment
-
-        """
-        return self.message_pull_request(
-            pr_number,
-            PR_E2E_FAILED_MESSAGE.format(extra_text=extra_text),
-            PR_E2E_FAILED_MESSAGE.format(extra_text=''),
-            force_message
+            PR_MESSAGE_FORMAT.format(
```
BenjiLee, line 842: FYI Pycharm warns... "expected str type, got unicode instead"

inventhouse: What is that referring to; which part specifically expects a string and is getting unicode?  Is that a Python 2 vs. 3 issue?  Once I set my PyCharm to Py3 (which is what the tox tests for this repo are set for) it stopped warning me about that.

tubular/tests/test_github.py
```diff
@@ -386,14 +386,22 @@ def test_message_pr_does_not_exist(self):
     def test_message_pr_deployed_stage(self):
         deploy_date = github_api.default_expected_release_date()
         with patch.object(self.api, 'message_pull_request') as mock:
-            self.api.message_pr_deployed_stage(1, deploy_date=deploy_date)
+            # self.api.message_pr_deployed_stage(1, deploy_date=deploy_date)
```
BenjiLee: delete unused code

BenjiLee (approved):

inventhouse:
Squash-merged as https://github.com/edx/tubular/commit/19c99397d793fd4c04274de135c5dd96867a4bf3

macdiesel:
FYI this repository requires both python 2 and 3 support. Please make sure this code runs under both environments.

inventhouse:
The tox tests appear to only test Python3 versions; should I open a ticket to add Py2 to the tests?

macdiesel:
@doctoryes this code executes under python 2 for the retirement scripts correct?

doctoryes:
Unfortunately, yes. Not all the tubular code but some subset of it runs under Python 2 on the build.testeng.edx.org Jenkins. Python 3 has since been installed on that machine - but we haven't done the conversion yet. We removed the Py2 testing back when we thought that tubular would only be used with Py3, as it is on the GoCD server/agents.

If you add Py2 to the tox testing, you'll probably discover several test failures.  But feel free to try it.

inventhouse:
@macdiesel @doctoryes What project should I file the "re-enable tubular Py2 testing" ticket under?

doctoryes:
@inventhouse You can file it under PLAT - thanks!

inventhouse:
Filed PLAT-2295 and confirmed that the message_prs_in_range script (including my changes here) at least runs in a Py2 environment.

PR #275, opened 2018-09-18
Add segment_api to support Segment retirement

bmedx:


tubular/scripts/check_migrate_duration.py
```diff
@@ -17,7 +17,7 @@
 # Add top-level module path to sys.path before importing tubular code.
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-from tubular.email import send_email  # pylint: disable=wrong-import-position
+from tubular.tubular_email import send_email  # pylint: disable=wrong-import-position
```
doctoryes, line 20: Is `email` conflicting with some other module?

tubular/scripts/helpers.py
```diff
@@ -133,15 +134,20 @@ def _setup_all_apis_or_exit(fail_func, fail_code, config):
...
-            if (state[2] == 'ECOMMERCE' and ecommerce_base_url is None) or \
-                    (state[2] == 'CREDENTIALS' and credentials_base_url is None):
-                fail_func(fail_code, 'Service URL is not configured, but required for state {}'.format(state))
+            for service, service_url in (
+                    ('ECOMMERCE', ecommerce_base_url),
+                    ('CREDENTIALS', credentials_base_url),
+                    ('SEGMENT', segment_base_url),
+            ):
+                if state[2] == service and service_url is None:
+                    fail_func(fail_code, 'Service URL is not configured, but required for state {}'.format(state))
```
doctoryes, line 150: The rule of three!

doctoryes: Nit: Maybe add in which service wasn't configured into the error message?

tubular/segment_api.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    def inner(func):  # pylint: disable=missing-docstring
+        func_with_backoff = backoff.on_exception(
+            backoff.expo,
+            requests.exceptions.HTTPError,
+            max_time=600,  # 10 minutes
+            giveup=_exception_not_internal_svr_error,
+            # Wrap the actual _backoff_handler so that we can patch the real one in unit tests.  Otherwise, the func
+            # will get decorated on import, embedding this handler as a python object reference, precluding our ability
+            # to patch it in tests.
```
doctoryes: Good comment.

tubular/segment_api.py
```diff
@@ -0,0 +1,153 @@
...
+mutation {{
+  createSourceRegulation(
+    workspaceSlug: "{}"
+    sourceSlug: "{}"
+    type: SUPPRESS_AND_DELETE
+    userId: "{}"
+  ) {{
+    id
+  }}
+}}"""
```
doctoryes, line 28: Is this GraphQL? If so, perhaps state as much for those of us who know nothing of it.

tubular/segment_api.py
```diff
@@ -0,0 +1,153 @@
...
+        # previous retirement step takes a while the token might time out.
+        access_token = self._get_auth_token()
+        headers = {"Authorization": "Bearer {}".format(access_token)}
+
+        for project in self.projects_to_retire:
+            for user_id in user_ids:
+                mutation = {
+                    'query': SUPPRESS_MUTATION.format(workspace_slug, project, user_id)
+                }
+                resp = requests.post(self.base_url, json=mutation, headers=headers)
```
doctoryes: Is there no way to batch-query even a portion of the possible IDs?

doctoryes (commented):
Questions and nits.

tubular/scripts/check_migrate_duration.py
bmedx: Yes, any time we've tried to add `requests` in to Tubular this conflict has popped up. Just got tired of working around it. Do you see any reason why any other repo would be using `tubular.email`?

tubular/scripts/helpers.py
bmedx: The state that's getting logged is the tuple that includes the service and call to be made, so it should be in there.

tubular/segment_api.py
bmedx: Not sure if your comment landed before or after my last changes, but there is a comment right above it now that calls it out as GraphQL.

tubular/scripts/helpers.py
doctoryes: Ah - 👍

tubular/scripts/check_migrate_duration.py
doctoryes: AFAIK it's only used inside of tubular scripts - not outside the repo.

tubular/segment_api.py
doctoryes: 👍

doctoryes (approved):

tubular/segment_api.py
```diff
@@ -0,0 +1,164 @@
...
+        # previous retirement step takes a while the token might time out.
+        access_token = self._get_auth_token()
+        headers = {"Authorization": "Bearer {}".format(access_token)}
+
+        for project in self.projects_to_retire:
+            for id_key in IDENTIFYING_KEYS:
+                mutation = {
+                    'query': SUPPRESS_MUTATION.format(self.workspace_slug, project, learner[id_key])
+                }
+                resp = requests.post(self.base_url, json=mutation, headers=headers)
```
pwnage101: Could this line be in its own function, then we can move the retry decorator there?  I normally wouldn't ask for a refactor based on a hunch, but in this case the change seems simple enough.

pwnage101: To clarify: i'm just worried that all the redundant calls that could result from retrying the entire function multiple times could blow something up or otherwise extend the stage duration some unreasonable amount.

AUTHORS
```diff
@@ -30,6 +30,8 @@ Mike Dikan <mike.dikan@gmail.com>
 Renzo Lucioni <renzo@lucioni.xyz>
 Simon Chen <schen@edx.org>
 Sofiya Semenova <sofiya@sofiya.io>
+Stu Young <estute@users.noreply.github.com>
```
pwnage101, line 33: This was added automatically, right?  Unfortunately, I don't think it has any legal weight :P

tubular/segment_api.py
bmedx: I gave some thought to that and had a fear that if their service is iffy we might have some extremely long running retirements. At is stands I think if errors were happening this would time out at 10 mins per learner, but making that change would bring it to a max of 10 mins per id-service or something like 6 hrs at the outside. I'm sure it wouldn't be that, but I do worry about long-running retirement steps. 

I'm honestly on the fence about it and happy to go either way. @macdiesel @doctoryes got any opinions?

pwnage101: If we coupled the proposed change with reducing the value of the `max_time` parameter to something lower, like 60 seconds, wouldn't that be the best of both worlds :P

bmedx: Ooooooor I could do that. :)

pwnage101 (approved):

PR #277, opened 2018-09-21
prevent service account from being mentioned in partner reporting notifications

pwnage101:
Tests passed locally under python 2.7

### DIFF
--- a/tubular/scripts/retirement_partner_report.py
+++ b/tubular/scripts/retirement_partner_report.py
@@ -245,12 +245,15 @@ def _add_comments_to_files(config, file_ids):
         for partner in file_ids
     }
 
-    # throw out all edx.org addresses, and flatten the permissions dicts to just the email:
+    # throw out all blacklisted addresses, and flatten the permissions dicts to just the email:
     external_emails = {
         partner: [
             perm['emailAddress']
             for perm in permissions[partner]
-            if not perm['emailAddress'].lower().endswith('@edx.org')
+            if not any(
+                perm['emailAddress'].lower().endswith(blacklisted_domain.lower())
+                for blacklisted_domain in config['blacklisted_notification_domains']
+            )
         ]
         for partner in permissions
     }
--- a/tubular/scripts/retirement_partner_report.py
+++ b/tubular/scripts/retirement_partner_report.py
@@ -260,9 +260,16 @@ def _add_comments_to_files(config, file_ids):
 
     file_ids_and_comments = []
     for partner in file_ids:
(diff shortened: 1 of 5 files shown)


tubular/scripts/retirement_partner_report.py
```diff
@@ -74,6 +74,12 @@
 Hello from edX. Dear {tags}, a new report listing the learners enrolled in your institution’s courses on edx.org that have requested deletion of their edX account and associated personal data within the last week has been published to Google Drive. Please access your folder to see the latest report. If you have any questions, please contact your edX Project Coordinator.
 """.strip()
 
+# List of domains for email addresses that should NOT be notified by the notification automation.
+BLACKLISTED_NOTIFICATION_DOMAINS = [
+    '@edx.org',
+    '@partner-reporting-automation.iam.gserviceaccount.com',
+]
+
```
doctoryes: Should this list be in the config that's passed into the script from a config file? Since the config is already passed into this method, it'd be relatively easy to do.

tubular/scripts/retirement_partner_report.py
```diff
@@ -245,12 +251,15 @@ def _add_comments_to_files(config, file_ids):
...
+    # throw out all blacklisted addresses, and flatten the permissions dicts to just the email:
     external_emails = {
         partner: [
             perm['emailAddress']
             for perm in permissions[partner]
-            if not perm['emailAddress'].lower().endswith('@edx.org')
+            if not any(
+                perm['emailAddress'].lower().endswith(blacklisted_domain)
+                for blacklisted_domain in BLACKLISTED_NOTIFICATION_DOMAINS
+            )
```
doctoryes: Should either/both of these two comparison values be lower-cased before comparison?

pwnage101: The blacklisted domain probably should be lowercased if we move the blacklist to config.

tubular/scripts/retirement_partner_report.py
pwnage101: config sounds like a good idea.

pwnage101: Come to think about it, the service account email address is already stored in the google credential JSON blob.  It's crossing my mind to parse that JSON, but it's schema isn't defined by us so I don't believe it would be robust.

pwnage101:
I addressed John's comment.  Additionally, I added a second commit to prevent comments from being added when there is nobody to notify.  Please re-review for this new additional commit!

doctoryes (approved):

bmedx (approved):

PR #278, opened 2018-09-28
Add logging & skip report upload upon no retired learners.

doctoryes:
These changes will make it more clear when no learners to report are returned from the LMS, as seen in the output of this script run:

https://build.testeng.edx.org/job/retirement-partner-reporter/39/consoleFull

Also, lots of throttling and retries are occurring on each run - so decrease the batch size to cut down on/eliminate (hopefully) the number of backoffs that occur.


tubular/google_api.py
```diff
@@ -30,7 +30,7 @@
 LOG = logging.getLogger(__name__)
 
 # Maximum number of requests per batch, according to the google API docs.
-GOOGLE_API_MAX_BATCH_SIZE = 100
+GOOGLE_API_MAX_BATCH_SIZE = 20
```
bmedx, line 34: Is this still according to the API docs, or does the comment need updating?

bmedx (approved):
One nit, otherwise lgtm

pwnage101 (approved):

tubular/google_api.py
doctoryes: I will update the comment.

doctoryes:
@pwnage101 @bmedx Any ideas on why I'm getting this weird test failure about boto?

macdiesel:
Seems like this may be related: https://github.com/travis-ci/travis-ci/issues/7940

doctoryes:
@macdiesel Yes - that was it!

PR #279, opened 2018-10-10
Update retirement to fetch Segment Ecom ID into config when configured

bmedx:


tubular/scripts/retire_one_learner.py
```diff
@@ -129,6 +129,22 @@ def _get_learner_and_state_index_or_exit(config, username):
...
+    here than to pass all of the config down to SegmentApi.
+    """
+    try:
+        response = config['ECOMMERCE'].get_tracking_key(learner)
+        return response['ecommerce_tracking_id']
+    except HttpNotFoundError:
+        LOG('Learner {} not found in Ecommerce. Setting Ecommerce Segment ID to None'.format(learner))
+        return None
+    except Exception as exc:  # pylint: disable=broad-except
+        FAIL_EXCEPTION(ERR_SETUP_FAILED, 'Unexpected error fetching Ecommerce tracking id!', text_type(exc))
```
doctoryes, line 144: Will the learner username be included in the exception here?

tubular/scripts/retire_one_learner.py
```diff
@@ -129,6 +129,22 @@ def _get_learner_and_state_index_or_exit(config, username):
         FAIL_EXCEPTION(ERR_SETUP_FAILED, 'Unexpected error fetching user state!', text_type(exc))
 
 
+def _get_ecom_segment_id(config, learner):
```
doctoryes, line 132: Is this method being covered in tests?

tubular/tests/test_segment_api.py
```diff
@@ -30,7 +30,7 @@ def test_suppress_and_delete_success():
     Test simple success case
     """
     projects_to_retire = ['project_1', 'project_2']
-    learner = {'id': 1, 'ecommerce_id': 'ecommerce-20', 'original_username': 'test_user'}
+    learner = {'id': 1, 'ecommerce_segment_id': 'ecommerce-20', 'original_username': 'test_user'}
```
doctoryes, line 33: So the `ecommerce_id` wasn't being referenced anywhere? Only in `IDENTIFYING_KEYS`?

tubular/scripts/retire_one_learner.py
bmedx: Probably not, but at this point in the Jenkins process the task itself should have the username of the retiring learner.

tubular/tests/test_segment_api.py
bmedx: Correct, that is it's sole purpose

tubular/scripts/retire_one_learner.py
bmedx: Given that the API call itself will have to be mocked out there's precious little to test, but I can exercise the exception paths. I'll write up some tests for that.

doctoryes: Feel free to make the call not to do so if you feel there's little value in it.

tubular/scripts/retire_one_learner.py
```diff
@@ -129,6 +129,22 @@ def _get_learner_and_state_index_or_exit(config, username):
...
+    """
+    Calls Ecommerce to get the ecom-specific Segment tracking id that we need to retire.
+    This is only available from Ecommerce, unfortunately, and makes more sense to handle
+    here than to pass all of the config down to SegmentApi.
+    """
+    try:
+        response = config['ECOMMERCE'].get_tracking_key(learner)
+        return response['ecommerce_tracking_id']
```
pwnage101: nit: how about just returning the tracking key rather than the entire response dict?

pwnage101 (approved):
see one nit.

PR #280, opened 2018-10-17
Add SailthruApi class and configure it in script helpers.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2364

Call the Sailthru API to delete a user by email address.

### DIFF
--- a/tubular/sailthru_api.py
+++ b/tubular/sailthru_api.py
@@ -1,0 +1,48 @@
+"""
+Sailthru API classes that will call the Sailthru REST API using the Sailthru client.
+"""
+import logging
+from six import text_type
+
+from sailthru.sailthru_client import SailthruClient
+from sailthru.sailthru_error import SailthruClientError
+
+log = logging.getLogger(__name__)
+
+
+class SailthruApi(object):
+    """
+    Sailthru API client used to make all Sailthru calls.
+    """
+    def __init__(self, sailthru_key, sailthru_secret):
+        """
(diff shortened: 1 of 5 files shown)


tubular/sailthru_api.py
```diff
@@ -0,0 +1,44 @@
...
+        Delete a user from Sailthru using their email address.
+        """
+        email = learner.get('original_email', None)
+        if not email:
+            raise TypeError('Expected an email address for user to delete, but received None.')
+
+        try:
+            sailthru_response = self._sailthru_client.api_delete("user", {'id': email})
+        except SailthruClientError as exc:
+            error_msg = "Exception attempting to delete user {} from Sailthru - {}".format(email, text_type(exc))
```
pwnage101: Is there any reason to expect non-ascii from these API error messages?  if so, just a reminder that python 2 scripts in jenkins break unless we respect the limitations of that environment (only encoded utf-8 bytes, or ascii-only strings going to stdout): https://github.com/edx/tubular/blob/master/tubular/google_api.py#L6-L10

pwnage101 (approved):
LGTM, just one thought.

bmedx (approved):

tubular/sailthru_api.py
doctoryes: Good point - I'll add this to the log messages.

pwnage101:
:+1:

PR #282, opened 2018-10-19
Update retirement to fetch Segment Ecom ID into config when configured

bmedx:
This is a retry of https://github.com/edx/tubular/pull/279 which was reverted due to failing tests on master. The only differences are new mocks on the tests that were failing and making this call explicitly configurable for better open source compatibility / reduced chances of new tests accidentally not mocking this call.


doctoryes (approved):
👍

pwnage101 (approved):

PR #283, opened 2018-10-22
Add script to send messages to slack (OPS-3437)

nadeemshahzad:


tubular/slack.py
```diff
@@ -0,0 +1,49 @@
...
+def submit_slack_message(auth_token, channels, message):
+    """
+    Post a message to one or more slack channels.
+
+    Args:
+        auth_token (str): Slack authentication token which authorizes message posting.
+        channels (list(str)): List of channel names to which to post the message.
+        message (str): Message to post to Slack channel.
+    """
+    post_url = SLACK_API_URL + NOTIFICATION_POST
```
jdmulloy:
```suggestion
    post_url = "{}{}".format(SLACK_API_URL, NOTIFICATION_POST)
```

jdmulloy (dismissed):

tubular/slack.py
jdmulloy: Using the format string instead of plus concatenation is preferred in modern python for clarity and knowing that the result is a string.

nadeemshahzad:
Thanks @jdmulloy , I have made the change.

coryleeio (approved):

coryleeio:
I can see that you updated the thing @jdmulloy requested so I'm going to dismiss his review so you can merge whenever.

PR #284, opened 2018-10-24
Add retirement archive and cleanup script

bmedx:
PLAT-2238 - the final step of retirement.

### DIFF
--- a/tubular/edx_api.py
+++ b/tubular/edx_api.py
@@ -159,6 +159,10 @@ class LmsApi(BaseApiClient):
         Retrieves a list of learners in the given retirement state that were
         created in the retirement queue between the dates given. Date range
         is inclusive, so to get one day you would set both dates to that day.
+
+        :param state_to_request: String LMS UserRetirementState state name (ex. COMPLETE)
+        :param start_date: Date or Datetime object
+        :param end_date: Date or Datetime
         """
         params = {
             'start_date': start_date.strftime('%Y-%m-%d'),
@@ -291,6 +295,15 @@ class LmsApi(BaseApiClient):
         params = {'data': usernames}
         with correct_exception():
             return self._client.api.user.v1.accounts.retirement_partner_report_cleanup.post(**params)
+
+    @_retry_lms_api()
+    def bulk_cleanup_retirements(self, usernames):
+        """
+        Deletes the retirements for all given usernames
+        """
+        params = {'data': {'usernames': usernames}}
+        with correct_exception():
+            return self._client.api.user.v1.accounts.retirement_cleanup.post(**params)
 
 
 class EcommerceApi(BaseApiClient):
--- a/tubular/scripts/retirement_archive_and_cleanup.py
+++ b/tubular/scripts/retirement_archive_and_cleanup.py
@@ -1,0 +1,233 @@
+#! /usr/bin/env python3
+"""
+Command-line script to bulk archive and cleanup retired learners from LMS
+"""
+from __future__ import absolute_import, unicode_literals
+
+from functools import partial
+from os import path
+import datetime
+import gzip
+import json
+import logging
+import sys
+
+import backoff
+import click
+from boto.s3.connection import S3Connection
+from boto.s3.key import Key
+from boto.exception import BotoClientError, BotoServerError
+from six import text_type
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
+
+# pylint: disable=wrong-import-position
+from tubular.scripts.helpers import (
+    _config_or_exit,
+    _fail,
+    _fail_exception,
+    _log,
+    _setup_lms_api_or_exit
+)
+
+
+SCRIPT_SHORTNAME = 'Archive and Cleanup'
+
+# Return codes for various fail cases
+ERR_NO_CONFIG = -1
+ERR_BAD_CONFIG = -2
+ERR_FETCHING = -3
+ERR_ARCHIVING = -4
+ERR_DELETING = -5
+ERR_SETUP_FAILED = -5
+
+LOG = partial(_log, SCRIPT_SHORTNAME)
+FAIL = partial(_fail, SCRIPT_SHORTNAME)
+FAIL_EXCEPTION = partial(_fail_exception, SCRIPT_SHORTNAME)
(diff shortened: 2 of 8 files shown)


tubular/edx_api.py
```diff
@@ -292,6 +292,18 @@ def retirement_partner_cleanup(self, usernames):
...
 
+    @_retry_lms_api()
+    def bulk_cleanup_retirements(self, usernames):
+        """
+        Deletes the retirements for all given usernames
+        """
+        params = {'data': {'usernames': usernames}}
+        try:
+            with correct_exception():
+                return self._client.api.user.v1.accounts.retirement_cleanup.post(**params)
```
bmedx: Currently on the LMS side this is a `DELETE` instead of a `POST`, there will be a PR over there to change that since I forgot Slumber doesn't handle `DELETE`s with JSON bodies.

doctoryes: The tryanny of Slumber!

tubular/edx_api.py
```diff
@@ -161,8 +161,8 @@ def get_learners_by_date_and_status(self, state_to_request, start_date, end_date
         is inclusive, so to get one day you would set both dates to that day.
         """
         params = {
-            'start_date': start_date.strftime('%Y-%m-%d'),
-            'end_date': end_date.strftime('%Y-%m-%d'),
+            'start_date': start_date,
+            'end_date': end_date,
```
doctoryes: Why? Does it default to using this format?

tubular/edx_api.py
```diff
@@ -292,6 +292,18 @@ def retirement_partner_cleanup(self, usernames):
...
+    def bulk_cleanup_retirements(self, usernames):
+        """
+        Deletes the retirements for all given usernames
+        """
+        params = {'data': {'usernames': usernames}}
+        try:
+            with correct_exception():
+                return self._client.api.user.v1.accounts.retirement_cleanup.post(**params)
+        except HttpNotFoundError:
+            return True
```
doctoryes: Does this Slumber error mean that one of the usernames sent over wasn't found?

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -0,0 +1,233 @@
...
+    LOG('Fetching users in state {} created from {} to {}'.format(initial_state, start_date, end_date))
+    try:
+        return config['LMS'].get_learners_by_date_and_status(initial_state, start_date, end_date)
+    except Exception as exc:  # pylint: disable=broad-except
+        FAIL_EXCEPTION(ERR_FETCHING, 'Unexpected error occurred fetching users to update!', exc)
+
+
+def _on_s3_backoff(details):
+    """
+    Callback that is called when backoff... backs off
```
doctoryes, line 68: Elementary! 🔙 📴

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -0,0 +1,233 @@
...
+        key = Key(bucket, datestr + filename)
+        key.set_contents_from_filename(filename)
+    except Exception as exc:
+        LOG(text_type(exc))
+        raise
+
+
+def _format_datetime_for_athena(timestamp):
+    """
+    Takes a JSON serialized timestamp string and returns a format of it that is queriable as a datetime in Athena
```
doctoryes: Nit: `queriable` => `queryable` - though it seems controversial.

doctoryes (commented):
LGTM - just a few questions.

tubular/edx_api.py
bmedx: I'd thought it would simplify matters to format this in the script, but this call is actually being used in `retirement_bulk_status_update.py` so I'll change it back and update the script and comments accordingly.

tubular/edx_api.py
bmedx: Bah is a testing artifact this predates the call merging to platform, I'll remove it.

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -0,0 +1,233 @@
...
+FAIL = partial(_fail, SCRIPT_SHORTNAME)
+FAIL_EXCEPTION = partial(_fail_exception, SCRIPT_SHORTNAME)
+CONFIG_OR_EXIT = partial(_config_or_exit, FAIL_EXCEPTION, ERR_BAD_CONFIG)
+SETUP_LMS_OR_EXIT = partial(_setup_lms_api_or_exit, FAIL, ERR_SETUP_FAILED)
+
+
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+
+
+def _fetch_learners_to_update_or_exit(config, start_date, end_date, initial_state):
```
pwnage101: Nit: `_fetch_learners_to_cleanup_or_exit` maybe?  Just trying to think of something more specific than "update".

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -0,0 +1,233 @@
...
+        learners = _fetch_learners_to_update_or_exit(config, start_date, end_date, 'COMPLETE')
+
+        if learners:
+            _archive_retirements_or_exit(config, learners)
+            _cleanup_retirements_or_exit(config, learners)
+            LOG('Archive and delete complete')
+        else:
+            LOG('No learners found!')
+    except Exception as exc:
+        print(text_type(exc))
```
pwnage101: should this print be a LOG?

tubular/edx_api.py
```diff
@@ -292,6 +296,15 @@ def retirement_partner_cleanup(self, usernames):
         with correct_exception():
             return self._client.api.user.v1.accounts.retirement_partner_report_cleanup.post(**params)
 
+    @_retry_lms_api()
+    def bulk_cleanup_retirements(self, usernames):
+        """
+        Deletes the retirements for all given usernames
```
pwnage101, line 302: If this fails for any reason, or runs out of retries, is there any way to know from the logging which users were already archived and which weren't?  Can we double-archive users without consequence?

bmedx: By design it should be fine to double-archive learners. They may show up in Athena queries more than once, but that's fixable in the queries if need be. If we really need them DevOps can retrieve the archive file from S3.

doctoryes (approved):

PR #285, opened 2018-10-30
Segment API key fix and handling of missing projects.

bmedx:


tubular/segment_api.py
```diff
@@ -149,13 +149,21 @@ def suppress_and_delete(self, learner):
...
-                    supression_id = resp_json['data']['createWorkspaceRegulation']['id']
+                    supression_id = resp_json['data']['createSourceRegulation']['id']
                     print('Suppress and delete queued. Id: {}'.format(supression_id))
                 except (TypeError, KeyError):
                     # This message means the identifier has already been submitted
                     # for this project, we count this as success.
-                    if 'Regulation already exists' in resp_json['errors'][0]['message']:
-                        LOG.info(resp_json['errors'][0]['message'])
+                    if 'errors' in resp_json:
+                        if 'Regulation already exists' in resp_json['errors'][0]['message']:
```
doctoryes: Is there ever any need to check `resp_json['errors']` for error messages beyond the zero-th index? Should the code be iterating over all possible error messages?

tubular/segment_api.py
```diff
@@ -149,13 +149,21 @@ def suppress_and_delete(self, learner):
                 resp_json = resp.json()
 
                 try:
-                    supression_id = resp_json['data']['createWorkspaceRegulation']['id']
+                    supression_id = resp_json['data']['createSourceRegulation']['id']
```
doctoryes, line 152: I'm assuming this name change is correct because 🤷‍♀️

doctoryes (commented):
One question - otherwise LGTM.

tubular/segment_api.py
bmedx: It was a change that needed to be made due to moving from retiring an identifier from all of the edx workspace to just the current project. It wasn't in their initial documentation, alas, but I was able to dump the response to see it. It does work, though!

tubular/segment_api.py
bmedx: I looked at that, but haven't seen any indication that they actually use more than one. I think if more than one is returned we should probably bubble it up, even if the first is one of these (since we've seen that they're mutually exclusive in testing). I can add that.

pwnage101 (approved):
Thinking about the unicode limitations of jenkins again, is there any chance that the raw error messages would contain non-ascii?  Maybe we only ever deal with numeric user IDs?  I assume the project names are also all simple ascii?

otherwise, LGTM :+1:

doctoryes (approved):

PR #287, opened 2018-11-05
sailthru deletion: swallow "not found" errors

pwnage101:
We should not assume that every user on the platform has a corresponding
sailthru profile, especially now that we correctly respect email opt-in
checkboxes everywhere.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,7 +64,7 @@ test =
     mock==2.0.0
     pep8==1.7.0
     pylint==1.7.1
-    pytest==3.1.3
+    pytest==3.6.3
     pytest-xdist==1.20.0
     pytest-pylint==0.9.0
     pytest-pep8
--- a/tubular/sailthru_api.py
+++ b/tubular/sailthru_api.py
@@ -33,8 +33,8 @@ class SailthruApi(object):
         try:
             sailthru_response = self._sailthru_client.api_delete("user", {'id': email})
         except SailthruClientError as exc:
-            error_msg = u"Exception attempting to delete user {} from Sailthru - {}".format(
-                email, text_type(exc)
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -64,7 +64,7 @@ test =
     mock==2.0.0
     pep8==1.7.0
     pylint==1.7.1
-    pytest==3.1.3
+    pytest==3.6.3
```
pwnage101, line 67: This is for the caplog pytest fixture.

tubular/sailthru_api.py
```diff
@@ -39,10 +41,16 @@ def delete_user(self, learner):
...
-            ).encode('utf-8')
-            log.error(error_msg)
-            raise Exception(error_msg)
+            if SAILTHRU_ERROR_NOT_FOUND in error.get_message():
+                skip_msg = u"No action taken because no profile was found - {}".format(
+                    error.get_message()
+                ).encode('utf-8')
+                log.info(skip_msg)
+            else:
+                error_msg = u"Error attempting to delete user {} from Sailthru - {}".format(
```
bmedx: Do we want to be logging an email address here?

bmedx (commented):
Just the one question, otherwise LGTM

tubular/sailthru_api.py
pwnage101: The email address is already present in the contents of error.get_message(), the contents of which is not really predictable, so if we needed to redact that it would need a pattern/regex match with questionable robustness at best.

bmedx: We could at least not dump it in the vast majority "not found" case where we already know what the error was and not explicitly dump it in the other case, since we'll already have the username for context if the Sailthru error doesn't include it.

bmedx (approved):
👍

PR #291, opened 2018-11-19
added logic to send notifications to slack

syed-awais-ali:


coryleeio (approved):
Seems fine, but please make sure to update
edx-gomatic/edxpipelines/patterns/tasks/common.py
def generate_find_and_advance_release

and any place that calls it as well, you'll need to merge them both around the same time

(I only mention it because I did not see the corresponding gomatic PR)

coryleeio:
Oops - I found it, sorry, =]

PR #310, opened 2018-12-12
Lessen count by the pruner

coryleeio:


ormsbee:
@coryleeio: Could we actually change how it's invoked by us, and not the default value? Many more people run the LMS in the community than run course-discovery, so this probably isn't relevant to most of them. And I'd rather not change default values and make scripts do unexpected things if at all possible.

coryleeio:
That is fair -  I will change it in the dsl

coryleeio:
https://github.com/edx/jenkins-job-dsl/pull/561

PR #311, opened 2018-12-13
Segment: Add query all single-user deletes. Remove projects.

doctoryes:
The single-user query for the status of all suppress/deletes was implemented as part of debugging the bulk delete failures. So let's keep it? Maybe?

Also, Segment projects are no longer needed in the bulk user delete API - so remove the support for them.

### DIFF
--- a/tubular/scripts/query_all_segment_deletion_requests.py
+++ b/tubular/scripts/query_all_segment_deletion_requests.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python3
 """
-Command-line script to check status of a bulk delete users request from Segment.
+Command-line script to query all deletion requests made to Segment to check their status.
 """
 from __future__ import absolute_import, unicode_literals
 
@@ -23,7 +23,7 @@ from tubular.scripts.helpers import (
     _log
 )
 
-DEFAULT_CHUNK_SIZE = 5000
+DEFAULT_STATUSES_PER_PAGE = 50
 
 # Return codes for various fail cases
 ERR_NO_CONFIG = -1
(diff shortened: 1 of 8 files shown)


tubular/scripts/query_all_segment_deletion_requests.py
```diff
@@ -0,0 +1,81 @@
...
+
+@click.command()
+@click.option(
+    '--config_file',
+    help='YAML file that contains retirement-related configuration for this environment.'
+)
+@click.option(
+    '--limit',
+    default=DEFAULT_DELETION_LIMIT,
+    help='Maximum number of Segment deletions to perform in each deletion request.'
```
bmedx: I think this help text is out of date

tubular/segment_api.py
```diff
@@ -205,3 +224,17 @@ def get_bulk_delete_status(self, bulk_delete_id):
...
         LOG.info(text_type(resp_json))
+
+    def get_all_deletion_requests(self, request_limit):
+        """
+        Queries the status of all previously submitted deletion requests.
+
+        :param request_limit: Number of requests to return with each GraphQL query.
+        """
+        query = {
+            'query': DELETION_REQUEST_LIST_QUERY.format(self.workspace_slug, request_limit)
```
bmedx: This looks like it's not handling pagination, is that right? Is it worth the effort to make that work right now?

tubular/scripts/query_all_segment_deletion_requests.py
doctoryes: Will fix.

tubular/segment_api.py
doctoryes: How about I add a `TODO` and just leave it as-is? If we ever need to use this function, we'll add it then.

bmedx (approved):

PR #313, opened 2018-12-17
Differentiate between required/optional segment user ID keys.

doctoryes:
If a key is required, raise when not present. If a key is optional, just continue if not present.

Handles the case of environments without ECommerce support.


bmedx (approved):
Once the issue that's failing tests is fixed 👍

PR #315, opened 2019-01-05
Add no-elb mode to retrieve_base_ami.py OPS-3639

jdmulloy:
Add --no-elb flag to retrieve_base_ami.py to make it work with backend services that don't have an ELB.


jdmulloy:
I can't quite make up my mind on this one. I have 3 commit with 3 different options.

1. Add --no-elb flag for backend services, which requires wiring changes in gomatic.
2. Try the old way and if no instances are behind ELBs try finding active ASGs by checking for disabled procedures.
3. Remove check of ELBs entirely, determine active ASG(s) by checking for disabled procedures only.

Options 3 is the easiest and most consistent but does mean that the way this works will change for all pipelines. I think it should be fine.

feanil (approved):

PR #317, opened 2019-01-10
Use specified region when connecting to S3 bucket.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2380

I had created a `us-east-2` bucket to test with - this change fixed the 400 errors I got.

After this change, stage-edx was able to successfully archive!:
https://build.testeng.edx.org/job/user-retirement-archiver/4/console

### DIFF
--- a/tubular/scripts/retirement_archive_and_cleanup.py
+++ b/tubular/scripts/retirement_archive_and_cleanup.py
@@ -67,7 +67,7 @@ def _on_s3_backoff(details):
     """
     Callback that is called when backoff... backs off
     """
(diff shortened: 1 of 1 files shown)


bmedx (approved):
I don't know if it matters here, but it's important to note that the S3 bucket needs to be in the same region as the Athena server.

doctoryes:
@bmedx I assume that'll be handled by the one who configures the archiving? BTW - the config section looks just like the test code one:
https://github.com/edx/tubular/blob/master/tubular/tests/retirement_helpers.py#L61-L66

PR #321, opened 2019-01-25
Added method to call the edx_proctoring endpoint for deleting proctoring data

davestgermain:
This depends on [edx_proctoring/517](https://github.com/edx/edx-proctoring/pull/517) landing in production.

### DIFF
--- a/tubular/edx_api.py
+++ b/tubular/edx_api.py
@@ -297,6 +297,14 @@ class LmsApi(BaseApiClient):
             return self._client.api.user.v1.accounts.retirement_partner_report_cleanup.post(**params)
 
     @_retry_lms_api()
+    def retirement_retire_proctoring_data(self, learner):
+        """
+        Removes the given learner from 3rd party proctoring backends
+        """
+        with correct_exception():
+            return self._client.api.edx_proctoring.v1.retire_backend_user(learner['user_id']).post()
+
+    @_retry_lms_api()
     def bulk_cleanup_retirements(self, usernames):
         """
         Deletes the retirements for all given usernames


bmedx:
Hey Dave, couple questions on this:

We're really trying to avoid putting these 3rd party network calls into the LMS as when we run these in batches it can really trash performance. Is there a reason you went in that direction? 

Looking at the Proctoring PR it seems like we attempt to retire from all of the backends in one call (and maybe eat the errors?). Can you guarantee that the backend provider calls are idempotent in case we need to re-run a user if one backend fails?

How would the Tubular script detect a failure of one or more backends?

davestgermain:
> We're really trying to avoid putting these 3rd party network calls into the LMS as when we run these in batches it can really trash performance. Is there a reason you went in that direction?

I think making an API request will simplify the complexity, because the API with the proctoring providers is already well defined in edx-proctoring. Any changes would be confined to edx-proctoring. Another minor point is that user_id sent to the backend is a particular sha1 hash; that code lives in edx-proctoring. 

Also, enabling the deletion API in edx-proctoring is beneficial for open edx installations which don't use this repo for management tasks. They can implement their own batch process. 

> Looking at the Proctoring PR it seems like we attempt to retire from all of the backends in one call (and maybe eat the errors?). Can you guarantee that the backend provider calls are idempotent in case we need to re-run a user if one backend fails?

Although the code loops through "all" backends, there is only one provider that will support this API -- proctortrack. It's likely that it'll remain the only provider that supports this method for a while. 

Since Proctortrack will return a response asynchronously to the actual deletion, I am unsure what will happen if you request a second deletion before the first deletion has completed. They haven't implemented their end of the API yet.

> How would the Tubular script detect a failure of one or more backends?

It doesn't... I couldn't see a pattern for handling failures in this code. Should the method raise an exception? The return result from the API request is always `True` (I assume it's returning whether the response returned a 200 status. That's not awesome). 

How do you recommend the API return errors?  

To be clear, for the foreseeable future, the only possible responses to the API will be either `{"proctortrack": True}` or `{"proctortrack": False}`

bmedx:
I think it makes sense to try to put it in LMS if that's the case, with the caveat that if it ends up being a performance issue it will need to get moved out. 2/3 of the times we've done this it's ended up being an issue.

It would be nice if we could pass in the backend with retirement steps for each. That would give us some granularity if one service or another went down, though I know that's not an issue at the moment. Someday it will be and it could cause a lot of pain if one service goes down and the other isn't idempotent (for example).

Ideally for the Proctortrack API it would be great if the following could be "success":
- User doesn't exist on their end (or has already been deleted)
- User is already in the deletion queue
- User is added to the deletion queue

Then from the LMS API we would want the 204 HTTP status on success and a 500 on any kind of failure. When 500s happen whoever is on call for Platform gets paged and can manually reset the user to retry, or if lots of people are failing (ie. service is down, credentials expired, some queue on Proctortrack's side is full, etc) we can stop the pipeline or remove that retirement step. Alternatively you could raise an error here if `{"proctortrack": False}`, but having a call for each service + using HTTP codes is how we've been doing things.

davestgermain:
[edx-platform/19690](https://github.com/edx/edx-platform/pull/19690) will allow edx-proctoring to use the correct permission for the retirement API

[edx-proctoring/524](https://github.com/edx/edx-proctoring/pull/524) changes the error response to a 500 (and won't make any backend request if the user has never taken a proctored exam.)

I'll work with Proctortrack to make sure that their endpoint is idempotent.

tubular/edx_api.py
```diff
@@ -296,6 +296,14 @@ def retirement_partner_cleanup(self, usernames):
         with correct_exception():
             return self._client.api.user.v1.accounts.retirement_partner_report_cleanup.post(**params)
 
+    @_retry_lms_api()
+    def retirement_retire_proctoring_data(self, learner):
+        """
+        Removes the given learner from 3rd party proctoring backends
+        """
+        with correct_exception():
+            return self._client.api.edx_proctoring.v1.retire_backend_user(learner['user_id']).delete()
```
bmedx: Just a reminder that if you change the call to a POST, you'll have to update here as well.

bmedx (approved):
Thanks for addressing the feedback!

PR #322, opened 2019-02-07
Jkantor/salesforce leads

jansenk:
Given a learner, check if that learner exists as a lead in Salesforce.
        If they do, create a Salesforce Task instructing someone to manually retire
        their personal information.

### DIFF
--- a/tubular/salesforce_api.py
+++ b/tubular/salesforce_api.py
@@ -1,0 +1,101 @@
+"""
+Salesforce API class that will call the Salesforce REST API using simple-salesforce.
+"""
+import logging
+from simple_salesforce import Salesforce
+
+log = logging.getLogger(__name__)
+RETIREMENT_TASK_DESCRIPTION = (
+    "A user data retirement request has been made for "
+    "{email} who has been identified as a lead in Salesforce. "
+    "Please manually retire the user data for this lead."
+)
+
+
+class SalesforceApi(object):
+    """
+    Class for making Salesforce API calls
+    """
+    def __init__(self, username, password, security_token, domain, assignee_username):
+        """
+        Create API with credentials
+        """
+        self._sf = Salesforce(
+            username=username,
+            password=password,
+            security_token=security_token,
+            domain=domain
+        )
+        self.assignee_id = self.get_user_id(assignee_username)
+        if not self.assignee_id:
+            raise Exception("Could not find Salesforce user with username " + assignee_username)
+
+    def get_lead_ids_by_email(self, email):
+        """
+        Given an id, query for a Lead with that email
+        Returns a list of ids tht have that email or None if none are found
+        """
+        query_string = "SELECT Id FROM Lead WHERE Email='{email}'"
+        id_query = self._sf.query(query_string.format(email=email))
+        total_size = int(id_query['totalSize'])
+        if total_size == 0:
+            return None
+        else:
+            ids = [record['Id'] for record in id_query['records']]
+            if len(ids) > 1:
+                log.warning("Multiple Ids returned for Lead with email {}".format(email))
+            return ids
+
+    def get_user_id(self, username):
+        """
+        Given a username, returns the user id for the User with that username
+        or None if no user is found
+        Used to get a the user id of the user we will assign the retirement task to
+        """
+        query_string = "SELECT Id FROM User WHERE Username='{username}'"
+        id_query = self._sf.query(query_string.format(username=username))
+        total_size = int(id_query['totalSize'])
+        if total_size == 0:
+            return None
+        else:
+            return id_query['records'][0]['Id']
+
+    def _create_retirement_task(self, email, lead_ids):
+        """
+        Creates a Salesforce Task instructing a user to manually retire the
+        given lead
+        """
+        task_params = {
+            'Description': RETIREMENT_TASK_DESCRIPTION.format(email=email),
+            'Subject': "GDPR Request: " + email,
+            'WhoId': lead_ids[0],
+            'OwnerId': self.assignee_id
+        }
+        if len(lead_ids) > 1:
+            note = "\nNotice: Multiple leads were identified with the same email. Please retire all following leads:"
+            for lead_id in lead_ids:
+                note += "\n{}".format(lead_id)
+            task_params['Description'] += note
+        created_task = self._sf.Task.create(task_params)
+        if created_task['success']:
+            log.info("Successfully salesforce task created task " + created_task['id'])
+        else:
+            log.error("Errors while creating task:")
+            for error in created_task['errors']:
+                log.error(error)
+            raise Exception("Unable to create retirement task for email " + email)
+
+    def retire_learner(self, learner):
+        """
+        Given a learner email, check if that learner exists as a lead in Salesforce
+        If they do, create a Salesforce Task instructing someone to manually retire
+        their personal information
+        """
+        email = learner.get('original_email', None)
+        if not email:
+            raise TypeError('Expected an email address for user to delete, but received None.')
+        lead_ids = self.get_lead_ids_by_email(email)
+        if lead_ids is None:
+            log.info("No action taken because no lead was found in Salesforce.")
+            return
+        self._create_retirement_task(email, lead_ids)
--- a/tubular/tests/test_salesforce.py
+++ b/tubular/tests/test_salesforce.py
@@ -1,0 +1,125 @@
+"""
+Tests for the Salesforce API functionality
+"""
+import logging
+from contextlib import contextmanager
+import mock
+import pytest
+from tubular import salesforce_api
+from simple_salesforce import SalesforceError
+
+
+@pytest.fixture
+def test_learner():
+    return {'original_email': 'foo@bar.com'}
+
+
+def make_api():
+    """
+    Helper function to create salesforce api object
+    """
+    return salesforce_api.SalesforceApi("user", "pass", "key", "domain", "user")
+
+
+@contextmanager
+def mock_get_user():
+    """
+    Context manager method to mock getting the assignee user id when the api object is created
+    """
+    with mock.patch('tubular.salesforce_api.SalesforceApi.get_user_id') as getuser:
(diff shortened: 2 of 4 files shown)


tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
+"""
+Salesforce API class that will call the Salesforce REST API using simple-salesforce.
+"""
+import logging
+from simple_salesforce import Salesforce
+
+log = logging.getLogger(__name__)
+RETIREMENT_TASK_DESCRIPTION = ("A user data retirement request has been made for"
```
iloveagent57: Nit: for multi-line statements like this, we usually newline and indent 4 spaces rather than doing it this way.  See https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#breaking-long-lines

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+                               " %s who has been identified as a lead in Salesforce."
```
iloveagent57: Nit: Prefer `string.format()` to string interpolation.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+                               " Please manually retire the user data for this lead.")
+
+
+class SalesforceApi():
```
iloveagent57: Nit: inherit from `object`.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Class for making Salesforce API calls
+    """
+    def __init__(self, username, password, security_token, domain, assignee_email):
+        """
+        Create API with credentials
+        """
+        self._sf = Salesforce(username=username,
```
iloveagent57: Nit: the line-breaking thing again.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+                              password=password,
+                              security_token=security_token,
+                              domain=domain)
+        self.assignee_id = self.get_user_id_by_email(assignee_email)
```
iloveagent57: It seems weird to me to require an assignee just to initialize this API client.  Would it be better to add this to `self.retire_learner()`?  I *think* you can define more than one argument to pass to that function in the config yaml file.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+        if not self.assignee_id:
+            raise Exception("Could not find Salesforce user with email %s" % assignee_email)
+
+    def _get_id_by_email(self, object_type, email):
+        """
+        Given an id, query for an <object_type> with that email
+        Returns the id of that object or None if not found
+        Raises an exception if multiple are found
+        """
+        id_query = self._sf.query("SELECT Id FROM %s WHERE Email='%s'" % (object_type, email))
```
iloveagent57: Nit: string.format()

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+        total_size = int(id_query['totalSize'])
+        if total_size == 0:
+            return None
+        elif total_size == 1:
+            return id_query['records'][0]['Id']
+        else:
+            raise Exception("Multiple results for email %s" % email)
```
iloveagent57: format

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+    def get_lead_id_by_email(self, email):
+        """
+        Given an id, query for an <object_type> with that email
+        Returns the id of that object or None if not found
+        Raises an exception if multiple are found
+        """
+        return self._get_id_by_email("Lead", email)
+
+    def get_user_id_by_email(self, email):
+        """
```
iloveagent57: Nit: maybe add something to the docstring indicating that this is to help get an internal user that the task will be assigned to.

tubular/tests/test_salesforce.py
```diff
@@ -0,0 +1,116 @@
+"""
```
iloveagent57, line 1: Great tests!

iloveagent57 (commented):
Looks great, overall, I just have a bunch of tiny complaints.  Good job!

tubular/salesforce_api.py
jansenk: I agree that would make more sense, but I didn't know that I could do that. I'll give it a shot.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
+"""
+Salesforce API class that will call the Salesforce REST API using simple-salesforce.
+"""
+import logging
+from simple_salesforce import Salesforce
+
+log = logging.getLogger(__name__)
+RETIREMENT_TASK_DESCRIPTION = ("A user data retirement request has been made for"
+                               " %s who has been identified as a lead in Salesforce."
+                               " Please manually retire the user data for this lead.")
```
bmedx: Are they actually manually doing these? We generate a lot. If we re-ran everyone in the retirement queue they would get hit with tens of thousands of these.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+            'WhoId': lead_id,
+            'OwnerId': self.assignee_id
+        }
+        created_task = self._sf.Task.create(task_params)
+        if created_task['success']:
+            log.info("Successfully salesforce task created task " + created_task['id'])
+        else:
+            log.error("Errors while creating task:")
+            for error in created_task['errors']:
+                log.error(error)
```
bmedx, line 85: We should raise an error here as we will need to manually reset and re-run the user if this doesn't succeed.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,91 @@
...
+        return self._get_id_by_email("Lead", email)
+
+    def get_user_id_by_email(self, email):
+        """
+        Given an email, returns the user id for the User with that email
+        or None if no user is found
+        """
+        return self._get_id_by_email("User", email)
+
+    def _create_retirement_task(self, email, lead_id):
```
bmedx: I'm not familiar with the API, but if there are errors that are safe to retry we should do that here, possibly using `backoff` as we do in the other API clients. It will cut down on the amount of manual intervention needed, which my team would appreciate a lot. :)

bmedx (commented):
Agreed on all of the previous nits and comments. Tests are really nice!

tubular/salesforce_api.py
schenedx: Should this be an exception case we should raise? Shouldn't we just log this a proceed with the first user?

schenedx (commented):
Just a minor question. Good work

tubular/salesforce_api.py
bmedx: @schenedx I would think that we would either raise this or delete them all. Deleting the first seems like it would not be sufficient to meet our obligations.

tubular/salesforce_api.py
schenedx: @bmedx Yes. The actual retirement action is a manual task. 
The way we scale this down to manageable manual task is to check whether the user exists on salesforce.
We only have thousands of learners on Salesforce so far. The number of user to retire from that pool of thousands should be pretty small.

bmedx: @schenedx is there any throttling we might run into on the Salesforce side?

tubular/salesforce_api.py
schenedx: @jansenk @bmedx Good point, I think we should delete them all. Maybe we will still generate one manual task in Salesforce, but provide the number of user to delete is greater than 1.

tubular/salesforce_api.py
schenedx: Good question, let me ask the Salesforce admin Darek.

schenedx: @bmedx The limit for Salesforce API is 121k per 24 hour period. Our current rate of consumption is about 12k per day. So we are only consuming 10% of the quota

tubular/salesforce_api.py
jansenk: Me and @schenedx  talked and we think that for now I think we might just keep it as is. We can always put in another ticket for that.

jansenk:
@schenedx @bmedx @iloveagent57  please review again

schenedx (approved):
👍

tubular/salesforce_api.py
```diff
@@ -53,27 +59,34 @@ def get_user_id_by_email(self, email):
...
         task_params = {
-            'Description': RETIREMENT_TASK_DESCRIPTION % email,
+            'Description': RETIREMENT_TASK_DESCRIPTION.format(email=email),
             'Subject': "GDPR Request: " + email,
-            'WhoId': lead_id,
+            'WhoId': lead_ids[0],
             'OwnerId': self.assignee_id
         }
+        if len(lead_ids) > 1:
+            note = "\nNotice: Multiple leads were identified with the same email:"
```
bmedx: Are we expecting them to retire those leads as well? If so, I think we should be explicit about that here.

bmedx (approved):
Thanks for the hard work and working through the feedback with me! 👍

bmedx:
One last suggestion while I'm thinking of it. Once we have all of the secrets, please try running the API locally from a Python shell to make sure everything works as expected before turning on the retirement step. We do this with all of the external APIs as an extra precaution since they can be particularly finicky.

tubular/salesforce_api.py
```diff
@@ -0,0 +1,101 @@
...
+Salesforce API class that will call the Salesforce REST API using simple-salesforce.
+"""
+import logging
+from simple_salesforce import Salesforce
+
+log = logging.getLogger(__name__)
+RETIREMENT_TASK_DESCRIPTION = (
+    "A user data retirement request has been made for "
+    "{email} who has been identified as a lead in Salesforce. "
+    "Please manually retire the user data for this lead."
```
pwnage101, line 11: Should this link to a runbook or additional instructions?

PR #326, opened 2019-03-15
Add username replacement script

tuchfarber:
Username replacement script allows usernames to be changes across all
services. Usernames will be sent to script via csv file from jenkins job

### DIFF
--- a/tubular/scripts/replace_usernames.py
+++ b/tubular/scripts/replace_usernames.py
@@ -1,0 +1,149 @@
+#! /usr/bin/env python3
+
+"""
+Command-line script to replace the usernames for all passed in learners.
+Accepts a list of current usernames and their preferred new username. This
+script will call LMS first which generates a unique username if the passed in
+new username is not unique. It then calls all other services to replace the
+username in their DBs.
+
+"""
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+from os import path
+import csv
+import io
+import sys
+import logging
+import click
+import yaml
+
+# Add top-level module path to sys.path before importing tubular code.
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
+
+from tubular.edx_api import CredentialsApi, DiscoveryApi, EcommerceApi, LmsApi  # pylint: disable=wrong-import-position
+
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+LOG = logging.getLogger(__name__)
+
+
+def write_responses(writer, replacements, status):
+    for replacement in replacements:
+        original_username = replacement.keys()[0]
+        new_username = replacement.values()[0]
+        writer.writerow([original_username, new_username, status])
+
+
+@click.command()
+@click.option(
+    '--config_file',
+    help='File in which YAML config exists that overrides all other params.'
+)
+@click.option(
+    '--username_replacement_csv',
+    help='File in which YAML config exists that overrides all other params.'
+)
+def replace_usernames(config_file, username_replacement_csv):
+    """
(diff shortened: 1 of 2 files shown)


feanil:
@tuchfarber going to defer to your team for reviewing this script.  My only assumptions are that the script runs and that you and your team agree that it does what you believe it's supposed to.  If you would like a deeper review from me, let me know.

tuchfarber:
Sounds good @feanil! I should have clarified that the review on this was just to show that all the connected pieces are in the right spots. More of an FYI.

tubular/scripts/replace_usernames.py
```diff
@@ -0,0 +1,144 @@
...
+    # The LMS list has already verified usernames and made any duplicate
+    # usernames unique (e.g. 'matt' => 'mattf56a'). We pass successful
+    # replacements onto the next service and store all failed replacments.
+    replacement_methods = [
+        ecommerce_api.replace_usernames,
+        discovery_api.replace_usernames,
+        credentials_api.replace_usernames,
+        lms_api.replace_forums_usernames,
+    ]
+    for replacement_method in replacement_methods:
```
bessiesteinberg, line 127: optional NIT: consider adding a comment explaining what is happening here. I understood it after looking at it for a while, but it could probably use a comment explaining that if any username replacements fail on one api they should not be propagated to the next api in an attempt to limit the negative effects of partial failures

tubular/scripts/replace_usernames.py
```diff
@@ -0,0 +1,144 @@
...
+    successful_replacements = in_progress_replacements
+
+    with open('username_replacement_results.csv', 'w', newline='') as output_file:
+        csv_writer = csv.writer(output_file)
+        # Write header
+        csv_writer.writerow(['Original Username', 'New Username', 'Status'])
+        write_responses(csv_writer, successful_replacements, "SUCCESS")
+        write_responses(csv_writer, partially_failed_replacements, "PARTIALLY FAILED")
+        write_responses(csv_writer, fully_failed_replacements, "FAILED")
+
```
bessiesteinberg, line 141: If there are any failures or partial failures maybe this should ```exit(-1)```?
To clarify I want to make sure failures and especially partial failures fail noticeably.  So it doesn't matter to me how it's done as long as it's easy to catch and alert on failures

bessiesteinberg (approved):
Looks good! Nice job!

tubular/scripts/replace_usernames.py
tuchfarber: Added a comment explaining what's happening here.

tubular/scripts/replace_usernames.py
tuchfarber: Good catch! This should definitely be a failed job if there are issues in replacements. Added below

bessiesteinberg:
both changes look good!

tubular/scripts/replace_usernames.py
```diff
@@ -0,0 +1,150 @@
...
+from os import path
+import csv
+import io
+import sys
+import logging
+import click
+import yaml
+
+# Add top-level module path to sys.path before importing tubular code.
+# sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
```
georgebabey: Is there a reason this commented line is left in?

georgebabey (approved):

tubular/scripts/replace_usernames.py
tuchfarber: Good catch! It was changed for testing because I was calling it from a different location and forgot to change it back.

PR #332, opened 2019-05-03
Added messages for jenkins job

syedimranhassan:


adzuci (approved):
LGTM, if we wanted to be super consistent, we could add periods at the end of the two new messages.

PR #333, opened 2019-05-14
Fix bugs in partner report cleanup script, increase debug output

bmedx:


tubular/google_api.py
```diff
@@ -349,7 +352,10 @@ def walk_files(self, top_folder_id, file_fields='id, name', mimetype=None, recur
 
         while folders_to_visit:
             current_folder = folders_to_visit.pop()
+            print("Current folder: {}".format(current_folder))
             visited_folders.append(current_folder)
+            extra_kwargs = {}
```
bmedx, line 357: This was the first bug, this wasn't being reset so the paging parameter was being sent when it wasn't expected.

tubular/google_api.py
```diff
@@ -349,7 +352,10 @@ def walk_files(self, top_folder_id, file_fields='id, name', mimetype=None, recur
 
         while folders_to_visit:
             current_folder = folders_to_visit.pop()
+            print("Current folder: {}".format(current_folder))
```
bmedx: Oops. I'll fix this to a `LOG.info`

tubular/scripts/delete_expired_partner_gdpr_reports.py
```diff
@@ -102,10 +102,11 @@ def delete_expired_reports(config_file, google_secrets_file, age_in_days):
     try:
         delete_before_dt = datetime.now(UTC) - timedelta(days=age_in_days)
         drive = DriveApi(config['google_secrets_file'])
+        LOG('DriveApi configured')
         drive.delete_files_older_than(
             config['drive_partners_folder'],
             delete_before_dt,
-            mimetype='application/csv',
+            mimetype='text/csv',
```
bmedx, line 109: This was the second bug, no files were being found with `application/csv` this change found what we're looking for.

doctoryes (approved):
:shipit:

PR #334, opened 2019-05-14
Make FE build and deploy multisite friendly

tuchfarber:
A mutlideployment friendly frontend will have a `MULTISITE` key in their config. Each site under that will have a `BUCKET_NAME` and a `HOSTNAME`. E.g.
```
APP_CONFIG:
    SOME_VARIABLE: "somevalue"
MULTISITE:
    - HOSTNAME: "www.example1.com"
      BUCKET_NAME: "bucket1.name"
    - HOSTNAME: "www.example2.com
      BUCKET_NAME: "bucket2.name"
```

If the application has a `MULTISITE` key, we deploy the app once for each site listed, passing the `HOSTNAME` into the build as a `SITENAME` environment variable and deploying to the `BUCKET_NAME` bucket.

### DIFF
--- a/scripts/frontend_multi_build.py
+++ b/scripts/frontend_multi_build.py
@@ -1,0 +1,1 @@
+../tubular/scripts/frontend_multi_build.py
--- a/scripts/frontend_multi_deploy.py
+++ b/scripts/frontend_multi_deploy.py
@@ -1,0 +1,1 @@
+../tubular/scripts/frontend_multi_deploy.py
--- a/tubular/scripts/frontend_multi_build.py
+++ b/tubular/scripts/frontend_multi_build.py
@@ -21,6 +21,7 @@ from tubular.scripts.frontend_utils import FrontendBuilder  # pylint: disable=wr
 SCRIPT_SHORTNAME = 'Build frontend'
 LOG = partial(_log, SCRIPT_SHORTNAME)
 FAIL = partial(_fail, SCRIPT_SHORTNAME)
+MULTISITE_PATH = './multisite/dist'
 
 
 @click.command()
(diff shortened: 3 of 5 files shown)


douglashall:
What if you only want to deploy changes to a single site for some reason? Shouldn't we do this at the gocd level?

tuchfarber:
We looked at creating a new pipeline for each deployment, but with a possibility of 300+ deployments, we didn't want to flood GoCD with pipelines. This is our solution for while we have a few deployments to manage. Once the number increases, we've talked about moving towards a model where:
(edx-internal MULTISITE is no more)
1. Pipeline gets triggered by code push or config change
2. `npm run build` requests all sites that need to be updated from an API and builds them
3. Each sites folder would have a special file (say `.deploy.yml` dropped into it with which sites to deploy)
4. frontend_deploy would read the special file and only deploy those sites. 

So the future goal is to move towards a model where we don't need a PR to enable new deployments. This feels like a good middle ground between where we are and where we want to go.

douglashall (changes requested):
I'd like to see the multisite build/deploys a little more divorced from the standard build/deploys. Wondering if you should create separate scripts here for multisite and use those upstream. That said there is probably some existing code here that could be refactored to make it reusable in both sets of scripts.

tuchfarber:
Fair. I'll see what I can do!

tuchfarber:
Refactored to have two scripts that use the same utility classes to build and deploy. Related gomatic PR to call the correct script: https://github.com/edx/edx-gomatic/pull/527

douglashall (approved):
Love it! Thanks.

Note that the convention in this repo is to put symbolic links in the top-level `scripts` directory that point to the files in `tubular/scripts`. I'm forgetting the reason for this at the moment.

PR #338, opened 2019-05-16
Extend backoff retries to Segment auth calls

bmedx:


tubular/segment_api.py
```diff
@@ -144,6 +144,7 @@ def _get_auth_token(self):
         resp_json = None
         try:
             resp = requests.post(self.base_url, json=mutation)
+            resp.raise_for_status()
```
doctoryes, line 147: I don't understand why this method call was added?

tubular/segment_api.py
```diff
@@ -144,6 +144,7 @@ def _get_auth_token(self):
...
             resp_json = resp.json()
             return resp_json['data']['login']['access_token']
         except (TypeError, KeyError, JSONDecodeError):
```
pwnage101: Should this now also catch and log HTTPError?

tubular/segment_api.py
doctoryes: Oh - nm - I see that this is a `requests` thing.

tubular/segment_api.py
doctoryes: Agreed - from `requests` doc:
Response.raise_for_status() will raise an HTTPError if the HTTP request returned an unsuccessful status code.

tubular/tests/test_segment_api.py
```diff
@@ -114,26 +134,24 @@ def test_bulk_delete_error(setup_bulk_delete, caplog):  # pylint: disable=redefi
...
+                'fake_base_url', 'fake_email', 'fake_password', 'fake_workspace'
+            ]]
+        )
 
-                learner = TEST_SEGMENT_CONFIG['learner']
+        learner = TEST_SEGMENT_CONFIG['learner']
 
-                with pytest.raises(JSONDecodeError):
-                    segment.delete_learners(learner, 1000)
+        with pytest.raises(JSONDecodeError):
```
pwnage101: How come this works for FakeAuthErrorResponse?  Isn't that one supposed to raise HTTPError due to the `resp.raise_for_status()`?

bmedx: Updated to make the fakery more realistic

pwnage101 (approved):
Approved, but you can't fool me---definitely fake.

PR #342, opened 2019-06-18
Report commit statuses by using both github statuses and check-suite …

cpennington:
…results

### DIFF
--- a/tubular/github_api.py
+++ b/tubular/github_api.py
@@ -307,6 +307,77 @@ class GitHubAPI(object):
 
         return commit.get_combined_status()
 
+    def get_commit_check_suites(self, commit):
+        """
+        Calls GitHub's `<commit>/check-suites` endpoint for a given commit. See
+        https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-specific-ref
+
+        Arguments:
+            commit: One of:
+                - string (interprets as git SHA and fetches commit)
+                - GitCommit (uses the accompanying git SHA and fetches commit)
+                - Commit (directly gets the combined status)
+
+        Returns:
+            Json representing the check suites
+        """
+        if isinstance(commit, six.string_types):
+            commit = self.github_repo.get_commit(commit)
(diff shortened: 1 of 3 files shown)


tubular/github_api.py
```diff
@@ -307,6 +308,97 @@ def get_commit_combined_statuses(self, commit):
...
+        Returns:
+            Json representing the check suites
+        """
+        if isinstance(commit, six.string_types):
+            commit = self.github_repo.get_commit(commit)
+        elif isinstance(commit, GitCommit):
+            commit = self.github_repo.get_commit(commit.sha)
+        elif not isinstance(commit, Commit):
+            raise UnknownObjectException(
+                500, "commit is neither a valid sha nor github.Commit.Commit object."
```
crice100, line 330: Missing an "or" (GitCommit or Commit)?

tubular/github_api.py
```diff
@@ -307,6 +308,97 @@ def get_commit_combined_statuses(self, commit):
...
+            suite['app']['name']: (
+                suite.get('conclusion', 'pending').lower(),
+                suite['url']
+            )
+            for suite in check_suites['check_suites']
+        })
+
+        return results
+
+    def filter_validation_results(self, results, exclude_apps='', include_apps=''):
```
crice100: Where is this called? Should we add a test for the include/exclude?

tubular/github_api.py
cpennington: I'm not sure what you mean. Where you would put an `or` in that sentence?

crice100: Minor (possibly irrelevant) point: I read the doc to say "commit could be 1 of these 3 things: _string_, _GitCommit_ or _Commit_". But the exception message talks about "_github.Commit.Commit_", which looks to be a typo.

tubular/github_api.py
cpennington: I'll delete it. It's not currently used.

crice100 (approved):

PR #353, opened 2019-07-10
Fix issues with non-string exception data

bmedx:


tubular/scripts/helpers.py
```diff
@@ -55,16 +55,27 @@ def _fail_exception(kind, code, message, exc):
     """
     A version of fail that takes an exception to be utf-8 decoded
     """
+    exc_msg = _get_error_str_from_exception(exc)
+    message += '\n' + exc_msg
```
pwnage101, line 59: Isn't `\n` double-prepended at this point?

doctoryes (approved):

tubular/scripts/helpers.py
bmedx: Good, it will make it very obvious where the real error we care about is. ;)

pwnage101 (approved):

PR #368, opened 2019-09-09
Update Segment API to REST from GraphQL

bmedx:
Must merge with https://github.com/edx-ops/user-retirement-secure/pull/61


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{34,35,36}
+envlist = py{35,36}
```
bmedx, line 2: 3.4 is not supported any more, and is not tested on Travis so I've removed it.

tubular/tests/test_segment_api.py
```diff
@@ -145,31 +110,6 @@ def test_bulk_delete_error(setup_bulk_delete, caplog):  # pylint: disable=redefi
     with pytest.raises(Exception):
         segment.delete_learners(learner, 1000)
 
-    assert mock_post.call_count == 1
+    assert mock_post.call_count == 4
     assert "Error was encountered for learners between start/end indices (0, 0)" in caplog.text
     assert "{'error': 'Test error message'}" in caplog.text
-
-
-def test_auth_error(caplog, fake_auth_data_responses):  # pylint: disable=redefined-outer-name
```
bmedx, line 153: All of these auth tests and their supporting functions are no longer necessary as the API uses a fixed auth token.

tubular/segment_api.py
```diff
@@ -254,27 +211,6 @@ def get_bulk_delete_status(self, bulk_delete_id):
...
         """
-        query = {
-            'query': BULK_DELETE_STATUS_QUERY.format(bulk_delete_id)
-        }
-
-        resp = self._call_segment_graphql(query)
-        resp_json = resp.json()
-        LOG.info(text_type(resp_json))
-
-    def get_all_deletion_requests(self, per_page):
```
bmedx, line 265: I've removed this API and the script that called it. It never worked correctly as noted in the TODO below and I think we have way too many deletions to use this sanely at this point. We should be able to get this information from the Segment UI, but it can be re-added with tests if we need it in the future.

tubular/scripts/query_segment_bulk_delete_status.py
```diff
@@ -63,11 +63,10 @@ def query_bulk_delete_id(config_file, bulk_delete_id):
     config = CONFIG_OR_EXIT(config_file)
 
     segment_base_url = config['base_urls']['segment']
-    auth_email = config['segment_email']
-    auth_password = config['segment_password']
+    auth_token = config['segment_auth_tokenz']
```
pwnage101: typo: tokenz

tubular/tests/test_segment_api.py
pwnage101: So just so I understand, we don't need to request an auth token anymore?

bmedx: Correct. Now we generate fixed tokens in the Segment UI.

tubular/tests/test_segment_api.py
```diff
@@ -145,31 +110,6 @@ def test_bulk_delete_error(setup_bulk_delete, caplog):  # pylint: disable=redefi
     with pytest.raises(Exception):
         segment.delete_learners(learner, 1000)
 
-    assert mock_post.call_count == 1
+    assert mock_post.call_count == 4
```
pwnage101, line 113: Where did the extra 3 requests come from?

pwnage101:
LGTM, just had some confusion around the mock_post in tests.

tubular/tests/test_segment_api.py
bmedx: The test now actually exercises the backoff, so it gets called for each retry

tubular/segment_api.py
```diff
@@ -15,55 +17,15 @@
...
-      next
-      limit
-    }}}}
-  }}}}
-}}}}""".format(DELETION_REQUEST_LIST_QUERY_OPNAME, '{}', '{}')
-
-# According to Segment, represents the maximum limits of the bulk delete mutation call.
-MAXIMUM_USERS_IN_DELETE_REQUEST = 16 * 1024  # 16k users
+# The Segment Config API for bulk deleting users for a particular workspace
+BULK_DELETE_URL = 'v1beta/workspaces/{}/regulations'
```
brianhw, line 21: Do we have any idea how long "v1beta" will last?   It sounds very transient....

brianhw (approved):
LGTM.  👍    

Just one idle question, more for the future....

tubular/segment_api.py
bmedx: No, they just mentioned "oh that never came out of beta", but they do monitor the endpoint usage and have been good in the past about warning us when things are going to change.

pwnage101 (approved):

PR #370, opened 2019-09-11
feat: add npm package overrides to frontend builder

abutterworth:
Consume config to override npm packages, leveraging npm aliases.

Sample config in edx-internal: https://github.com/edx/edx-internal/blob/pipeline-test/frontends/frontend-app-profile/stage_config.yml
```
APP_CONFIG:
  # 
  # 
NPM_ALIASES:
  "@edx/frontend-component-open-edx-header": "npm:@edx/frontend-component-edx-header"
```

### DIFF
--- a/tubular/scripts/frontend_utils.py
+++ b/tubular/scripts/frontend_utils.py
@@ -57,13 +57,28 @@ class FrontendBuilder:
         return_code = proc.wait()
         if return_code != 0:
             self.FAIL('Could not run `npm install` for app {}.'.format(self.app_name))
-        npm_overrides = self.get_override_config()
-        if npm_overrides:
-            aliased_installs = ' '.join(['{}@{}'.format(k, v) for k, v in npm_overrides.items()])
-            override_proc = subprocess.Popen(['npm install npm && ./node_modules/.bin/npm install {}'.format(aliased_installs)], cwd=self.app_name, shell=True)
-            override_proc_return_code = override_proc.wait()
-            if override_proc_return_code != 0:
-                self.FAIL('Could not run `npm install npm && npm install` overrides {} for app {}.'.format(
+
+        self.install_npm_alias_requirements()
+
+    def install_npm_alias_requirements(self):
+        """ Install npm alias requirements for app to build """
+        npm_aliases = self.get_npm_aliases_config()
+        if npm_aliases:
+            # Install the latest version of npm that supports aliases (>= 6.9.0)
+            proc = subprocess.Popen(['npm install npm'], cwd=self.app_name, shell=True)
+            return_code = proc.wait()
+            if return_code != 0:
+                self.FAIL('Could not run `npm install npm` for app {}.'.format(self.app_name))
+
+            aliased_installs = ' '.join(['{}@{}'.format(k, v) for k, v in npm_aliases.items()])
+            install_aliases_proc = subprocess.Popen(
+                ['./node_modules/.bin/npm install {}'.format(aliased_installs)],
+                cwd=self.app_name,
+                shell=True
+            )
+            install_aliases_proc_return_code = install_aliases_proc.wait()
+            if install_aliases_proc_return_code != 0:
+                self.FAIL('Could not run `npm install` aliases {} for app {}.'.format(
                     aliased_installs, self.app_name
                 ))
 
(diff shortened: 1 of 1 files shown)


tubular/scripts/frontend_utils.py
```diff
@@ -53,10 +53,19 @@ def _get_configs(self):
 
     def install_requirements(self):
         """ Install requirements for app to build """
-        proc = subprocess.Popen(['npm install'], cwd=self.app_name, shell=True)
+        proc = subprocess.Popen(['npm install -g npm && npm install'], cwd=self.app_name, shell=True)
```
douglashall: What does this do?

tubular/scripts/frontend_utils.py
```diff
@@ -53,10 +53,19 @@ def _get_configs(self):
...
+            self.FAIL('Could not run `npm install -g npm && npm install` for app {}.'.format(self.app_name))
+        npm_overrides = self.get_override_config()
+        if npm_overrides:
+            aliased_installs = ' '.join(['{}@{}'.format(k, v) for k, v in npm_overrides.items()])
+            override_proc = subprocess.Popen(['npm install {}'.format(aliased_installs)], cwd=self.app_name, shell=True)
+            override_proc_return_code = override_proc.wait()
+            if override_proc_return_code != 0:
+                self.FAIL('Could not run `npm install` overrides {} for app {}.'.format(
+                    aliased_installs, self.app_name
+                ))
```
douglashall, line 84: Maybe move this to a separate function `install_requirements_overrides` with a docstring explaining what's going on here.

tubular/scripts/frontend_utils.py
```diff
@@ -66,6 +75,14 @@ def get_app_config(self):
             self.LOG('Config variables do not exist for app {}.'.format(self.app_name))
         return app_config
 
+    def get_override_config(self):
```
douglashall: `get_requirements_override_config`

douglashall (approved):

tubular/scripts/frontend_utils.py
tuchfarber: Agreed on the above. Also, not entirely sure the process of installing these aliases. Are we expecting a base version to be installed, then these install over it? I'm trying to think through if there's going to be any issues (performance, side-effect, etc) of running the two version of `npm install`

tuchfarber (commented):
Mostly just agreeing with Doug's points, as far as structure.

I don't particularly like the work `override` being used as the whole system (esp. when we get into multi-deployment) is a series of layered overrides. Under the assumption that this new key and code is going to be used just for npm aliases, could we more specifically call it NPM_ALIASES and use that terminology throughout the code? Less time groking what's an alias override vs an app_config override.

tubular/scripts/frontend_utils.py
abutterworth: We're having trouble upgrading node and npm in the go-agent Dockerfiles. This is a temporary work around to upgrade npm without needing the get that to work first. Less than ideal, but the easiest path forward for now.

tubular/scripts/frontend_utils.py
abutterworth: We're expecting a default, open-edx version of packages to be installed. To your concerns:
 - Performance-wise the only issue here is that we'll slow down frontend builds
 - Side-effects: I think npm should re-evaluate the dependency tree given a new package to consume. This should be okay assuming we are thoughtful about using peer-dependencies in these swappable packages. If we aren't careful it is possible we could get duplicate versions of react in our builds.

tubular/scripts/frontend_utils.py
douglashall: I'd probably wait on the node upgrade, unless there is a pressing need to get this change in before that.

tubular/scripts/frontend_utils.py
davidjoy: One thing that might make the code a bit more consistent / remove some ambiguity around npm:

1. Upgrade npm using `npm install -g npm` prior to using it for the first time in `install_requirements`.
2. Then you can just call the global `npm install` in both `install_requirements` and in `install_requirements_npm_aliases`.

That way you're not using two separate versions of `npm install`, one from global, and one from node_modules/.bin`.  I'd be worried that if they're different enough we might see some strange interactions, as NPM is pretty notorious for changing things version to version.

To @tuchfarber's comment about performance w/r/t using `npm install` multiple times - the first call will install everything in package.json, and the call in `install_requirements_npm_aliases` is doing targeted installs of specific packages.  So it shouldn't be doing unnecessary work or doing the same work twice.

abutterworth: > Upgrade npm using npm install -g npm prior to using it for the first time in install_requirements.

We can't install anything globally from this context (for lack of permissions).

PR #372, opened 2019-09-19
tubular script

syed-awais-ali:


tubular/scripts/retrieve_latest_base_ami.py
```diff
@@ -0,0 +1,95 @@
...
+
+    try:
+        # edp_ami_id = ec2.active_ami_for_edp(environment, deployment, play)
+        if override:
+            ami_id = override
+        else:
+            url = "https://cloud-images.ubuntu.com/locator/ec2/releasesTable"
+            ami_id = requests.get(url).json
+            ubuntu_version = config['base_ami_ubuntu_version']
+            # parse to get ami for ubuntu_version
```
syed-awais-ali: Just to single out the ami from here. Will add that

tubular/scripts/retrieve_latest_base_ami.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+    try:
+        # edp_ami_id = ec2.active_ami_for_edp(environment, deployment, play)
+        if override:
+            ami_id = override
+        else:
+            url = "https://cloud-images.ubuntu.com/locator/ec2/releasesTable"
+            ami_id = requests.get(url).json
+            ubuntu_version = config['base_ami_ubuntu_version']
```
syed-awais-ali: Should we pass an infile here as well to read the version of ubuntu distro?

tubular/scripts/retrieve_latest_base_ami.py
```diff
@@ -0,0 +1,95 @@
...
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
+
+from tubular import ec2  # pylint: disable=wrong-import-position
+
+logging.basicConfig(level=logging.INFO)
+
+
+@click.command()
+@click.option(
+    '--environment', '-e',
```
coryleeio: You can replace the log statement with something that just prints the ami it found.  It has no need of knowledge of the environment / deployment / or play.  It is only used in logging.

coryleeio (changes requested):

tubular/scripts/retrieve_latest_base_ami.py
coryleeio: I would prefer you pass it as an arg if possible, as that is easier to run locally for testing.

But a file is fine too if that is difficult.

coryleeio: We discussed and Awais confirmed that we cannot read the config from here.  He will need to pass something in, either an arg or filename.  I would prefer individual arguements as it is easier to use the script by using 

my_script.py --help

and running it locally.

coryleeio (changes requested):

tubular/scripts/retrieve_latest_base_ami.py
```diff
@@ -0,0 +1,87 @@
...
+    Method used to retrieve the last base AMI ID used for an environment/deployment/play.
+    """
+
+    try:
+        # edp_ami_id = ec2.active_ami_for_edp(environment, deployment, play)
+        if override:
+            ami_id = override
+        else:
+            url = ""
+            ubuntu_version = config['base_ami_ubuntu_version']
```
coryleeio: config is not defined so this can't run ATM.

coryleeio (changes requested):

tubular/scripts/retrieve_latest_base_ami.py
syed-awais-ali: @coryleeio passed this from script in edx-gomatic.

tubular/scripts/retrieve_latest_base_ami.py
```diff
@@ -0,0 +1,89 @@
...
+        if override:
+            ami_id = override
+        else:
+            url = ""
+            if ubuntu_version == "16.04":
+                url = "http://cloud-images.ubuntu.com/query/xenial/server/released.current.txt"
+            elif ubuntu_version == "18.04":
+                url = "http://cloud-images.ubuntu.com/query/bionic/server/released.current.txt"
+            data = requests.get(url)
+            parse_ami = re.search('us-east-1(.+?)hvm', data.content)
```
coryleeio: parameterize the region here

coryleeio: you can default to us-east-1

coryleeio (dismissed):

syedimranhassan (approved):

PR #380, opened 2019-10-07
Fix ami building from scratch

jdmulloy:
This is a bit of a hack, the AMI id will be the latest Ubuntu AMI, but
the tags will be from the AMI of the running instances so that the
scripts that message PRs and write release wiki pages will still work.

I'll be adding a PR for gomatic so this works


bderusha (approved):

fredsmith (approved):
:shipit:

PR #384, opened 2019-10-17
Don't require jenkins tokens because they aren't both always required

cpennington:


cpennington:
I'm going to aggressively merge this. It won't break anything that is currently working.

PR #387, opened 2019-12-04
Format the current timestamp into release wiki titles on request

cpennington:


tubular/scripts/update_release_page.py
```diff
@@ -119,6 +120,8 @@ def create_release_page(
     if title is None:
         title = "{:%Y-%m-%d} Release".format(EXPECTED_RELEASE_DATE)
 
+    title = title.format(timestamp=datetime.now())
```
jdmulloy, line 128: Does this work? What is the content of title when before we get to this line?

jdmulloy (changes requested):

tubular/scripts/update_release_page.py
cpennington: This basically allows callers to use a title like `"{timestamp:%Y-%m-%d} Release"` to get a page like we have now, or to add more detail (like I do in the linked PR). If the title doesn't have a format field in it, then nothing is changed.

jdmulloy: Got it.

jdmulloy (approved):

tubular/scripts/update_release_page.py
jdmulloy: @cpennington Can you add a comment above this explaining this?

jdmulloy (changes requested):

jdmulloy (approved):

PR #398, opened 2020-01-29
Add hubpot API for user retirement

tuchfarber:


tubular/hubspot_api.py
```diff
@@ -0,0 +1,82 @@
...
+        HubspotException,
+        max_tries=MAX_ATTEMPTS
+    )
+    def delete_user(self, learner):
+        """
+        Delete a learner from hubspot using their email address.
+        """
+        email = learner.get('original_email', None)
+        if not email:
+            raise TypeError('Expected an email address for user to delete, but received None.')
```
crice100, line 40: I'm wondering if we should log any sort of id if the delete fails. Or will the job automatically detect and handle this?

staubina (approved):

tubular/hubspot_api.py
tuchfarber: All of the other jobs specifically don't list any user details, so I assume that's either because we don't want to store that data in logs, or the job is posting that at the top of this log. I'll confirm when I talk to DE about implementation. I'll merge as is, and open a new one if neither of those are the case.

PR #402, opened 2020-02-20
Updating message_pr_in_range logging so we can debug what it cannot find

coryleeio:


tubular/scripts/message_prs_in_range.py
```diff
@@ -173,7 +173,12 @@ def message_pr(api, message_type, pull_request, extra_text):
     """
     LOG.info("Github API Rate Limit: {}".format(api.get_rate_limit()))
     LOG.info(u"Posting message type %r to %d.", message_type.name, pull_request.number)
-    api.message_pr_with_type(pr_number=pull_request, message_type=message_type, extra_text=extra_text)
+
+    try:
+        api.message_pr_with_type(pr_number=pull_request, message_type=message_type, extra_text=extra_text)
+    except github.GithubException.UnknownObjectException as exc:
+        LOG.error(u"message_pr_with_type args were: pr_number={0} message_type={1} extra_text={2}", pull_request, message_type, extra_text)
```
feanil: add a `raise` line after this so that we still raise the full original exception after logging.  Otherwise it will not seem like this task failed.

feanil (changes requested):

tubular/scripts/message_prs_in_range.py
jdmulloy: @feanil does line 182 address your concern?

feanil (approved):
Assuming you tested this.

feanil (approved):

PR #412, opened 2020-04-06
Backoff on MultipleImagesFoundException.

feanil:
This exception can be thrown if the cluster we care about is in a bad
state but it can also be thrown when a deploy is happening.  Also doing
a backoff when we see this exception should make us more resilient to
running this while a deploy is happening.


feanil:
I've seen this multiple times today as I try to get code out.  I think with CD we will see this a lot more for edx-platform.  This should make the pipeline more resilient.

dianakhuang (approved):

PR #420, opened 2020-05-21
Ignore the dependabot context on test status.

feanil:
It seems to be putting a context on the master commit via the API that
is not visible through github UI but is causig our deploy pipelines to
fail because it remains in pending state for a long time.

We don't want to block deploys on whether or not dependabot has run some
check on master.


robrap (approved):
Makes sense to me generally.  Searching for usage helped me learn at least a little piece of the pipeline.

PR #422, opened 2020-05-28
Add retry to Salesforce api call

doctoryes:
A recent retirement user failed in the pipeline due to a failed API call that did not have retry logic around it. Add logic to retry the call.

Also, upgrade the requirements to fix this vulnerability alert:
https://github.com/edx/tubular/network/alerts

### DIFF
--- a/pip-tools.txt
+++ b/pip-tools.txt
@@ -5,8 +5,8 @@
 #    pip-compile --no-emit-trusted-host --no-index pip-tools.in
 #
 click==7.1.2              # via pip-tools
-pip-tools==5.1.2          # via -r pip-tools.in
-six==1.14.0               # via pip-tools
+pip-tools==5.2.0          # via -r pip-tools.in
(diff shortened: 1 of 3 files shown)


doctoryes:
Admittedly, I added no tests for the retry logic. If reviewers feel it necessary, I'll add tests for it.

macdiesel (approved):

bmedx (approved):
I have a vague recollection that at least one of these failures was getting the auth token, in the `Salesforce()` constructor. It doesn't look like there's a retry added there, but I'm also not sure if my memory is correct. So thumbing, but if your memory matches we might want to see if there's a reasonable way to add that. I know we get a lot of failures on LMS auth tokens as well.

doctoryes:
@bmedx I didn't see that failure on this on-call shift. But I vaguely remember that one as well. I'll add a retry on that call as well.

PR #433, opened 2020-07-09
MICROBA-447 | Add support for removing user data in Demographics IDA

justinhynes:
Add support for removing data in Demographics IDA

### DIFF
--- a/tubular/scripts/helpers.py
+++ b/tubular/scripts/helpers.py
@@ -22,7 +22,7 @@ from six import text_type
 # Add top-level module path to sys.path before importing tubular code.
 sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
 
-from tubular.edx_api import CredentialsApi, EcommerceApi, LmsApi  # pylint: disable=wrong-import-position
+from tubular.edx_api import CredentialsApi, DemographicsApi, EcommerceApi, LmsApi  # pylint: disable=wrong-import-position
 from tubular.segment_api import SegmentApi   # pylint: disable=wrong-import-position
 from tubular.sailthru_api import SailthruApi   # pylint: disable=wrong-import-position
 from tubular.salesforce_api import SalesforceApi   # pylint: disable=wrong-import-position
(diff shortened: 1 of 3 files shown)


tubular/scripts/helpers.py
```diff
@@ -198,6 +200,9 @@ def _setup_all_apis_or_exit(fail_func, fail_code, config):
         if credentials_base_url:
             config['CREDENTIALS'] = CredentialsApi(lms_base_url, credentials_base_url, client_id, client_secret)
 
+        if demographics_base_url:
+            config['DEMOGRAPHICS'] = DemographicsApi(lms_base_url, demographics_base_url, client_id, client_secret)
```
crice100, line 204: It seems odd to pass the _lms_base_url_ here if we're directly calling demographics, but that does seem to be how this works.

crice100 (approved):

tubular/scripts/helpers.py
justinhynes: Agreed. I see why it's needed now that I've debugged through the code few times, but I thought it was odd as well. 

My understanding is that its used when creating instances of the various API clients to build a request to retrieve-and-store a JWT token the client will use later.

doctoryes: It uses the LMS base URL to authN with the LMS before calling demographics.

tubular/scripts/retire_one_learner.py
```diff
@@ -10,9 +10,11 @@
     lms: http://localhost:18000/
     ecommerce: http://localhost:18130/
     credentials: http://localhost:18150/
+    demographics: http://localhost:18360/
 retirement_pipeline:
     - ['RETIRING_CREDENTIALS', 'CREDENTIALS_COMPLETE', 'CREDENTIALS', 'retire_learner']
     - ['RETIRING_ECOM', 'ECOM_COMPLETE', 'ECOMMERCE', 'retire_learner']
+    - ['RETIRING_DEMOGRAPHICS', 'DEMOGRAPHICS_COMPLETE', 'DEMOGRAPHICS', 'retire_learner']
```
doctoryes, line 17: Ugh - this docstring list is out-of-date. It's missing a few state transitions, as you can see in the edx-internal PR: 
https://github.com/edx/edx-internal/pull/2398
But - it's a nit.

doctoryes (commented):
👍

doctoryes (approved):

tubular/scripts/retire_one_learner.py
justinhynes: Yeah. I see what you are saying. I would love to update the docstring further... I'm just not sure I have all the information to update it correctly.  I'm not 100% sure what the service each stage belongs to. I can update this in a follow up PR if you could share what those other values may be (I'm assuming they are in the config I can't see in `edx-ops`).

doctoryes: @justinhynes I think it's OK enough as is. The docstring's purpose is just to show an example anyway. So IMO go ahead and merge - no need for a further change.

PR #434, opened 2020-07-09
Fix Segment retirement ID issue

bmedx:


tubular/segment_api.py
```diff
@@ -151,17 +164,17 @@ def delete_learners(self, learners, chunk_size, beginning_idx=0):
...
-                learners[end_idx]['id'], learners[end_idx]['original_username']
+                learners[start_idx]['user']['id'], learners[start_idx]['original_username'],
+                learners[end_idx]['user']['id'], learners[end_idx]['original_username']
             )
 
             learner_vals = []
             for idx in range(start_idx, end_idx + 1):
                 for id_key in REQUIRED_IDENTIFYING_KEYS:
-                    learner_vals.append(text_type(learners[idx][id_key]))
+                    learner_vals.append(self._get_value_from_learner(learners[idx], id_key))
```
bmedx, line 174: This is, in combination with line 17 and the new `_get_value_from_learner` are the real fix. Everything else is test cleanup.

tubular/tests/retirement_helpers.py
```diff
@@ -46,9 +46,9 @@ def fake_config_file(f, orgs=None, fetch_ecom_segment_id=False):
         'client_id': 'bogus id',
         'client_secret': 'supersecret',
         'base_urls': {
-            'credentials': 'https://credentials.stage.edx.org/',
-            'lms': 'https://stage-edx-edxapp.edx.org/',
-            'ecommerce': 'https://ecommerce.stage.edx.org/',
+            'credentials': 'https://credentials.stage.edx.invalid/',
+            'lms': 'https://stage-edx-edxapp.edx.invalid/',
+            'ecommerce': 'https://ecommerce.stage.edx.invalid/',
```
doctoryes, line 51: Good move! This must be where stage servers where being hit...

tubular/tests/retirement_helpers.py
```diff
@@ -72,6 +72,55 @@ def fake_config_file(f, orgs=None, fetch_ecom_segment_id=False):
...
+            "username": current_username,
+            "email": current_email,
+            "profile": {
+                "id": 10009,
+                "name": current_name
+            }
+        },
+        "created": "2018-10-18T20:08:03.724805",
+        "modified": "2018-10-18T20:08:03.724805",
+    }
```
doctoryes, line 121: The one test I was looking for - and haven't yet found - is the one that uses this fake user retirement - and then checks to ensure that `9009` is sent as the id and `1` is *not*.  Does that test exist somewhere?

tubular/tests/test_segment_api.py
```diff
@@ -80,10 +80,12 @@ def test_bulk_delete_success(setup_bulk_delete):  # pylint: disable=redefined-ou
 
     assert mock_post.call_count == 1
 
-    learners_vals = []
-    for curr_key in ['id', 'original_username', 'ecommerce_segment_id']:
-        curr_id = learner[0][curr_key]
-        learners_vals.append(text_type(curr_id))
+    expected_learner = get_fake_user_retirement()
+    learners_vals = [
+        text_type(expected_learner['user']['id']),
```
bmedx, line 85: @doctoryes this is the test you're looking for

doctoryes: Yes. Yes it is.

doctoryes (approved):

PR #436, opened 2020-07-16
MICROBA-447 | Add logging for demographics retirement

justinhynes:
[MICROBA-447]
- Add logging when no data is found for user during demographics retirement.


bmedx (approved):
👍

PR #450, opened 2020-10-21
Allow HubspotAPI to send email to marketing upon successful deletion

bseverino:
### [MST-482](https://openedx.atlassian.net/browse/MST-482)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 CHANGES
 =======
 
+* MST-482 - Send email alert upon successful Hubspot deletion
 * ISRE-588 - update javasocketerror raise logic
 * ISRE-587 - update giveup logic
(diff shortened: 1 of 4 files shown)


schenedx (approved):
👍

pwnage101:
My only question is about whether the hubspot VID would be considered PII?

bseverino:
@pwnage101 That's a good question, I'm not 100% sure. I'll see if I can find out.

tuchfarber (approved):

bseverino:
@pwnage101 According to the Hubspot docs, anything using the VID requires authentication, so I don't think it would be considered PII: https://legacydocs.hubspot.com/docs/methods/contacts/get_contact

pwnage101 (approved):

PR #451, opened 2020-10-22
removed exception

syed-awais-ali:


cpennington (approved):
This feels fine in a vacuum. Do we want to/have we added any alerts that trigger if multiple AMIs are deployed for a single edp for longer than some period of time, to make sure that we aren't running in split-brain mode for long periods?

PR #456, opened 2020-11-09
Upgrading from click 6.7 to 7 to fix a collision in gocd20 pipelines that occurs due to a missing separate install of python which causes an unintentional downgrade

coryleeio:


coryleeio:
https://click.palletsprojects.com/en/7.x/upgrading/#upgrading-to-7-0

PR #459, opened 2020-11-17
Revert "Allow HubspotAPI to send email to marketing upon successful deletion"

bseverino:
Reverts edx/tubular#450

`ses:SendEmail` failed due to possible missing AWS permissions, reverting to unblock the build until it is resolved


schenedx (approved):
👍

PR #487, opened 2021-01-05
Removing update recent deployers related code

christopappas:
Bumped tox.ini's py35 env to py36 because `make test` was failing.

Confirmed all tests pass with py36


coryleeio (changes requested):
I believe none of the methods in gocd_api.py are used





12:09
or would work
12:09
so i think you can delete it wholesale
12:10
What I was referring to is that there is a script that usees one of the other methods in that file
12:10
but it itself is no longer used in internal
12:10
so you can delete it to.
12:10
if you grep for the other methods in gocd_api.py you will find the unused script calling them

tubular/scripts/update_recent_deployers.py
```diff
@@ -14,19 +14,7 @@
 from tubular.opsgenie_api import OpsGenieAPI
```
coryleeio: I think you can delete this whole file

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{35,38}
+envlist = py{36,38}
```
coryleeio: There are a few other places this needs to be updated, grep around for 35 3.5 etc

it should atleast appear in the Dockerfile.

I'm not sure if we can rip this out, i'm asking Julia why we were trying to support 3.5 last week, I think we might still be using it in some places... tools jenkins at the very least

coryleeio: What were the 3.5 errors like, do you have a link?

tubular/scripts/update_recent_deployers.py
christopappas: oh i'm missing a commit, yes this file is going. oops

tox.ini
christopappas: reverted this

christopappas:
> I believe none of the methods in gocd_api.py are used
> 
> 12:09
> or would work
> 12:09
> so i think you can delete it wholesale
> 12:10
> What I was referring to is that there is a script that usees one of the other methods in that file
> 12:10
> but it itself is no longer used in internal
> 12:10
> so you can delete it to.
> 12:10
> if you grep for the other methods in gocd_api.py you will find the unused script calling them

tubular/scripts/find_and_advance_pipeline.py uses GOCDAPI... was that the script you were saying is also no longer used?

coryleeio (approved):

PR #494, opened 2021-01-11
AA-595: Add script to delete Braze user information

mikix:
Adds two new config variables:
- braze_api_key: a REST API key from Braze with users.delete access
- braze_instance: a Braze instance key like iad-01

And sets the BRAZE service variable & 'delete_user' method, for use in retirement pipelines.

### DIFF
--- a/tubular/braze_api.py
+++ b/tubular/braze_api.py
@@ -1,0 +1,85 @@
+"""
+Helper API classes for calling Braze APIs.
+"""
+import logging
+import os
+
+import backoff
+import requests
+
+LOG = logging.getLogger(__name__)
+MAX_ATTEMPTS = int(os.environ.get('RETRY_BRAZE_MAX_ATTEMPTS', 5))
+
+
+class BrazeException(Exception):
+    pass
+
+
+class BrazeRecoverableException(BrazeException):
+    pass
+
+
+class BrazeApi:
+    """
+    Braze API client used to make calls to Braze
+    """
+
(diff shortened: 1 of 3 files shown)


tubular/scripts/helpers.py
```diff
@@ -183,6 +187,12 @@ def _setup_all_apis_or_exit(fail_func, fail_code, config):
 
         config['LMS'] = LmsApi(lms_base_url, lms_base_url, client_id, client_secret)
 
+        if braze_api_key:
+            config['BRAZE'] = BrazeApi(
+                braze_api_key,
+                braze_instance,
+            )
```
Dillon-Dumesnil, line 194: Nit: one line?

tubular/braze_api.py
```diff
@@ -0,0 +1,85 @@
...
+
+class BrazeApi:
+    """
+    Braze API client used to make calls to Braze
+    """
+
+    def __init__(self, braze_api_key, braze_instance):
+        self.api_key = braze_api_key
+
+        # https://www.braze.com/docs/api/basics/#endpoints
```
Dillon-Dumesnil, line 30: 👏

Dillon-Dumesnil (approved):
LGTM

tubular/scripts/helpers.py
mikix: Fair nit, but all the other stanzas just like this were doing one-arg-per-line, so I figured consistency was better here.

PR #496, opened 2021-01-20
ISRE-618 | Delete 0 Instance ASGs during deploy step  if they are cre…

christopappas:
…ated

### DIFF
--- a/tubular/tests/test_asgard.py
+++ b/tubular/tests/test_asgard.py
@@ -1720,3 +1720,135 @@ class TestAsgard(unittest.TestCase):
             asgard.get_asg_info(asg)
         error_message = "Call to asgard failed with status code: {}".format(403)
         self.assertTrue(str(context_manager.exception).startswith(error_message))
+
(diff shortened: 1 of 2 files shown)


tubular/tests/test_asgard.py
```diff
@@ -1720,3 +1720,135 @@ def test_get_asg_info_403(self, req_mock):
...
+                'maxSize': 0,
+                'autoScalingGroupName': 'loadtest-edx-edxapp-v1338'
+            },
+            {
+                'desiredCapacity': 8,
+                'minSize': 2,
+                'maxSize': 20,
+                'autoScalingGroupName': 'loadtest-edx-edxapp-v1339'
+            },
+        ]
```
christopappas, line 1821: i honestly don't think we'll ever get into this state, but I think this is indeed how I'd want it to behave (the spirit off it being: make sure the latest thing can be copied without being empty... but also don't be greedy and delete more than you need to)

coryleeio (approved):

adzuci (approved):
:shipit:

PR #498, opened 2021-01-21
Python Code Cleanup

mzulqarnain1:
The last attempt to merge these changes failed as the pipelines broke. We need to find the reason this happened and attempt again.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.5', '3.8']
+        python-version: ['3.8']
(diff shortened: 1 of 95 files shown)


jmbowman:
Cory's explanation after the previous pipeline breakage:

> Yeah upgrading this repo requires every job using tubular in build jenkins, tools jenkins and GoCD be upgraded.  We've recently got tools and build up above 3.6
> All have 3.8 available but the jobs would all have to be updated to use 3.8 at the same time to go to 3.8 everywhere

So I think the next step is creating a ticket to enumerate the places this is used and ticket the updates of each one to use Python 3.8 if it isn't already.  It should be possible to do those independently, and then merge this once they're all updated.

PR #506, opened 2021-03-16
Added script to create k8s job

syedimranhassan:


tubular/scripts/create_k8s_job.py
```diff
@@ -0,0 +1,55 @@
...
+    Create the k8s job in given namespace.
+    """
+    configuration = client.Configuration()
+    configuration.host = "127.0.0.1:8888"
+    image = get_deploymemnt(configuration, namespace, deployment_name)
+    config.load_kube_config()
+    batch_v1 = client.BatchV1Api()
+    job = create_job_object(deployment_name, image, command)
+    create_job(batch_v1, job, namespace)
+    # An exit code of 0 means success and non-zero means failure.
```
christopappas, line 59: I know this is a WIP, but just adding a note that we should be sure that once a job is complete or fails etc, the job should get deleted (to prevent bloat)

syedimranhassan: I have added the delete function, that will delete the job after getting the logs

tubular/scripts/create_k8s_job.py
```diff
@@ -0,0 +1,48 @@
...
+    job = create_job_object(deployment_name, image, command)
+    create_job(k8s_client, job, namespace)
+    while True:
+        job_status = get_status(k8s_client, deployment_name, namespace)
+        if job_status.failed is None and job_status.succeeded is None:
+            time.sleep(2)
+            continue
+        else:
+            break
+    get_logs(k8s_client, deployment_name, namespace)
```
coryleeio: Is it possible to move this loop into get logs or into a tail logs function and print the logs every few seconds and clear the terminal when the job is completed? so that you get a natural scrolling by of the logs if you create a job in the terminal instead of waiting for the job to complete before any logging is returned? 

For jobs that take 8+ hours getting the feedback immediately will often be helpful

tubular/eks.py
```diff
@@ -0,0 +1,93 @@
+""" Commands to interact with the Slack API. """
```
coryleeio: Can you rename this file to k8s or kubernetes.py since it is not specific to EKS in any way currently that I can tell

coryleeio: (it does not make any boto calls)

tubular/scripts/create_k8s_job.py
```diff
@@ -0,0 +1,48 @@
...
+def create_k8s_job(deployment_name, namespace, command):
+    """
+    Create the k8s job in given namespace.
+    """
+    configuration = client.Configuration()
+    configuration.host = "127.0.0.1:8888"
+    image = get_deploymemnt(configuration, namespace, deployment_name)
+    config.load_kube_config()
+    k8s_client = client.BatchV1Api()
+    job = create_job_object(deployment_name, image, command)
```
coryleeio: When you retrieve the command from the deployment, can you detect the newrelic-admin command and if it is present in the existing deployment can you keep it and append the new command, otherwise just set the command to the new command.

The intent is that if the application sends logs to newrelic we want to also send the newrelic metrics when we run the mgmt command.

Please implement this in the tubular script, and not in the K8s one since the K8s library is very pure ATM and this is business logic

coryleeio (commented):
Looks great Imran, I left a few comments, but it looks like it's very close, thank you!

tubular/scripts/create_k8s_job.py
```diff
@@ -0,0 +1,48 @@
...
+@click.option(
+    '--namespace',
+    required=True,
+    help="Namespace in which create the job."
+)
+@click.option(
+    '--command',
+    required=True,
+    help="Command to run in job.",
+)
```
coryleeio, line 21: Can you add the ability to overwrite the cpu and memory here as well.  Since sometimes the deployment pod will have much smaller memory needs than the mgmt commands

tubular/eks.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+LOG = logging.getLogger(__name__)
+LOG.setLevel(logging.INFO)
+
+
+def create_job_object(name, image, command):
+    try:
+        # Configureate Pod template container
+        container = client.V1Container(
```
christopappas: nitpicky, but I would love things like:
```
container = client.V1Container(
            name=name,
            image=image,
            command=command.split(" "))
```
to have the last parens moved to its own line (white space changed too)

```
container = client.V1Container(
    name=name,
    image=image,
    command=command.split(" ")
)
```

tubular/eks.py
```diff
@@ -0,0 +1,93 @@
...
+
+def get_logs(k8s_client, job_name, namespace):
+    job_def = k8s_client.read_namespaced_job(name=job_name, namespace=namespace)
+    controllerUid = job_def.metadata.labels["controller-uid"]
+    core_v1 = client.CoreV1Api()
+    pod_label_selector = "controller-uid=" + controllerUid
+    pods_list = core_v1.list_namespaced_pod(namespace=namespace, label_selector=pod_label_selector, timeout_seconds=10)
+    pod_name = pods_list.items[0].metadata.name
+    try:
+        pod_log_response = core_v1.read_namespaced_pod_log(name=pod_name, namespace=namespace,
```
christopappas: this one i feel more strongly about because it gets each arg on its own line:
```
pod_log_response = core_v1.read_namespaced_pod_log(
    name=pod_name,
    namespace=namespace,
    _return_http_data_only=True,
    _preload_content=False,
    follow=True
)
```

tubular/eks.py
```diff
@@ -0,0 +1,93 @@
...
+    except ApiException as e:
+        print("Exception: %s\n" % e)
+
+
+def get_logs(k8s_client, job_name, namespace):
+    job_def = k8s_client.read_namespaced_job(name=job_name, namespace=namespace)
+    controllerUid = job_def.metadata.labels["controller-uid"]
+    core_v1 = client.CoreV1Api()
+    pod_label_selector = "controller-uid=" + controllerUid
+    pods_list = core_v1.list_namespaced_pod(namespace=namespace, label_selector=pod_label_selector, timeout_seconds=10)
```
christopappas: kind of a long line. you may consider splitting this the args over a number of lines too

```
pods_list = core_v1.list_namespaced_pod(
    namespace=namespace,
    label_selector=pod_label_selector,
    timeout_seconds=10
)
```

tubular/scripts/create_k8s_job.py
```diff
@@ -0,0 +1,48 @@
...
+def create_k8s_job(deployment_name, namespace, command):
+    """
+    Create the k8s job in given namespace.
+    """
+    configuration = client.Configuration()
+    configuration.host = "127.0.0.1:8888"
```
christopappas: will this always be the configuration host?

syedimranhassan: nope, It's for minikube, I'll remove it. Thanks

tubular/kubernetes.py
```diff
@@ -0,0 +1,134 @@
...
+        namespace=namespace,
+        label_selector=pod_label_selector,
+        timeout_seconds=10
+    )
+    pod_name = pods_list.items[0].metadata.name
+    while True:
+        pod_status = core_v1.read_namespaced_pod_status(name=pod_name, namespace=namespace)
+        if pod_status.status.phase != "Running" and pod_status.status.phase != "Succeeded":
+            time.sleep(5)
+            continue
```
christopappas: I don't think you need this continue here

tubular/kubernetes.py
```diff
@@ -0,0 +1,134 @@
...
+
+def create_job_object(name, command, args_command, deployment_specs_container, deployment_specs,
+                      cpu_request, memory_request, cpu_limit, memory_limit):
+    try:
+        # Configured Pod template container
+        container = client.V1Container(
+            name=name,
+            image=deployment_specs_container.image,
+            env=deployment_specs_container.env,
+            # image="perl",
```
christopappas: can drop the `# image="perl",` if not needed

tubular/kubernetes.py
```diff
@@ -0,0 +1,134 @@
...
+                # minimum amount of compute resources required
+                requests={"cpu": cpu_request, "memory": memory_request},
+                # maximum amount of compute resources allowed
+                limits={"cpu": cpu_limit, "memory": memory_limit}
+            )
+        )
+        # Create and Configured a spec section
+        template = client.V1PodTemplateSpec(
+            metadata=client.V1ObjectMeta(labels={"app": name}),
+            spec=client.V1PodSpec(restart_policy="Never", containers=[container], volumes=deployment_specs.volumes,
```
christopappas: nitpick - i would keep 1 argument per line

tubular/kubernetes.py
```diff
@@ -0,0 +1,134 @@
+import logging
+import time
+from kubernetes.client.rest import ApiException
+from kubernetes import client, config
+
+LOG = logging.getLogger(__name__)
+LOG.setLevel(logging.INFO)
```
christopappas: doesn't look like `LOG` is used, and that there are print statements throughout. is that intended?

tubular/kubernetes.py
```diff
@@ -0,0 +1,134 @@
...
+
+def get_status(k8s_client, name, namespace):
+    try:
+        job_status = k8s_client.read_namespaced_job_status(name, namespace, pretty='true')
+        return job_status.status
+    except client.rest.ApiException as e:
+        print("Exception when calling CoreV1Api->read_namespaced_pod_status: %s\n" % e)
+
+
+def check_create_args(source_arg, args, args_command):
```
christopappas: Maybe it's selfish, but i would looove if there was a test case for this function. I have no idea what the args look like, so it's hard to visualize/reason about at a glance

tubular/scripts/create_k8s_job.py
```diff
@@ -0,0 +1,65 @@
...
+    required=False,
+    help="Maximum amount of memory resources allowed.",
+)
+def create_k8s_job(deployment_name, namespace, command, args_command, cpu_limit, memory_limit):
+    """
+    Create the k8s job in given namespace.
+    """
+    source_arg = ["source /vault-api-secrets/secrets/secret.env", "newrelic-admin"]
+    configuration = client.Configuration()
+    configuration.host = "127.0.0.1:8888"  # Will remove this before merging
```
christopappas: don't forget to remove this before merging

christopappas (approved):
Gonna give a 👍 if it's working, but commented with a few minor/nitpicky things.

mrtmm:
Hi folks! If I am not mistaken, setting `cryptography` to `3.4.7` broke the Retirement service for us. `TASK [user_retirement_pipeline : Install python requirements]` is failing with the following message:
```
Could not find a version that satisfies the requirement cryptography==3.4.7 (from -r requirements/base.txt (line 18)) (from versions: 0.1, ... , 3.2.1)
```
`edx-platform` has [cryptography==3.2.1](https://github.com/edx/edx-platform/blob/master/requirements/edx/base.txt#L47)

syedimranhassan:
@mrtmm cryptography from version 3.3 drops the support for py3.5
You may need to update your python version.

mrtmm:
Alright, so this should only affect the Juniper release then. Thanks for the info, much appreciated! :)

PR #514, opened 2021-04-22
fix: re-add python 3.6 support (for GoCD)

pwnage101:
Turns out python 3.6 was still in our stack somewhere.


adzuci (approved):
:shipit: We may need to bump ISRE-466 and also look at asym-crypto and other prospectus build pipeline things before removing 3.6 support here.

PR #516, opened 2021-04-26
chore: Remove the Sailthru retirement stage

pwnage101:
edX no longer needs the Sailthru stage of the user retirement pipeline.

DESUPPORT-845


bmedx (commented):
I think you can remove this too: https://github.com/edx/tubular/blob/master/tubular/sailthru_api.py

brianhw (approved):
If it helps to move the retirement of Sailthru retirement along....

PR #520, opened 2021-05-17
fix: wrap frontend config override values in quotes

kdmccormick:
Avoids bash word-splitting when passing the values to
`npm run build`.

Without this change, we have had to work around the word-
splitting by writing config overrides like:
`{ "site_name": "'Your Platform Here'" }`
(notice that we need both single- and double-quotes)

Example PR working around the status quo: https://github.com/edx/edx-internal/pull/5101

PR #521, opened 2021-05-21
fix: make frontend_build handle numeric/bool config values again

kdmccormick:
(Also... `test: add some cursory tests for frontend building`)

I [broke frontend pipelines yesterday](https://github.com/edx/tubular/pull/520#pullrequestreview-665338006) while trying to make sure that all frontend config values were wrapped in quotes before being passed through bash to `npm build`. This worked, except for that I assumed that incoming config values were all strings, which broke frontends with numeric or boolean config settings like `cool_feature: true`.

### DIFF
--- a/tubular/tests/example-frontend-config/app.yml
+++ b/tubular/tests/example-frontend-config/app.yml
@@ -1,0 +1,35 @@
+S3_BUCKET_NAME: "coolfrontend.openedx.example.org"
+
+APP_CONFIG:
+
+  # Make sure we can override vars from common config.
+  COMMON_OVERRIDE_ME: overriden_value
+
+  # Check strings with spaces and special chars. Should all yield `key='The // value!'`.
+  VAR_WITH_SINGLE_QUOTES: 'The // value!'
+  VAR_WITH_DOUBLE_QUOTES: "The // value!"
+  VAR_WITH_SINGLE_THEN_DOUBLE_QUOTES: "'The // value!'"
+  # Except this one, which should yield `key="The // value!"`.
+  VAR_WITH_DOUBLE_THEN_SINGLE_QUOTES: '"The // value!"'
+
+  # Check ints. Should all yield `key='-100'`.
+  INT: -100
+  INT_WITH_QUOTES: '-100'
+
+  # Check floats. Should all yield `key='3.14'`.
+  FLOAT: 3.14
+  FLOAT_WITH_QUOTES: '3.14'
+
+  # Check True bools. Should all yield `key='True'`.
+  BOOL_TRUE: true
+  BOOL_TRUE_ANOTHER_WAY: yes  # YAML is funny - this is the same as `true`.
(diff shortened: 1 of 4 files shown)


adamstankiewicz (commented):
Overall, LGTM. I appreciate the test cases you added here 😄 

I have 2 questions:
1. What should happen if an app config variable is set to `false`? Is it worth having an explicit test case for this as well? Does it get treated as a falsey value or as a string that says `'False'`?
2. What would happen if we specified an array in the app config for an MFE (I'm not sure we handle this today actually 🤔 )? I don't have an exact use case in mind at the moment, but an example might be something like:

```yaml
FEATURES:
  - FEATURE_A: true
  - FEATURE_B: false
```

kdmccormick:
Great questions @adamstankiewicz 

Agreed that testing the False case is worthwhile. It should resolve to `'False'`; I'll add that test data now.

If you specify a list or dict, it will get Python-stringified and then stuffed into the `npm build` command line... which, depending on quoting nuances and what values have in the list, could "just work" or could break spectacularly. I would go as far to call this undefined behavior, actually. A robust solution would probably require dumping the list/dict to JSON and then carefully escaping any quotation marks in the output before passing it to bash. Yikes.

In terms of lists, what would you like to see for this PR? I could explicitly outlaw it by throwing an exception if the input isn't a simple string/bool/number/none, or I could leave things as-is and say that lists in config is undefined behavior. As much as a real solution would be good, I can't afford to bite that off right now 😛

adamstankiewicz:
@kdmccormick Yeah, agreed. Let's not tackle a robust list/dict solution through this PR. It might be worthwhile to throw an exception when the input isn't a simple string/bool/number/none just to make it more explicit that arrays aren't supported data types here.

Since it currently is undefined behavior today as you mentioned, I'm OK if we leave things as is for now but I think I'd be in favor of throwing an exception in this case just to make it more explicit/obvious in case an MFE maintainer decides to try using a list in their app config today.

doctoryes (approved):
With the addition of the tests, I feel confident about this fix. (Edited:...once tests are passing.)

tubular/tests/example-frontend-config/app.yml
```diff
@@ -0,0 +1,35 @@
...
+  # Check floats. Should all yield `key='3.14'`.
+  FLOAT: 3.14
+  FLOAT_WITH_QUOTES: '3.14'
+
+  # Check True bools. Should all yield `key='True'`.
+  BOOL_TRUE: true
+  BOOL_TRUE_ANOTHER_WAY: yes  # YAML is funny - this is the same as `true`.
+  BOOL_TRUE_WITH_QUOTES: 'True'  # Uppercase because Python loads this file.
+
+  # Check False bools. Should all yield `key='True'`.
```
adamstankiewicz: nit: Should this say `key='False'`?

adamstankiewicz (approved):

PR #530, opened 2021-09-08
Estute/user retirement archiver params

estute:
This PR adds a few improvements:
- Clearing up logging to indicate _which_ file is being published to s3. This is useful in the case that the script errors out after publishing to s3 but before finally clearing the users out of the LMS, which would require manual remediation.
- Adds `start_date` and `end_date` CLI params so we can target a specific set of user retirements for testing purposes
- Adds a `dry_run` CLI param so we can test the script out without impacting anything
- Adds a `batch_size` CLI param to limit the size of the request we are sending to the LMS to clean out retired users. In case we need to run this on a large number of users, being able to batch it reduces the amount of clean up we need to do if we were to hit an error.

### DIFF
--- a/tubular/scripts/retirement_archive_and_cleanup.py
+++ b/tubular/scripts/retirement_archive_and_cleanup.py
@@ -41,6 +41,7 @@ ERR_FETCHING = -3
 ERR_ARCHIVING = -4
 ERR_DELETING = -5
 ERR_SETUP_FAILED = -5
+ERR_BAD_CLI_PARAM = -6
 
 LOG = partial(_log, SCRIPT_SHORTNAME)
 FAIL = partial(_fail, SCRIPT_SHORTNAME)
@@ -63,6 +64,27 @@ def _fetch_learners_to_archive_or_exit(config, start_date, end_date, initial_sta
         FAIL_EXCEPTION(ERR_FETCHING, 'Unexpected error occurred fetching users to update!', exc)
 
 
+def _batch_learners(learners=None, batch_size=None):
+    """
+    To avoid potentially overwheling the LMS with a large number of user retirements to
+    delete, create a list of smaller batches of users to iterate over. This has the
+    added benefit of reducing the amount of user retirement archive requests that can
+    get into a bad state should this script experience an error.
+
+    Args:
+        learners (list): List of learners to portion into smaller batches (lists)
+        batch_size (int): The number of learners to portion into each batch. If this
+            parameter is not supplied, this function will return one batch containing
+            all of the learners supplied to it.
+    """
+    if batch_size:
+        return [
+            learners[i:i+batch_size] for i, _ in list(enumerate(learners))[::batch_size]
+        ]
+    else:
+        return [learners]
+
+
 def _on_s3_backoff(details):
     """
     Callback that is called when backoff... backs off
@@ -79,7 +101,7 @@ def _on_s3_backoff(details):
     on_backoff=lambda details: _on_s3_backoff(details),  # pylint: disable=unnecessary-lambda,
     max_time=120,  # 2 minutes
 )
-def _upload_to_s3(config, filename):
+def _upload_to_s3(config, filename, dry_run=False):
     """
     Upload the archive file to S3
     """
@@ -99,8 +121,14 @@ def _upload_to_s3(config, filename):
         datestr = datetime.datetime.now().strftime('%Y/%m/')
 
         bucket = s3_connection.get_bucket(config['s3_archive']['bucket_name'])
+        # Dry runs of this script should only generate the retirement archive file, not push it to s3.
         key = Key(bucket, datestr + filename)
-        key.set_contents_from_filename(filename)
+        if dry_run:
+            LOG('Dry run. Skipping the step to upload data to {}'.format(key))
+            return
+        else:
+            key.set_contents_from_filename(filename)
+            LOG('Successfully uploaded retirement data to {}'.format(key))
     except Exception as exc:
         LOG(text_type(exc))
         raise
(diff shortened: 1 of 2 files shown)


tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -63,6 +64,21 @@ def _fetch_learners_to_archive_or_exit(config, start_date, end_date, initial_sta
         FAIL_EXCEPTION(ERR_FETCHING, 'Unexpected error occurred fetching users to update!', exc)
 
 
+def _batch_learners(learners=None, batch_size=None):
+    """
+    To avoid potentially overwheling the LMS with a large number of user retirements to
+    delete, create a list of smaller batches of users to iterate over. This has the
+    added benefit of reducing the amount of user retirement archive requests that can
+    get into a bad state should this script experience an error.
```
pwnage101, line 72: I would complete this documentation and describe each parameter.  For example, not specifying batch_size results is no batching at all.

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -201,10 +225,38 @@ def _cleanup_retirements_or_exit(config, learners):
...
+        archive files are not pushed to s3 and retirements are not cleaned up in the LMS
+    ''',
+    type=bool,
+    default=False
+)
+@click.option(
+    '--start_date',
+    help='Start of window used to select user retirements for archival',
+    type=click.DateTime(formats=['%Y-%m-%d'])
+)
```
pwnage101: one thing that might be worth clarifying in the help text is whether these dates refer to when the learner initiated retirement vs. when they were actually retired.

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -201,10 +225,38 @@ def _cleanup_retirements_or_exit(config, learners):
...
+@click.option(
+    '--end_date',
+    help='''
+        End of window used to select user retirments for archival. This parameter does not
+        include the cool off period
+    ''',
+    type=click.DateTime(formats=['%Y-%m-%d'])
+)
```
pwnage101: hI think I would prefer that the end_date refer to the actual end date passed to the API query.  We can still incorporate the cool off days by making sure the given end_date is cool_off_days days before utcnow(), and error out if not.  What do you think about that?

pwnage101 (commented):
Still requires commit squashing too

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -172,8 +190,14 @@ def _archive_retirements_or_exit(config, learners):
                 }
                 json.dump(user, out)
                 out.write("\n")
-
-        _upload_to_s3(config, filename)
+        # Dry runs of this script should only generate the retirement archive file, not push it to s3.
```
bmedx: It might be worth passing `dry_run` down to `_upload_to_s3` and not write the file to test the S3 connection configuration.

bmedx (approved):
These are great additions! One small optional suggestion, otherwise I agree with Troy's comments.

tubular/scripts/retirement_archive_and_cleanup.py
estute: Yes, that makes sense. We normally will not be passing `end_date` to this script when run in production, it is only for testing purposes. I'll find a way to express that in the code/comments.

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -201,10 +240,44 @@ def _cleanup_retirements_or_exit(config, learners):
...
+        archive files are not pushed to s3 and retirements are not cleaned up in the LMS
+    ''',
+    type=bool,
+    default=False
+)
+@click.option(
+    '--start_date',
+    help='''
+        Start of window used to select user retirements for archival. Only user retirements
+        created after this date will be processed.
```
pwnage101: nit: I would still reword this to: Only user retirements added to the queue after this date will be processed.  Otherwise, it still feels vague to me whether this refers to when the learner was added to the queue vs. actually processed, which could differ by 7+ days.

tubular/scripts/retirement_archive_and_cleanup.py
```diff
@@ -217,17 +290,45 @@ def archive_and_cleanup(config_file, cool_off_days):
...
-        LOG('Fetching learners in COMPLETE status from {} to {}.'.format(start_date, end_date))
-        learners = _fetch_learners_to_archive_or_exit(config, start_date, end_date, 'COMPLETE')
+        if start_date >= end_date:
+            FAIL(ERR_BAD_CLI_PARAM, 'Conflicting start and end dates passed on CLI')
+
+        if end_date > _get_utc_now() - datetime.timedelta(days=cool_off_days):
+            FAIL(ERR_BAD_CLI_PARAM, 'End date cannot occur within the cool_off_days period')
+
+        LOG(
+            'Fetching learners in COMPLETE status from {} to {}.'.format(
```
pwnage101: hmmm...is this text wrong?  I know it was copied from before, but reading the function documentation it seems wong:

https://github.com/edx/tubular/blob/71aa585c4ac5eb385999b65994dc0a1551e9a967/tubular/edx_api.py#L174-L176

So, the function claims that it refers to dates when the learner was enqueued, but this log message claims that it refers to dates when the learner completed retirement processing.  Which is correct?

pwnage101 (approved):
so, the actual code logic all looks good to me now---but I still have nits on docs.

tubular/scripts/retirement_archive_and_cleanup.py
pwnage101: hold up--see my next comment!

PR #533, opened 2021-09-13
feat: script errors and exits if wrong version

jinder1s:
Before merging approved prs into target branch, merge_approved_prs
script now looks to make sure the target branch's head is on correct
version.

This is to make sure nothing gets merged into target branch between the
start of this script and the point at which Prs are merged in.

This mainly affects edxapp pipelines and makes sure the version that
started the pipeline gets into production and not any subsequent commit.

### DIFF
--- a/tubular/scripts/merge_approved_prs.py
+++ b/tubular/scripts/merge_approved_prs.py
@@ -13,6 +13,8 @@ import click
 import click_log
 import yaml
 
+from github import GithubException
+
 from tubular import github_api  # pylint: disable=wrong-import-position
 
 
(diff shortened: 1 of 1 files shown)


dianakhuang (approved):

tubular/scripts/merge_approved_prs.py
```diff
@@ -115,10 +122,12 @@ def octomerge(
     """
     target_github_repo = github_api.GitHubAPI(*target_repo, token=token)
     source_github_repo = github_api.GitHubAPI(*source_repo, token=token)
-
     with target_github_repo.clone(target_branch, target_reference_repo).cleanup() as local_repo:
         local_repo.add_remote('source', source_github_repo.github_repo.ssh_url)
         local_repo.force_branch_to(target_branch, source_branch, remote='source')
+        target_head_sha = local_repo.get_head_sha(branch=target_branch)
+        if sha is not None and target_head_sha != sha:
+            raise GithubException(f"The head sha of target branch does not match inputed sha, head sha: {target_head_sha}, sha: {sha}")
```
timmc-edx: What's the rationale for using GithubException here?

jinder1s: I wasn't sure for what to go with this one. I also do not think GithubException is a good choice. The best alternative I have is to create a new exception class. Thoughts?

timmc-edx: Nothing immediately jumps out at me among the builtin classes, but `AssertionError` is probably close enough.

Ultimately, it doesn't really matter since it's going to be thrown from the CLI code, not from library code.

jinder1s: Okay, switched to assertion error

timmc-edx (approved):

PR #537, opened 2021-09-21
chore: edX API client: Added retry on ConnectionError(104).

HassanJaveed84:


bmedx (approved):
Seems reasonable. This might be a good time to finally rename `_exception_not_internal_svc_error` to something that makes more sense (and is more technically correct).

estute (approved):

PR #539, opened 2021-09-21
fix: add sleep to calls to batches of users in user retirement archiver

estute:
This adds a 10 second sleep between batches.


adzuci (approved):
:shipit:

PR #545, opened 2021-12-07
feat: Add logging to diagnose race condition and branch behavior

timmc-edx:
I believe there's a race condition in the use of `--source-branch` here,
but I don't want to try fixing it until I understand the behavior of the
target branch -- this code should result in isolated commits hanging off of
the history of source-branch, but instead the commits form a merge chain.
I'll be able to compare the additional logging here against the actual
results in the git repos.

ref: ARCHBOM-1895

### DIFF
--- a/tubular/scripts/merge_approved_prs.py
+++ b/tubular/scripts/merge_approved_prs.py
@@ -121,11 +121,13 @@ def octomerge(
     target_github_repo = github_api.GitHubAPI(*target_repo, token=token)
     source_github_repo = github_api.GitHubAPI(*source_repo, token=token)
     with target_github_repo.clone(target_branch, target_reference_repo).cleanup() as local_repo:
+        logging.info(f"Initial target branch at commit {local_repo.get_head_sha(branch=target_branch)}")
         local_repo.add_remote('source', source_github_repo.github_repo.ssh_url)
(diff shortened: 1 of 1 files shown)


tubular/scripts/merge_approved_prs.py
```diff
@@ -141,11 +143,14 @@ def octomerge(
...
         release_name = 'release-{date}'.format(date=datetime.now().strftime("%Y%m%d%H%M%S"))
+        logging.info(f"Octopus merge commit is {merge_sha} and will be tagged as {release_name}")
+        logging.info(f"Target branch now locally at commit {local_repo.get_head_sha(branch=target_branch)}")
         local_repo.repo.create_tag(
             release_name,
             ref=merge_sha,
         )
-        local_repo.push_branch(target_branch, force=True)
+        local_repo.push_branch(target_branch, force=True, log_info=True)
+        logging.info(f"Target branch, after push, is at commit {local_repo.get_head_sha(branch=target_branch)}")
```
jinder1s, line 153: is the output of local_repo.get_head_sha a string?

timmc-edx: It appears so, yeah.

jinder1s (approved):

PR #546, opened 2021-12-08
fix: Don't try to merge security patches when there aren't any

timmc-edx:
When there are no approved PRs, don't perform a merge -- just take the
source_branch commitish as the deployment SHA and new target_branch value.

When `octopus_merge` was called with an empty `approved_prs` list, it was
invoking the no-arg version of `git merge`, which merges the upstream
branch instead (as an implicit argument). So, instead of creating a single
commit off of each source-branch commit, the target-branch commits formed
a *chain* parallel to the source branch (each having two parents -- one a
source branch commit, and the other the previous target-branch commit.)

From `git help merge`

> If no commit is given from the command line, merge the remote-tracking
> branches that the current branch is configured to use as its upstream.

This was probably harmless but created a git history that was overly
difficult to understand.

Also:

- Rename `merge_sha` variable to `deploy_sha` to clarify intent, especially
  since it is now not always a merge commit
- Remove some logging that turned out to be redundant (logging had been
  added to figure out where that strange merge commit was coming from)

ref: ARCHBOM-1895

### DIFF
--- a/tubular/git_repo.py
+++ b/tubular/git_repo.py
@@ -154,9 +154,10 @@ class LocalGitAPI:
     def octopus_merge(self, base_branch, commitishes):
         """
         Merge all ``commitishes`` into ``base_branch`` in this repo.
+        ``base_branch`` will be checked out.
         """
         self.checkout_branch(base_branch)
-        if commitishes:
+        if len(commitishes):
             self.repo.git.merge(*commitishes)
         return self.get_head_sha()
 
--- a/tubular/git_repo.py
+++ b/tubular/git_repo.py
@@ -157,6 +157,9 @@ class LocalGitAPI:
         ``base_branch`` will be checked out.
         """
         self.checkout_branch(base_branch)
+        # Specifically use len here so that generators cause an
+        # exception (`if commitishes` with a generator would always
+        # return True)
         if len(commitishes):
             self.repo.git.merge(*commitishes)
         return self.get_head_sha()
--- a/tubular/scripts/merge_approved_prs.py
+++ b/tubular/scripts/merge_approved_prs.py
@@ -153,7 +153,7 @@ def octomerge(
             release_name,
             ref=deploy_sha,
         )
-        # Logging the push info may help us detect a race condition where concurrent builds
+        # `log_info=True` may help us detect a race condition where concurrent builds
         # both push the target branch, although the tags really matter much more.
         local_repo.push_branch(target_branch, force=True, log_info=True)
         local_repo.push_tags()
--- a/tubular/scripts/merge_approved_prs.py
+++ b/tubular/scripts/merge_approved_prs.py
@@ -141,7 +141,7 @@ def octomerge(
                     for pr in approved_prs
                 )
             ))
-            local_repo.octopus_merge(target_branch, (pr.head.sha for pr in approved_prs))
+            local_repo.octopus_merge(target_branch, [pr.head.sha for pr in approved_prs])
         else:
             logging.info("No PRs to merge")
 
--- a/tubular/scripts/merge_approved_prs.py
+++ b/tubular/scripts/merge_approved_prs.py
@@ -132,30 +132,35 @@ def octomerge(
         approved_prs = list(find_approved_prs(
             target_github_repo, source_github_repo, target_base_branch, source_base_branch
         ))
-        logging.info("Merging the following prs into {}:\n{}".format(
-            target_branch,
-            "\n".join(
-                "    {.html_url}".format(pr)
-                for pr in approved_prs
-            )
-        ))
 
-        merge_sha = local_repo.octopus_merge(target_branch, (pr.head.sha for pr in approved_prs))
+        if approved_prs:
+            logging.info("Merging the following prs into {}:\n{}".format(
+                target_branch,
+                "\n".join(
+                    "    {.html_url}".format(pr)
+                    for pr in approved_prs
+                )
+            ))
+            local_repo.octopus_merge(target_branch, (pr.head.sha for pr in approved_prs))
+        else:
+            logging.info("No PRs to merge")
+
         # The tag encodes the time to ensure that it is a distinct tag.
         release_name = 'release-{date}'.format(date=datetime.now().strftime("%Y%m%d%H%M%S"))
-        logging.info(f"Octopus merge commit is {merge_sha} and will be tagged as {release_name}")
-        logging.info(f"Target branch now locally at commit {local_repo.get_head_sha(branch=target_branch)}")
+        deploy_sha = local_repo.get_head_sha(branch=target_branch)
+        logging.info(f"Pushing commit {deploy_sha} to {target_branch} branch and tagging as {release_name}")
         local_repo.repo.create_tag(
             release_name,
-            ref=merge_sha,
+            ref=deploy_sha,
         )
+        # Logging the push info may help us detect a race condition where concurrent builds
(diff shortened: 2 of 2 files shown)


tubular/scripts/merge_approved_prs.py
```diff
@@ -132,30 +132,35 @@ def octomerge(
...
         ))
-        logging.info("Merging the following prs into {}:\n{}".format(
-            target_branch,
-            "\n".join(
-                "    {.html_url}".format(pr)
-                for pr in approved_prs
-            )
-        ))
 
-        merge_sha = local_repo.octopus_merge(target_branch, (pr.head.sha for pr in approved_prs))
```
timmc-edx, line 143: Actually, it seems like I'm wrong about the mechanism -- `octopus_merge` [has a check](https://github.com/edx/tubular/blob/8dacff97101065e3daef7072b99238875d482abe/tubular/git_repo.py#L159) for an empty list! But it does it using logical booleans, and this `(pr.head.sha for pr in approved_prs)` thing right here is a *generator*, which is Truthy. That's the real failure.

tubular/scripts/merge_approved_prs.py
```diff
@@ -132,30 +132,35 @@ def octomerge(
...
+        if approved_prs:
+            logging.info("Merging the following prs into {}:\n{}".format(
+                target_branch,
+                "\n".join(
+                    "    {.html_url}".format(pr)
+                    for pr in approved_prs
+                )
+            ))
+            local_repo.octopus_merge(target_branch, [pr.head.sha for pr in approved_prs])
```
timmc-edx, line 144: Note the change here from a generator to a list.

tubular/scripts/merge_approved_prs.py
```diff
@@ -132,30 +132,35 @@ def octomerge(
...
-        logging.info("Merging the following prs into {}:\n{}".format(
-            target_branch,
-            "\n".join(
-                "    {.html_url}".format(pr)
-                for pr in approved_prs
-            )
-        ))
 
-        merge_sha = local_repo.octopus_merge(target_branch, (pr.head.sha for pr in approved_prs))
+        if approved_prs:
```
timmc-edx, line 136: This means there's a redundant check, which isn't ideal. I think I'd prefer to have `octopus_merge` raise an exception if it receives an empty list, rather than it being a quasi-no-op (it currently still does a checkout).

tubular/git_repo.py
```diff
@@ -154,9 +154,10 @@ def track_remote_branch(self, remote_name, remote_branch):
     def octopus_merge(self, base_branch, commitishes):
         """
         Merge all ``commitishes`` into ``base_branch`` in this repo.
+        ``base_branch`` will be checked out.
```
robrap, line 157: Can you add an explanation of what a commitish is? Maybe via an arguments section? Note that elsewhere in the file it is described as just being a commit, and for people like me, that didn’t really answer the question of why it is called a commitish.

tubular/git_repo.py
```diff
@@ -154,9 +154,10 @@ def track_remote_branch(self, remote_name, remote_branch):
...
         """
         self.checkout_branch(base_branch)
-        if commitishes:
+        if len(commitishes):
```
robrap, line 163: Since it is non-obvious, can you add a brief comment about the use of len? Since the earlier version is such a common practice, I wonder why this isn’t a more common bug.

tubular/scripts/merge_approved_prs.py
robrap: I’m not clear on the purpose of your comment. Are you saying you plan to make changes, or that you think there would be a better solution but the current code is good enough? Other?

tubular/scripts/merge_approved_prs.py
```diff
@@ -132,30 +132,35 @@ def octomerge(
...
-        logging.info(f"Octopus merge commit is {merge_sha} and will be tagged as {release_name}")
-        logging.info(f"Target branch now locally at commit {local_repo.get_head_sha(branch=target_branch)}")
+        deploy_sha = local_repo.get_head_sha(branch=target_branch)
+        logging.info(f"Pushing commit {deploy_sha} to {target_branch} branch and tagging as {release_name}")
         local_repo.repo.create_tag(
             release_name,
-            ref=merge_sha,
+            ref=deploy_sha,
         )
+        # Logging the push info may help us detect a race condition where concurrent builds
```
robrap: The intent of this comment is unclear. I am unclear if you are describing what we are logging, or what we are not logging and why, or something else. Also, why is this not a comment immediately before the log statement? If this is to comment why logging was removed below, consider leaving it out or saying “We are I tensionally not logging … because…”. Hopefully, my confusion is clear. 🙂

robrap (commented):
Mostly just looking to clear up comments.

tubular/git_repo.py
timmc-edx: I don't actually know. It's a term that's used all through these files, and I think it means "commit, tag, branch, reference, or anything else that can be resolved to a commit".

tubular/git_repo.py
timmc-edx: Will do.

tubular/scripts/merge_approved_prs.py
timmc-edx: Oh, I guess I didn't finish my thought, there. I was going to say that I wasn't sure if it would be too much of an (additional) rabbit hole to go down and that I would likely just leave it as-is. But it looks like `octopus_merge` isn't used anywhere else, so maybe I should just make that change. What do you think?

tubular/scripts/merge_approved_prs.py
timmc-edx: It's about the use of `log_info=True` here, and why it's useful. If I change it to this, does it make more sense?
```suggestion
        # `log_info=True` may help us detect a race condition where concurrent builds
```

tubular/git_repo.py
robrap: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/#_identifier_terminology
```
<commit-ish>
Indicates a commit or tag object name. A command that takes a <commit-ish> argument ultimately wants to operate on a <commit> object but automatically dereferences <tag> objects that point at a <commit>.
```

timmc-edx: Hmm, interesting that that doesn't list branches. Anyway, it looks like `octopus_merge` is only ever called with commit SHA-1 IDs.

tubular/scripts/merge_approved_prs.py
robrap: I think it is fine as-is.

tubular/git_repo.py
robrap: You could note that. But I also think it would be nice to let people know what a commitish is if we are using the name.

tubular/scripts/merge_approved_prs.py
robrap: Yes. That helps a lot, because I had messed that logging. Thanks.

tubular/git_repo.py
timmc-edx: I'd like to skip it here because it's git domain knowledge and is used in another function as well.

robrap (approved):
Thanks. Other requests for comments might help, but are non-blocking.

timmc-edx:
Oh, and I plan to change this to a `fix` conventional commit type on merge. (Was `feat`.)

PR #548, opened 2022-01-07
feat: Allow specifying source commit in merge-PRs; rm unused --sha option

timmc-edx:
This is intended as a temporary expand-contract measure to allow use of the
`--source-branch` option while it is being removed in the calling script;
that option is deprecated because it results in a race condition.

`--sha` was intended as a stop-gap, but it didn't do what we wanted and
isn't currently in use, so this removes it entirely.

ref: ARCHBOM-1895


rgraber (approved):
LGTM

PR #550, opened 2022-01-12
feat: Add requirements upgrade workflow

adzuci:
This workflow should help us more easily keep python requirements up to date.


loucicchese (approved):
Approving based on additional comment from @adzuci _"No risk to master, it opens PRs to be reviewed by the team.  If it's broken the PR won't open or will open with the wrong content."_

PR #565, opened 2022-05-03
Update DriveApi to add support for user accounts

AbdouSeck:
Update the `google_api.py` module to add user account authentication support for the `DriveApi` class. This also adds an `--as_user_account` option to the `delete_expired_partner_gdpr_reports.py` script.

### DIFF
--- a/tubular/google_api.py
+++ b/tubular/google_api.py
@@ -11,12 +11,14 @@ DriveApi is for managing files in google drive.
 
 from itertools import count
 
+import json
 import logging
 from six import iteritems, text_type
 import backoff
 
 from dateutil.parser import parse
 from google.oauth2 import service_account
+from google.oauth2.credentials import Credentials
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
 from googleapiclient.http import MediaIoBaseUpload
@@ -35,6 +37,12 @@ GOOGLE_API_MAX_BATCH_SIZE = 10
 
 # Mimetype used for Google Drive folders.
 FOLDER_MIMETYPE = 'application/vnd.google-apps.folder'
+
+# Fields to be extracted from OAuth2 JSON token files
+OAUTH2_TOKEN_FIELDS = [
+    'client_id', 'client_secret', 'refresh_token',
+    'token_uri', 'id_token', 'scopes', 'access_token'
+]
 
 
 class BatchRequestError(Exception):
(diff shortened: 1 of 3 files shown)


tubular/google_api.py
```diff
@@ -203,7 +230,9 @@ class DriveApi(BaseApiClient):
     _api_version = 'v3'
     _api_scopes = [
         # basic file read-write functionality.
-        'https://www.googleapis.com/auth/drive.file',
+        # 'https://www.googleapis.com/auth/drive.file',
+        # Full read write functionality
+        'https://www.googleapis.com/auth/drive',
```
pwnage101, line 235: what's the purpose of this change?

tubular/google_api.py
```diff
@@ -294,7 +323,11 @@ def delete_files_older_than(self, top_level, delete_before_dt, mimetype=None, pr
             prefix (str): Filename prefix - only files started with this prefix will be deleted.
         """
         LOG.info("Walking files...")
-        all_files = self.walk_files(top_level, 'id, name, createdTime', mimetype)
+        all_files = self.walk_files(
+            top_level,
+            'id, name, createdTime, modifiedTime',
+            mimetype
+        )
```
pwnage101: are any of the changes related to modifiedTime still necessary?

tubular/google_api.py
AbdouSeck: Because `https://www.googleapis.com/auth/drive.file` only allows read and write for files that are BOTH _created_ and _owned_ by the user. However, the old files are created by `Couchdrop` (via the migration) but owned by `No Reply`. On the other hand, `https://www.googleapis.com/auth/drive` gives permissions to the owner to manage the files, regardless of who _created_ said files.

tubular/google_api.py
AbdouSeck: `modifiedTime` should be removed now.

tubular/tests/test_delete_expired_reports.py
```diff
@@ -67,12 +67,22 @@ def test_successful_report_deletion(*args):
     mock_driveapi = args[2]
 
     test_created_date = '2018-07-13T22:21:45.600275+00:00'
+    test_modified_date = '2018-07-13T22:21:45.600275+00:00'
```
pwnage101: I think all the changes to both test files can be removed too, right?

AbdouSeck: They should have been. They should all be gone now.

pwnage101 (approved):

PR #567, opened 2022-05-11
build: Stop testing on Python 3.6 and 3.7

timmc-edx:
3.8 is the only version we support right now.


timmc-edx:
There are still 3.6 and 3.7 references in setup.cfg and setup.py indicating that those versions are "supported", but in practice no one has been using those versions, so it's probably worth getting rid of them as well.

oel8man (approved):

PR #578, opened 2022-07-07
fix: Stop crashing on non existent id ISRE-1377

jdmulloy:


tubular/scripts/structures.py
```diff
@@ -97,8 +97,13 @@ def cli(ctx, connection, database_name):
     type=click.IntRange(1, None),
     help="How many Structures do we fetch at a time?"
 )
+@click.option(
+    '--force/--no-force',
```
ormsbee: Maybe something more specific to the problem like: `--repair-missing`?

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
             for int_structure_id in int_structure_ids_to_save:
                 structure_ids_to_save.add(int_structure_id)
 
+        for structure_id in structure_ids_to_save:
+            if structure_id not in structures:
+                missing_structure_ids.add(structure_id)
```
ormsbee: I think this can be simplified to:

```python
missing_structure_ids = structure_ids_to_save - structures.keys()
```

(The empty set declaration of it from line 222 can also be removed in that case.)

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+
+        for missing_structure_id in missing_structure_ids:
```
ormsbee: This should have comments about what it's trying to do.

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+                    family_ids.append(sid)
+
+            family_ids.sort()
+
+            for sid in family_ids:
+                LOG.debug(f"Traversing {sid}")
+                for tid in structures_graph.traverse_ids(sid):
+                    if tid in structures:
+                        LOG.debug(f"traversed structure id: {tid}, original_id: {structures[tid].original_id}, previous_id: {structures[tid].previous_id}")
+
```
ormsbee, line 253: This would be a good place to abort with a clean error message, if we haven't gotten the special flag that lets us repair the history. We _know_ it's going to blow up in the next batch of code anyway, but we can do a cleaner error message here and exit, and then not have to worry about whether we're in that state or not later on.

ormsbee:
In the middle of reviewing this and had a silly question: Do we know if the nodes that were incorrectly deleted were intermediate structures or original structures? Because as long as the original survived, repairing it would be straightforward. But it gets weird and worse if original docs were deleted.

Ugh, maybe this was some off-by-one error in the traversal code?

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+            if force:
+                rewrite_candidates = takewhile(
+                    lambda s: s in structure_ids_to_save and s in structures and not structures[s].is_original(),
+                    structures_graph.traverse_ids(branch.structure_id, include_start=True)
+                )
+            else:
+                rewrite_candidates = takewhile(
+                    lambda s: s in structure_ids_to_save and not structures[s].is_original(),
+                    structures_graph.traverse_ids(branch.structure_id, include_start=True)
+                )
```
ormsbee: At this point, would it make more sense to leave the code as is, and have some logic above this for loop that looks something like:

```python
# Can't save missing structures.
structure_ids_to_save -= missing_structures_ids
```

(With appropriate logging, of course.)

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+        for structure_id in structure_ids_to_save:
+            if structure_id not in structures:
+                missing_structure_ids.add(structure_id)
+
+        for missing_structure_id in missing_structure_ids:
+            LOG.error(f"Missing structure ID: {missing_structure_id}")
+            original_id = None
+            # family_ids is a list of ids with the same original_id
+            family_ids = []
+            for sid, structure in structures.items():
```
ormsbee: If you only need the values, you can iterate through `structures.values()` instead of `structures.items()`.

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+                if structure.previous_id == missing_structure_id:
+                    LOG.info(f"structure id: {structure.id}, original_id: {structure.original_id}, previous_id: {structure.previous_id}")
```
ormsbee: Maybe error level? This kind of information is also nice to have in the details file, but that's not critical.

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+                    original_id = structure.original_id
```
ormsbee: What is this assignment for?

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+
+            for sid, structure in structures.items():
+                if structure.original_id == original_id:
+                    family_ids.append(sid)
```
ormsbee: Nit: A list comprehension might be useful here.

tubular/splitmongo.py
```diff
@@ -240,13 +241,45 @@ def create(cls, structures_graph, num_intermediate_structures, details_file=None
...
+
+            family_ids.sort()
+
+            for sid in family_ids:
+                LOG.debug(f"Traversing {sid}")
+                for tid in structures_graph.traverse_ids(sid):
+                    if tid in structures:
+                        LOG.debug(f"traversed structure id: {tid}, original_id: {structures[tid].original_id}, previous_id: {structures[tid].previous_id}")
```
ormsbee: I'm not clear on what this is showing us.

ormsbee (changes requested):
Some requests and questions. Also a thought: Is there a way to connect the missing structure history to a specific course + branch and then show the details of that history to see where it might have gotten messed up?

Thank you.

tubular/splitmongo.py
jdmulloy: Thanks @ormsbee I know there's better ways to do some of these things that are not as easy to understand and it takes so long to run this script that I went with code I knew would work vs code that might fail. I'll implement what I can.

tubular/splitmongo.py
jdmulloy: @ormsbee I'm not sure how to do that. It's related to your question about original_id above.

I look through all the structures to find the structure that references the missing_id in previous_id and then I go through again to find all structures with the same original_id so I can traverse the structures to print out the whole chain if debug is turned on. Although it does the work even if it doesn't print.

This structures stuff melts my sleep deprived brain. The output is definitely repetitive.

For instance if there is a chain A->B->C->D->E it will traverse and print for A->B->C->D->E, B->C->D->E, C->D->E, and D->E. It may not really be helpful. I'm just trying to make something useful if we run into this again.

PR #582, opened 2022-10-28
fix: The error logging for FrontendBuilder object needs two arguments supplied

schenedx:
This is a bug fix. See the impact of this bug at this [GOCD error message](https://gocd.tools.edx.org/go/tab/build/detail/stage-frontend-app-learning/592/stage_frontend-app-learning_build_frontend/1/stage_frontend-app-learning_build_frontend_job).
Make sure the logging statements supplies 2 arguments

### DIFF
--- a/tubular/scripts/frontend_utils.py
+++ b/tubular/scripts/frontend_utils.py
@@ -59,7 +59,7 @@ class FrontendBuilder:
         proc = subprocess.Popen(['npm install'], cwd=self.app_name, shell=True)
         return_code = proc.wait()
         if return_code != 0:
-            self.FAIL('Could not run `npm install` for app {}.'.format(self.app_name))
+            self.FAIL(1, 'Could not run `npm install` for app {}.'.format(self.app_name))
 
         self.install_requirements_npm_aliases()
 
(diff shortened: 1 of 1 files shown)


tubular/scripts/frontend_utils.py
```diff
@@ -59,7 +59,7 @@ def install_requirements(self):
         proc = subprocess.Popen(['npm install'], cwd=self.app_name, shell=True)
         return_code = proc.wait()
         if return_code != 0:
-            self.FAIL('Could not run `npm install` for app {}.'.format(self.app_name))
+            self.FAIL(1, 'Could not run `npm install` for app {}.'.format(self.app_name))
```
mattcarter, line 62: Did we assign a meaning to the 0-index variable?

zacharis278 (approved):

adzuci (approved):
LGTM, I'm also curious about how you landed on this fix, but it seems reasonable to me.

tubular/scripts/frontend_utils.py
schenedx: It's the [system exit code](https://github.com/openedx/tubular/blob/58c9dd12e53d9e3a249a528674dfb25dc4274dbd/tubular/scripts/helpers.py#L39). I believe 1 is for errored exits. 0 is successful exit.

PR #585, opened 2022-11-22
fix: slack bot authentication for GoCD notifications

matthugs:
evidently the old way we were doing this was grandfathered in in a way I couldn't find any information about

this also will turn non-notifying pipelines red going forward

JIRA:PSRE-2003

### DIFF
--- a/tubular/slack.py
+++ b/tubular/slack.py
@@ -5,7 +5,6 @@ import requests
 
 SLACK_API_URL = "https://slack.com"
 NOTIFICATION_POST = "/api/chat.postMessage"
-CONTENT_TYPE = "application/json"
 AUTH_HEADER_FIELD = "Authorization"
 
 LOG = logging.getLogger(__name__)
@@ -30,19 +29,23 @@ def submit_slack_message(auth_token, channels, message):
     post_url = "{}{}".format(SLACK_API_URL, NOTIFICATION_POST)
     # to remove slack API warning
     headers = {
-        'Content-type': CONTENT_TYPE
+        AUTH_HEADER_FIELD: 'Bearer ' + auth_token
     }
     for channel in channels:
-        params = {
-            "token": auth_token,
+        arguments = {
             "channel": channel,
             "text": message
         }
         response = requests.post(post_url,
-                                 params=params,
+                                 data=arguments,
                                  headers=headers
                                  )
         if response.status_code not in (200, 201, 204):
             raise SlackMessageSendFailure(
-                "Message send to channel '{}' failed: {}".format(channel, response.text)
+                f"Message send to channel '{channel}' failed: {response.text}"
+            )
+        response_json = response.json()
+        if not response_json['ok']:
+            raise SlackMessageSendFailure(
+                f"Message send to channel '{channel}' failed: {response.text}"
             )


matthugs:
manually tested in devops slack where we have greater permissions to add apps with more impunity: Screenshot 2022-11-22 at 11 45 52 AM

tubular/slack.py
```diff
@@ -30,19 +29,23 @@ def submit_slack_message(auth_token, channels, message):
     post_url = "{}{}".format(SLACK_API_URL, NOTIFICATION_POST)
     # to remove slack API warning
     headers = {
-        'Content-type': CONTENT_TYPE
```
jdmulloy, line 33: DO we not need content type anymore?

ohnickmoy (approved):
Huh, bizzare

tubular/slack.py
```diff
@@ -30,19 +29,23 @@ def submit_slack_message(auth_token, channels, message):
...
+        AUTH_HEADER_FIELD: 'Bearer ' + auth_token
     }
     for channel in channels:
-        params = {
-            "token": auth_token,
+        arguments = {
             "channel": channel,
             "text": message
         }
         response = requests.post(post_url,
```
jdmulloy, line 39: SHould we want to wrap this whole thing in a try catch in case the requests.post() call throws an exeception?

jdmulloy (approved):

adzuci (approved):
LGTM, could you link to the docs you used?  I hunted in https://slack.dev/python-slack-sdk/oauth/index.html#sign-in-with-slack quickly, but didn't see a headers example like this.

I'm also not sure why you removed the json content type bit.

tubular/slack.py
matthugs: we do not https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests https://github.com/psf/requests/blob/bda7f0171f8bba17989d3a2c28dfa9a9261b1b65/requests/models.py#L563

matthugs:
> LGTM, could you link to the docs you used? I hunted in https://slack.dev/python-slack-sdk/oauth/index.html#sign-in-with-slack quickly, but didn't see a headers example like this.
> 
> I'm also not sure why you removed the json content type bit.

Interestingly we were never posting our arguments as json; the python `requests` library pattern for doing that looks like this:

```
        response = requests.post(post_url,
                                 data=json.dumps(arguments),
                                 headers=headers
                                 )
```

Note that the keyword argument to requests.post in the old way has `params` instead of `data`, instructing requests to post these as url parameters rather than as the body of the request.

Slack API docs: https://api.slack.com/web#slack-web-api__basics__post-bodies__json-encoded-bodies
Reference for the above: https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls & https://github.com/psf/requests/blob/bda7f0171f8bba17989d3a2c28dfa9a9261b1b65/requests/models.py#L474

tubular/slack.py
matthugs: I'm leaning in the opposite direction. I'd like this to fail more loudly in the future rather than failing silently as it has been.

PR #589, opened 2022-12-22
fix: tox failures due to multiple package-looking directories

matthugs:
but best I can tell we no longer use this script so let's just get rid of it

If tox hangs on this then my experiment was a failure


adzuci (approved):
LGTM

PR #600, opened 2023-01-17
chore: Add new option to prune batch size and change defaults for max size

dnuzz:
allowing pruning of returned user list to prevent overloading downstream processes.

### DIFF
--- a/tubular/scripts/get_learners_to_retire.py
+++ b/tubular/scripts/get_learners_to_retire.py
@@ -41,12 +41,19 @@ LOG = logging.getLogger(__name__)
     '--user_count_error_threshold',
     help="If more users than this number are returned we will error out instead of retiring. This is a failsafe"
          "against attacks that somehow manage to add users to the retirement queue.",
+    default=300
+)
+@click.option(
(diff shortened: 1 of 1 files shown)


tubular/scripts/get_learners_to_retire.py
```diff
@@ -72,6 +79,8 @@ def get_learners_to_retire(config_file,
 
     # Retrieve the learners to retire and export them to separate Jenkins property files.
     learners_to_retire = api.learners_to_retire(states_to_request, cool_off_days)
```
Hamza442, line 81: Shouldn't we limit batch size while making `api.learners_to_retire()` request?

dnuzz: I don't think the API allows us to limit the size of the returned payload, at least I did not see any references to that in the API: https://github.com/openedx/edx-platform/blob/c1e4cb3a8761acc336e2ca1da92a4477e04f63dd/openedx/core/djangoapps/user_api/accounts/views.py#L917

Hamza442 (approved):
LGTM

PR #602, opened 2023-01-18
fix: Increased number of records per page to 100.

awais786:
Increased number of records per hit to 100. We don't need to add pagination in depth. 
This script will run daily and not expecting many prs.

### DIFF
--- a/tubular/scripts/get_ready_to_merge_prs.py
+++ b/tubular/scripts/get_ready_to_merge_prs.py
@@ -17,14 +17,14 @@ sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
 logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


iamsobanjaved (approved):

awais786:
@nedbat Please review.

tubular/scripts/get_ready_to_merge_prs.py
```diff
@@ -17,14 +17,14 @@
 logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 LOG = logging.getLogger(__name__)
 
-GIT_API_URL = 'https://api.github.com/search/issues?'
+GIT_API_URL = 'https://api.github.com/search/issues?per_page=100'
```
nedbat, line 20: OK, this still won't get all of the pull requests, but we probably won't be getting more than 100 per day anyway.

nedbat (approved):

PR #615, opened 2023-02-22
fix: remove future package usages and dependency

UsamaSadiq:
## Description
- As uncovered in https://pyup.io/vulnerabilities/CVE-2022-40899/52510/, the `future` package is not secure to be used.
- `arbi-bom` is removing `future` package from all repos so this PR is being created under this effort.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -42,9 +42,9 @@ click==8.1.7
     #   edx-django-utils
 click-log==0.4.0
     # via -r requirements/base.in
-cloudflare==2.11.7
+cloudflare==2.12.4
     # via -r requirements/base.in
-cryptography==41.0.3
+cryptography==41.0.4
     # via
     #   pyjwt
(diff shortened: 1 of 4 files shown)


requirements/base.txt
```diff
@@ -77,7 +77,6 @@ freezegun==0.3.8
     # via -r requirements/base.in
 future==0.16.0
     # via
-    #   -r requirements/base.in
     #   cloudflare
```
UsamaSadiq: Created PR https://github.com/cloudflare/python-cloudflare/pull/156 to remove `future` dependency from `cloudflare` as well. Once the change gets merged there, we can completely remove the usage of `future` package.

iamsobanjaved (approved):

requirements/base.txt
UsamaSadiq: The upstream repo has been updated so we can now completely remove the future package usage.

PR #616, opened 2023-02-23
feat: user retirement from Amplitdue integrated to user_retirement script

usama101:


tubular/amplitude_api.py
```diff
@@ -0,0 +1,77 @@
...
+            'Authorization': 'Basic {api_key}:{secret_key}'.format(api_key=self.amplitude_api_key, secret_key=self.amplitude_secret_key),
+            'Content-Type': 'application/json'
+        }
+
+
+    def delete_params(self, user):
+        """Returns deletion params required for passing to the requests library"""
+        
+        return json.dumps({
+            "user_ids": [user['user']['id']],
```
pwnage101: style nit: don't mix single and double quotes.  I recommend using double quotes everwhere.

tubular/amplitude_api.py
```diff
@@ -0,0 +1,77 @@
...
+    def auth_headers(self):
+        """Returns authorization headers suitable for passing to the requests library"""
+        return {
+            'Authorization': 'Basic {api_key}:{secret_key}'.format(api_key=self.amplitude_api_key, secret_key=self.amplitude_secret_key),
+            'Content-Type': 'application/json'
+        }
+
+
+    def delete_params(self, user):
+        """Returns deletion params required for passing to the requests library"""
```
pwnage101: IMO, this method is too simple, and is only used once elsewhere, so just move the logic inline.

tubular/amplitude_api.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @backoff.on_exception(
+        backoff.expo,
+        AmplitudeRecoverableException,
+        max_tries=MAX_ATTEMPTS,
+    )
+    def delete_user(self, user):
+        """
+        Delete a user from Amplitude.
+        """
```
pwnage101: style nit: the org-wide python style adopted is to specify all arguments, return values, and possible exceptions.  See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
e.g.:

```suggestion
        """
        Delete a user from Amplitude.
        
        Returns:
            None
        
        Args:
            user (dict): raw data of user to delete.

        Raises:
          AmplitudeException: if the error from amplitude is unrecoverable/unretryable.
        """
```

tubular/amplitude_api.py
```diff
@@ -0,0 +1,77 @@
...
+    def delete_params(self, user):
+        """Returns deletion params required for passing to the requests library"""
+        
+        return json.dumps({
+            "user_ids": [user['user']['id']],
+            "requester": "user-retirement-pipeline"
+        })
+
+    def process_response(self, response, action):
+        """Log response status and raise an error as needed"""
```
pwnage101: I also don't think this needs to be its own function.  I'd rather see this consolidated into delete_user.

tubular/amplitude_api.py
```diff
@@ -0,0 +1,77 @@
...
+        if response.status_code == 200:
+            logger.info('Amplitude {action} succeeded'.format(action=action))
+            return
+
+        # We have some sort of error. Parse it, log it, and retry as needed.
+        error_msg = 'Amplitude {action} failed due to {reason}'.format(action=action, reason=response.reason)
+        logger.error(error_msg)
+
+        if response.status_code == 429 or 500 <= response.status_code < 600:
```
pwnage101, line 90: Can you add a comment that explains why 429 is recoverable?

tubular/amplitude_api.py
```diff
@@ -0,0 +1,77 @@
...
+        backoff.expo,
+        AmplitudeRecoverableException,
+        max_tries=MAX_ATTEMPTS,
+    )
+    def delete_user(self, user):
+        """
+        Delete a user from Amplitude.
+        """
+        response = requests.post(
+            self.base_url + 'api/2/deletions/users',
```
pwnage101: Nit: "api/2/deletions/users" is a string constant which should be defined in a global variable at the top of this file.

tubular/tests/test_amplitude.py
```diff
@@ -0,0 +1,68 @@
...
+            'https://amplitude.com/api/2/deletions/users',
+            request_headers={
+                'Authorization': 'Basic test-api-key:test-secret-key',
+                'Content-Type': 'application/json'
+            },
+            json={},
+            status_code=status_code
+        )
+
+    def test_delete_happy_path(self, req_mock):
```
pwnage101, line 42: These tests look great!  Only nit is that each test should have a docstring explaining what the test is for.

pwnage101 (approved):
I actually only have nits, the logic looks good!

PR #620, opened 2023-03-08
feat: Remove now-unused scripts relating to edxapp deployments

timmc-edx:
These scripts were used in parts of the edxapp deployment pipeline that have now been deleted. Some of the functionality may be useful again in the future. If you want to restore one, please do the following:

- Remove any parts that aren't needed, such as options that don't make sense for your specific use-case. These scripts grew some warts as their usage changed over time.
- Update the code to latest naming standards and coding practices.
- Generally review the code for problems, such as the write-to-file approach used by several of them, or confusing parameter names and docs.

Other changes:

- Remove copypasta import of `confluence_api` in gocd-related script
- Remove sketchy default for base branch in `create_pull_request`

Removed files:

- `create_pr.py` was used to create merge-back PRs in edxapp deployment (https://github.com/edx/edx-internal/pull/7950). This was also the last usage of `have_branches_diverged` in the git util.
- `cut_branch.py` was used for creating `release-candidate-1234` branches in edx-platform (https://github.com/edx/edx-internal/pull/7951). This and `create_release_candidate.py` were the only usages of `extract_message_summary`.
- `create_release_candidate.py` probably hasn't been used in 6 years and had a lot of duplication with cut_branch.py. I can't find where it was last used, but I think it predates edxapp.yaml in edx-internal. Anyway, we definitely won't need this one again, as it was used in scheduling releases. This was the last usage of `default_expected_release_date`, various scheduling constants, `most_recent_good_commit`, `NoValidCommitsError`, `rc_branch_name_for_date`, `create_branch`, `track_remote_branch`
- `update_release_page.py` stopped being useful after the wiki split, since it continued updating the openedx wiki in a space that 2U employees couldn't even access any more. No one has complained, so we're deleting it too. (https://github.com/edx/edx-arch-experiments/issues/215). This also allows us to remove the `confluence_api` module and a few Python dependencies as well.


timmc-edx:
Author's note: I took a maximalist approach to deletion here, removing any supporting utility code that was only used by the deleted scripts on the basis that 1) a lot of it was pretty old and grungy, and should be rewritten anyhow and 2) less to maintain. But feel free to ask that various deletions be backed out again, if you think they're in good shape and there's a strong likelihood we'll need them in the future in deployment pipelines.

tubular/git_repo.py
```diff
@@ -136,21 +136,6 @@ def get_head_sha(self, branch=None):
         else:
             return self.repo.head.commit.hexsha
 
-    def create_branch(self, branch_name, commit='HEAD'):
```
dianakhuang, line 139: Genuinely surprised to see that this isn't used for anything else in the tubular repo.

dianakhuang (approved):
Wow, this is so much unnecessary work we won't be doing anymore.

tubular/git_repo.py
timmc-edx: Yeah! But it looks like `create_private_to_public_pr` (which _is_ still used) uses `push_branch` to create a new branch on the remote, rather than explicitly creating it locally. :shrug:

PR #627, opened 2023-03-22
chore: Allow retirement redrive

dnuzz:
This allows the bulk update command to rewind the state of everyone in a particular state to their last known state. This is useful for rewinding a large batch of Errored users.

### DIFF
--- a/tubular/scripts/retirement_bulk_status_update.py
+++ b/tubular/scripts/retirement_bulk_status_update.py
@@ -81,7 +81,7 @@ def _update_learners_or_exit(config, learners, new_state=None, rewind_state=Fals
     will be reset to their previous state.
     """
     if (not new_state and not rewind_state) or (rewind_state and new_state):
-        raise ValueError("You must specify either the boolean rewind_state or a new state to set learners to.")
+        FAIL(ERR_BAD_CONFIG, "You must specify either the boolean rewind_state or a new state to set learners to.")
     LOG('Updating {} learners to {}'.format(len(learners), new_state))
     try:
         for learner in learners:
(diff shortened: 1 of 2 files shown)


koffemaria2u (approved):
LGTM

tubular/scripts/retirement_bulk_status_update.py
```diff
@@ -74,14 +74,19 @@ def _fetch_learners_to_update_or_exit(config, start_date, end_date, initial_stat
...
+    it will exit the script. If rewind_state is set to True then the learner
+    will be reset to their previous state.
     """
+    if (not new_state and not rewind_state) or (rewind_state and new_state):
+        raise ValueError("You must specify either the boolean rewind_state or a new state to set learners to.")
     LOG('Updating {} learners to {}'.format(len(learners), new_state))
     try:
         for learner in learners:
+            if rewind_state:
+                new_state = learner['last_state']['state_name']
```
HassanJaveed84, line 89: In case of an error, we reset the user's state to the last completed state. For example, when a state transitions from `ENROLLMENTS_COMPLETE` to `RETIRING_NOTES` and then goes into `ERRORED` state, we reset the state to `ENROLLMENTS_COMPLETE` and not `RETIRING_NOTES`. 

This change would reset it to `RETIRING_NOTES` instead, which might still be okay ?

HassanJaveed84:
Is it possible to add some tests ?

tubular/scripts/retirement_bulk_status_update.py
dnuzz: So when I looked at the config here: https://github.com/edx-ops/user-retirement-secure/blob/dbe3cc1e2d85d206e827b75cf0ce0269ae591739/prod-edx.yml#L48

It looks like a bunch of those states are consumed in this line: https://github.com/openedx/tubular/blob/30446885b468d8e2163407edc50189e7c426f4d4/tubular/scripts/get_learners_to_retire.py#L76

And then from there the logic actually tries the state it is on, instead of the next state: https://github.com/openedx/tubular/blob/30446885b468d8e2163407edc50189e7c426f4d4/tubular/scripts/retire_one_learner.py#L183

So I _think_ we are good?

HassanJaveed84 (approved):

PR #632, opened 2023-03-31
fix: Convert ASG code to use boto3.

dianakhuang:
This will definitely fail tests. 😭


tubular/ec2.py
```diff
@@ -136,36 +138,52 @@ def active_ami_for_edp(env, dep, play):
...
+    asg_client = boto3.client('autoscaling')
+
     all_elbs = get_all_load_balancers()
     LOG.info("Found {} load balancers.".format(len(all_elbs)))
-
     edp_filter = {
         "tag:environment": env,
         "tag:deployment": dep,
         "tag:play": play,
     }
```
matthugs, line 151: we'll wanna' remove this and related things

matthugs (approved):

PR #633, opened 2023-03-31
fix: Skip ASGs that are returning bad data.

dianakhuang:
This is a workaround for an issue on AWS we are
seeing with the boto library.


iloveagent57 (approved):
worth a shot!

dianakhuang:
Unfortunately, the data for some valid ASGs are missing tags, so this workaround won't work.

PR #649, opened 2023-04-21
docs: Update the contributing guidelines link.

feanil:
We're moving towards a single set of guidelines org-wide.


kdmccormick (approved):

feanil:
Looks like this is a known test issue unrelated to this change and people have been merging even with this test failing, I've raised the issue with @dianakhuang and @adzuci but I'm gonna merge this change as it is only a doc change.

PR #652, opened 2023-05-03
feat: update how auth to hubspot api is done

matthugs:
Hubspot API needs us to pass its auth token as a header rather than as a query param.

credit from https://github.com/openedx/tubular/pull/650, getting this started

JIRA:PSRE-2394

### DIFF
--- a/tubular/hubspot_api.py
+++ b/tubular/hubspot_api.py
@@ -12,8 +12,8 @@ from tubular.tubular_email import send_email
 LOG = logging.getLogger(__name__)
 MAX_ATTEMPTS = int(os.environ.get('RETRY_HUBSPOT_MAX_ATTEMPTS', 5))
 
-GET_VID_FROM_EMAIL_URL_TEMPLATE = "https://api.hubapi.com/contacts/v1/contact/email/{email}/profile?hapikey={apikey}"
-DELETE_USER_FROM_VID_TEMPLATE = "https://api.hubapi.com/contacts/v1/contact/vid/{vid}?hapikey={apikey}"
+GET_VID_FROM_EMAIL_URL_TEMPLATE = "https://api.hubapi.com/contacts/v1/contact/email/{email}/profile"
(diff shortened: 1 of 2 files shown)


ohnickmoy (approved):

tubular/hubspot_api.py
```diff
@@ -58,10 +58,14 @@ def delete_user_by_vid(self, vid):
         """
         Delete a learner from hubspot using their Hubspot `vid` (unique identifier)
         """
+        headers = {
+            'content-type': 'application/json',
+            'authorization': 'Bearer %s' % {self.api_key}
```
adzuci: We may have already sorted this out, but I believe this should be a new type of key which Taylor offered to help create for us with certain scoped permissions, is that correct and are there any remaining steps to get that new key?

adzuci (approved):
Is there a way to test this code locally with a key?  Possibly with help from marketing?

tubular/hubspot_api.py
matthugs: @adzuci Based on https://developers.hubspot.com/changelog/upcoming-api-key-sunset, yes it seems like that's correct.

I suspect whatever Taylor gives us would get plugged in here: https://github.com/edx-ops/user-retirement-secure/blob/c3319f3c9f54570634ecb5f6d75cdfce13819058/prod-edx.yml#L26
That is a code change we would likely need to synchronize with the merging of this PR.

matthugs:
> Is there a way to test this code locally with a key? Possibly with help from marketing?

we _could_ do something like this: https://2u-internal.atlassian.net/wiki/spaces/DE/pages/10784609/How+To+Run+a+set+of+users+through+part+of+the+retirement+process

Since the pipeline step is idempotent we could test on a learner who was about to be deleted anyway and be unlikely to interrupt the pipeline running cleanly for that user if all goes well.

EDIT: so tested with a couple users; although it took some guess-and-check before I found a not-yet-deleted user who had any data in hubspot

tubular/hubspot_api.py
matthugs: addressed in https://github.com/edx-ops/user-retirement-secure/pull/219

PR #658, opened 2023-05-18
feat: add codecov to tubular repo

ohnickmoy:
[PSRE-2395](https://2u-internal.atlassian.net/browse/PSRE-2395) Add Codecov to tubular repo

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -9,7 +9,6 @@ jobs:
       matrix:
         os: [ubuntu-20.04]
         python-version: ['3.8']
-        toxenv: [django32]
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
@@ -22,7 +21,7 @@ jobs:
     - run: tox -- -n0
 
     - name: Run coverage
-      if: matrix.python-version == '3.8' && matrix.toxenv == 'django32'
+      if: matrix.python-version == '3.8'
       uses: codecov/codecov-action@v3
       with:
         flags: unittests
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04]
         python-version: ['3.8']
-        toxenv: [quality, docs, pii_check, django32, django40]
+        toxenv: [django32]
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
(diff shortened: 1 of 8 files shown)


.github/workflows/tox.yml
```diff
@@ -4,10 +4,12 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-20.04]
         python-version: ['3.8']
+        toxenv: [quality, docs, pii_check, django32, django40]
```
ohnickmoy: Question: do all these variables need to be present for the build to take place?

.github/workflows/tox.yml
```diff
@@ -18,3 +20,10 @@ jobs:
     - run: pip3 install --upgrade pip setuptools
     - run: pip3 install 'tox'
     - run: tox -- -n0
+
```
ohnickmoy, line 22: Tacked on codecov into the tox workflow by matching it against [cookie cutter](https://github.com/openedx/edx-cookiecutters/blob/06f6914d194d840e050f3a1991c8d419ed6007b8/cookiecutter-django-app/%7B%7Bcookiecutter.repo_name%7D%7D/.github/workflows/ci.yml#L41). Would this be better off in it's own worklow?

.github/workflows/tox.yml
matthugs: no they almost assuredly don't

what these do is say "tox should be run with these environments, as configured in tox.ini" which I think some of them might not be in our tox.ini

.github/workflows/tox.yml
matthugs: no this is perf

adzuci (approved):
LGTM, but I see it isn't finding any test output, do we need to make it look in a different place or run tests with different args?

.github/workflows/tox.yml
```diff
@@ -18,3 +20,10 @@ jobs:
     - run: pip3 install --upgrade pip setuptools
     - run: pip3 install 'tox'
     - run: tox -- -n0
```
matthugs, line 21: in the absence of passing a `TOXENV` environment variable here (following [this](https://github.com/openedx/edx-cookiecutters/blob/master/cookiecutter-django-app/%7B%7Bcookiecutter.repo_name%7D%7D/.github/workflows/ci.yml#LL36C1-L37C1)), we probably shouldn't add a toxenv to the github action's matrix. To support that we can remote the `matrix.toxenv` condition below in the `if`

PR #659, opened 2023-05-19
chore: removed un-used script.

awais786:
Removed un-used script.


awais786:
https://github.com/openedx/tubular/pull/657

Upgrade boto3 PR which is not require.

matthugs (approved):

PR #672, opened 2023-06-05
chore: fixing import order to fix retirement job.

awais786:
user retirement job is breaking due to wrong import.

It was changed in previous PR using isort.


zubairshakoorarbisoft (approved):
LGTM 👍

PR #673, opened 2023-06-05
fix: Correct filter inputs

jdmulloy:
After the upgrade form boto to boto3 this code was causing a stack trace like the following.

botocore.exceptions.ParamValidationError: Parameter validation failed:
Unknown parameter in Filters[0]: "key-name", must be one of: Name, Values


matthugs (approved):
haven't checked docs to confirm but looks plausible

jdmulloy:
> haven't checked docs to confirm but looks plausible

I tested manually on my laptop, and it doesn't crash and wants to terminate old build instances.

PR #679, opened 2023-06-29
feat: close alerts via OpsGenie API

rgraber:
Update alert_opsgenie to allow us to add aliases to alerts. This will make them easier to find and close later.
Add close_opsgenie_alert command to allow us to close alerts by alias.
Also adds some tests.


adzuci (approved):
LGTM, thanks for adding the tests too!

